Issue #214, opened 2021-06-09
Use a dropdown for heading buttons in rich text editor


virajbahulkar:
Icon size should be same as h1 - h6 tags

hnhparitosh:
Please let me work on this issue.

Navaneeth-pk:
@hnhparitosh sorry, we missed your comment. Since it has been more than a month, assigning the issue to @saisumanthkumar

saisumanthkumar:
@Navaneeth-pk I guess both this and #211 are same

PR #216, opened 2021-06-09
Change icon size

virajbahulkar:
#214 

Screenshot 2021-06-09 at 6 13 05 PM


virajbahulkar:
@Navaneeth-pk Kept font size of icons similar to headings size.

Navaneeth-pk:
@virajbahulkar 
Can you please make them smaller? I think the icons still stands out as the most emphasized objects on the canvas. Can we also try changing the fill colour of icons to that of the text buttons? I think that makes the UI of the editor look consistent.

virajbahulkar:
> @virajbahulkar
> Can you please make them smaller? I think the icons still stands out as the most emphasized objects on the canvas. Can we also try changing the fill colour of icons to that of the text buttons? I think that makes the UI of the editor look consistent.

Screenshot 2021-06-09 at 11 47 06 PM

is it okay?

Navaneeth-pk:
Looks good to me. Thanks @virajbahulkar

PR #220, opened 2021-06-09
[223] Fixes the rails server connection to docker postgres service

tejaswinichile:
Fixes the issue :  #223 
`ActiveRecord::NoDatabaseError: could not connect to server: No such file or directory
    Is the server running locally and accepting
    connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?`

The issue was: Docker rails server was trying to find the Postgres service on the host machine.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,6 +3,7 @@ default: &default
   encoding: unicode
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
 
+
 development:
   <<: *default
   database: tooljet_development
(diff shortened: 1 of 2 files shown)


tejaswinichile:
I am not sure if this could be a fix, It worked for me, and my other Postgres connection working on my host machine.

This command is to make container the both accessible from TCP/IP and Unix socket. I am not so sure about it 
`docker run -p 5432:5432 -v /var/run/postgresql:/var/run/postgresql -d --name postgres postgres`

PR #221, opened 2021-06-09
Adding annotate gem

JibranKalia:
I personally really enjoy having the `annotate` gem putting the details of the schema on the models.
I find it useful for grepping column names but also verifying if appropriate indexes exist.

That being said I don't feel too strongly about it so feel free close this PR if you don't want this. Annotate can be annoying when it starts trailing migrations.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,7 @@ group :development, :test do
 end
 
 group :development do
+  gem 'annotate'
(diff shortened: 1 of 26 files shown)


JibranKalia:
Closing this PR for now as this isn't a priority for the time being. Plus, there is a certain operational cost to keeping annotate in sync.

PR #224, opened 2021-06-10
Add ability to archive organization_users

### LINKED ISSUE #131: Feature: Archive organization users

JibranKalia:
Closes #131

### DIFF
--- a/app/controllers/organization_users_controller.rb
+++ b/app/controllers/organization_users_controller.rb
@@ -35,4 +35,10 @@ class OrganizationUsersController < ApplicationController
     authorize org_user
     org_user.update(role: params[:role])
   end
+
+  def archive
+    org_user = OrganizationUser.find params[:organization_user_id]
+    authorize org_user
+    org_user.update(status: 'archived')
+  end
 end
--- a/app/controllers/organization_users_controller.rb
+++ b/app/controllers/organization_users_controller.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class OrganizationUsersController < ApplicationController
+  before_action :authorize_org_user, except: :create
+  attr_reader :org_user
+
   def create
     authorize OrganizationUser
 
@@ -31,14 +34,17 @@ class OrganizationUsersController < ApplicationController
   end
 
   def change_role
-    org_user = OrganizationUser.find params[:organization_user_id]
-    authorize org_user
     org_user.update(role: params[:role])
   end
 
   def archive
-    org_user = OrganizationUser.find params[:organization_user_id]
-    authorize org_user
     org_user.update(status: "archived")
   end
+
+  private
+
+    def authorize_org_user
+      @org_user = OrganizationUser.find params[:organization_user_id]
+      authorize org_user
+    end
 end
--- a/app/controllers/organization_users_controller.rb
+++ b/app/controllers/organization_users_controller.rb
@@ -39,6 +39,6 @@ class OrganizationUsersController < ApplicationController
   def archive
     org_user = OrganizationUser.find params[:organization_user_id]
     authorize org_user
-    org_user.update(status: 'archived')
+    org_user.update(status: "archived")
   end
 end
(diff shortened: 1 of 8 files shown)


frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
```diff
@@ -54,20 +57,34 @@ class ManageOrgUsers extends React.Component {
...
 
+  archiveOrgUser = (id) => {
+    this.setState({ archivingUser: id });
+
+    organizationUserService
+      .archive(id)
+      .then(() => {
+        toast.success('User has been removed', { hideProgressBar: true, position: 'top-center' });
+        this.setState({ archivingUser: null });
+        this.fetchUsers();
```
JibranKalia, line 74: I am not as comfortable with React as I am with Ruby. It seems a little ugly to re-fetch the list after archiving the user. However, I am struggling to figure out how to do it. This is getting the job done.

frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
```diff
@@ -54,20 +57,34 @@ class ManageOrgUsers extends React.Component {
...
+      })
+      .catch(({ error }) => {
+        toast.error(error, { hideProgressBar: true, position: 'top-center' });
+        this.setState({ archivingUser: null });
+      });
+  };
+
   createUser = () => {
     this.setState({
-      creatingUser: true
```
JibranKalia, line 65: Ran prettier on this file before modifying it. The diff is a lot easier to read with whitespace off.

app/models/organization_user.rb
```diff
@@ -4,6 +4,8 @@ class OrganizationUser < ApplicationRecord
   belongs_to :organization
   belongs_to :user
 
+  enum status: { invited: "invited", archived: "archived" }
```
Navaneeth-pk, line 7: Shouldn't there be an `active` status? when the user accepted the invitation?

frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
Navaneeth-pk: You are right. There are a few ways to do this without an additional query. Let's create an issue for this and prioritise later?

app/controllers/organization_users_controller.rb
```diff
@@ -35,4 +35,10 @@ def change_role
     authorize org_user
     org_user.update(role: params[:role])
   end
+
+  def archive
+    org_user = OrganizationUser.find params[:organization_user_id]
+    authorize org_user
```
subintp, line 41: This is common to all controller methods. Can you move this `before_action` filter?

app/models/organization_user.rb
subintp: Key and value aresame for enum. Why are defining it as hash

You could define it like this 

` enum status: [ :active, :invited, :archived ]`

frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
JibranKalia: Absolutely. Will make an issue as soon as this is merged.

app/models/organization_user.rb
JibranKalia: So I might be missing something but when using Rails enums it uses int values. 

https://rails.rubystyle.guide/#enums

app/controllers/organization_users_controller.rb
JibranKalia: Done.

frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
```diff
@@ -54,20 +57,34 @@ class ManageOrgUsers extends React.Component {
...
       });
   };
 
+  archiveOrgUser = (id) => {
+    this.setState({ archivingUser: id });
+
+    organizationUserService
+      .archive(id)
+      .then(() => {
+        toast.success('User has been removed', { hideProgressBar: true, position: 'top-center' });
```
Navaneeth-pk, line 72: Shouldn't it be "The user has been archived"?

PR #230, opened 2021-06-10
chore(docs): minor cleanup on readme

eric-hc:
Quick fix to add a space between characters on README


Navaneeth-pk:
@eric-hc Since you've already edited Readme, can you please make one more change if that's okay? The ProductHunt badge needs to be removed from Readme, it's been a few days since the launch.

eric-hc:
Sure thing!

PR #250, opened 2021-06-14
Bugfix/separate docker file for development

### LINKED ISSUE #249: Separate dockerfile for development
### Environment
Linux mint 19.3

The app boots up normally in development using docker-compose

The rails server is unable to connect to the pg database(which is in a different container)

Start `docker-compose up` on linux and see that the server containers crashes saying that it can't connect to the database.

### Any possible solutions?
Use a different Dockerfile and hence image for starting the app in development.

I've found out that currently in the server's Dockerfile, we set `RAILS_ENV=production`. This is what's making the app unable to connect to the pg container. For some reason the RAILS_ENV=development  set as part of the docker-compose file doesn't seem to be overwriting the previous value.
Using a different Dockerfile with ` RAILS_ENV=development` fixes the issue.

Yes

ashishmax31:
Fixes #249

### DIFF
--- a/docker/server.Dockerfile.dev
+++ b/docker/server.Dockerfile.dev
@@ -9,7 +9,7 @@ RUN mkdir -p /app
 WORKDIR /app
 
 COPY Gemfile Gemfile.lock ./
-RUN gem install bundler && RAILS_ENV=production bundle install --jobs 20 --retry 5
+RUN gem install bundler && bundle install --jobs 20 --retry 5
 
 ENV RAILS_ENV=development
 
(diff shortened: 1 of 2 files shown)


aksh93:
Let's also update the docs as well

docker/server.Dockerfile.dev
```diff
@@ -0,0 +1,18 @@
...
+RUN apt update && apt install -y \
+  build-essential  \
+  postgresql \
+  freetds-dev
+
+RUN mkdir -p /app
+WORKDIR /app
+
+COPY Gemfile Gemfile.lock ./
+RUN gem install bundler && RAILS_ENV=production bundle install --jobs 20 --retry 5
```
aksh93, line 12: We can remove this `RAILS_ENV=production` right?

ashishmax31: yes, its a typo! Good catch

PR #258, opened 2021-06-14
Reset selected component state to `null` when component is deleted

prasadwalvekar:
### Changes 

Had to reset the `selectedComponent` state to `null` after the `selectedComponent` was deleted.

Loom - https://www.loom.com/share/58cbbc60b3b34be5914a737117fd2af7

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -207,7 +207,8 @@ class Editor extends React.Component {
 
   switchSidebarTab = (tabIndex) => {
     this.setState({
-      currentSidebarTab: tabIndex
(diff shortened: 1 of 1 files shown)


prasadwalvekar:
@Navaneeth-pk Can you please review?

Navaneeth-pk:
I think there is an issue here. Add two button widgets, open inspector for the first widget, and then delete the second widget. The inspector is closed then, it shouldn't happen. We will have to add a condition: only if the deleted component is the selected component, set `selectedComponent` to `null`.

prasadwalvekar:
@Navaneeth-pk Fixed.

PR #259, opened 2021-06-15
fix broken link to docs

soorajshankar:
docusaurus widgets uses kebab case, but the link was pointing to a different route


Navaneeth-pk:
`name` property of the components doesn't have spaces in it.

PR #265, opened 2021-06-16
Add a minimal delay to promise

prasadwalvekar:
### Cause 

Loading happens too fast and the skeleton gets replaced immediately with content.

### Changes

Added a function that 
1. Creates a new `Promise` that resolves after 400ms
2. Waits for both the timer Promise and network request `Promise` to resolve
3. Returns the value

This handles following scenarios -
1. The network returns the data in less than a 400ms. The combined promise will resolve when the timer fires after 400ms. As both promises need to be complete, the function will return after 400ms.
2. The network request takes exactly or more than 400ms. The combined promise will resolve when the network request returns. As the timer has already expired while waiting, the response is available immediately.

### Loom 

https://www.loom.com/share/b325a9242a144f83b404a18638adfb2b

### DIFF
--- a/frontend/src/HomePage/HomePage.jsx
+++ b/frontend/src/HomePage/HomePage.jsx
@@ -197,7 +197,7 @@ class HomePage extends React.Component {
                             }
                             </>)
                           }
-                          {currentFolder.count == 0  && (
+                          {(!isLoading && currentFolder.count === 0)  && (
                             <div>
                               <img className = "mx-auto d-block" src ="assets/images/icons/empty-folder-svgrepo-com.svg" height="120px"/>
                               <span className= "d-block text-center text-body">This folder is empty</span>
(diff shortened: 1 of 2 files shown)


Navaneeth-pk:
@prasadwalvekar 
I don't agree with slowing down the interaction intentionally. We should be aiming for quicker interactions. 

I think the flicker happens due to the difference in the height of the table in its empty state and the loading state. Let's say we set `height` or `min-height` for the table, the resizing issue won't happen and it should solve the flicker issue, right?

prasadwalvekar:
@Navaneeth-pk Setting a `min-height` was already implemented and it does not seem to be working. 

See that min-height has already been set ->

https://github.com/prasadwalvekar/ToolJet/blob/develop/frontend/src/HomePage/HomePage.jsx#L125

Navaneeth-pk:
Can you please try making the min height 900px and see if flickering issue is less/gone ?

prasadwalvekar:
@Navaneeth-pk 

> Can you please try making the min height 900px and see if flickering issue is less/gone ?

No. It does not fix the issue.

Navaneeth-pk:
Let's close this issue then, I think flicker is better than slowing down the page load :)

PR #302, opened 2021-06-23
Cypress tests forgot password

prasadwalvekar:
### Tests

1. Should display error if email is not found 
2. Should send reset password confirmation code to email


prasadwalvekar:
@Navaneeth-pk Please review.

PR #308, opened 2021-06-24
Add E2E tests for Dashboard header

prasadwalvekar:
### Tests

1. Site header is visible with nav items
2. Users tab should navigate to Users page
3. Apps tab should navigate to Apps page
4. Should show User avatar and logout the user when user clicks logout

### TODO

Cypress has a few 'recipes' for testing hidden DOM elements like in our case we have the 'logout' dropdown item but I was not able to get the 'hover' functionality working with Cypress. This will need to be adressed.


prasadwalvekar:
@Navaneeth-pk Please review.

PR #318, opened 2021-06-25
Show tooltip containing widget description when hovering over a widget

prasadwalvekar:
### Changes

- Wrapped JSX for the widget component inside `<OverlayTrigger />` component to show tooltip when user hovers over a widget.

### Screenshot

Screenshot 2021-06-25 at 6 31 43 PM

### DIFF
--- a/frontend/src/Editor/Box.jsx
+++ b/frontend/src/Editor/Box.jsx
@@ -15,8 +15,8 @@ import { Multiselect } from './Components/Multiselect';
 import { Modal } from './Components/Modal';
 import { Chart } from './Components/Chart';
 import { Map } from './Components/Map';
-import OverlayTrigger from 'react-bootstrap/esm/OverlayTrigger';
 import { renderTooltip } from '../_helpers/appUtils';
+import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
 
 const AllComponents = {
   Button,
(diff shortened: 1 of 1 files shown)


prasadwalvekar:
@Navaneeth-pk Please review.

frontend/src/Editor/Box.jsx
```diff
@@ -15,6 +15,8 @@ import { Multiselect } from './Components/Multiselect';
 import { Modal } from './Components/Modal';
 import { Chart } from './Components/Chart';
 import { Map } from './Components/Map';
+import OverlayTrigger from 'react-bootstrap/esm/OverlayTrigger';
```
Navaneeth-pk, line 18: Just curious, what is the difference between this line and `import OverlayTrigger from 'react-bootstrap/OverlayTrigger';`?

prasadwalvekar: Just noticing this as it was auto-imported by VS Code. Both are the same, the only difference being one is a CJS module (Common JS) and the other is an EcmaScript module (ESM). 
The react-bootstrap library is using multiple formats for distribution.

prasadwalvekar: Refactoring the import now to keep it consistent with other places where <OverlayTrigger /> has been imported.

PR #340, opened 2021-06-28
Write backend code for github integration

apoorv1316:
Feature: Add GitHub as a data source


Navaneeth-pk:
This feature will be handled during the migration of the server to Node.js.

PR #345, opened 2021-06-28
Rename class to className

prasadwalvekar:


prasadwalvekar:
@Navaneeth-pk Please review.

PR #346, opened 2021-06-28
Add select dropdown that shows all apps.

prasadwalvekar:
### Fixes

[Replace dynamic query input with select dropdown for 'Go to app' action](https://github.com/ToolJet/ToolJet/issues/347)

### Changes

- Add `apps: []` to editor state so that it can be passed as part of `currentState`
- Replace `<CodeHinter />` with `<SelectSearch />`

### Loom

https://www.loom.com/share/416ef743d14d4dd9b7656dadd812d8cb

### DIFF
--- a/frontend/src/Editor/Inspector/EventSelector.jsx
+++ b/frontend/src/Editor/Inspector/EventSelector.jsx
@@ -129,9 +129,9 @@ export const EventSelector = ({
                   <SelectSearch
                     options={getAllApps()}
                     search={true}
-                    value={definition.options.appId}
+                    value={definition.options.slug}
                     onChange={(value) => {
-                      eventOptionUpdated(param, 'appId', value, extraData);
+                      eventOptionUpdated(param, 'slug', value, extraData);
                     }}
                     filterOptions={fuzzySearch}
                     placeholder="Select.."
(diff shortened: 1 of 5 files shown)


prasadwalvekar:
@Navaneeth-pk Please review.

frontend/src/Editor/Inspector/EventSelector.jsx
```diff
@@ -115,10 +126,15 @@ export const EventSelector = ({
...
-                  <CodeHinter
-                    currentState={currentState}
-                    initialValue={definition.options.appId}
-                    onChange={(value) => eventOptionUpdated(param, 'appId', value, extraData)}
+                  <SelectSearch
+                    options={getAllApps()}
+                    search={true}
+                    value={definition.options.appId}
+                    onChange={(value) => {
+                      eventOptionUpdated(param, 'appId', value, extraData);
```
Navaneeth-pk, line 134: It should be `slug` instead of `appId` right? I think I made that mistake while building the go to app action. Can you please clean that up in this PR?

prasadwalvekar: Done.

Issue #312, opened 2021-06-25
Feature: Delete rows from Google sheets


Navaneeth-pk:
Description: We need to add a new operation for the Google sheets query editor to delete a row from a sheet. 

API reference: https://developers.google.com/sheets/api/samples/rowcolumn#delete_rows_or_columns

Options needed for the operation: 
a) Sheet ID 
b) Sheet name 
c) Row Index 

The row at the given index should be deleted from the sheet.

nisusam:
Hey @Navaneeth-pk I'll pick this up

PR #354, opened 2021-06-29
Add feature `delete row from Google sheets`

nisusam:
- Fix https://github.com/ToolJet/ToolJet/issues/312


frontend/src/Editor/QueryManager/QueryEditors/Googlesheets.jsx
```diff
@@ -117,6 +118,24 @@ class Googlesheets extends React.Component {
           </div>
       )}
 
+      {this.state.options.operation === 'delete' && (
```
Navaneeth-pk, line 121: id the of the oepration should be `delete_row` instead of `delete` so that it won't be confusing when we add the 'delete spreadsheet' operation.

nisusam: Thanks @Navaneeth-pk  for pointing this out.

PR #356, opened 2021-06-29
Fixes share button failing to load modal intermittently

### LINKED ISSUE #305: Share and manage users modal fails to load intermittently on app editor page
### Environment
Any

Share and manage users modal on editor page should not fail to load

The modal fails to load intermittently due to API call to the server being failed

1. Click on edit app from homepage listing
2. Check at console logs if xhr failure is present to the server to fetch app users
3. Click on share button to observe it not being loaded

### Screenshots or Screencast
image
image

### Any possible solutions?
Render the modal only if app id is present

https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/Editor/Editor.jsx#L556

Yes

aksh93:
Fixes #305 

The component `ManageAppUsers` fetches users based on the app id but currently, we don't check explicitly whether app id is present before rendering the component

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -52,6 +52,7 @@ class Editor extends React.Component {
 
     this.state = {
       currentUser: authenticationService.currentUserValue,
+      app: {},
       allComponentTypes: componentTypes,
       queryPaneHeight: '30%',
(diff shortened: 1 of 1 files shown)


Navaneeth-pk (approved):
looks good to me.

Issue #177, opened 2021-05-28
User role drop down not collapsing properly


jithindasad:
**Description:**
User role's drop down menu can be enhanced.

**Refer:**
![IMAGE 2021-05-28 18:07:33](https://user-images.githubusercontent.com/26083775/119984884-93032080-bfdf-11eb-9ddb-67ef66981a0a.jpg)

virajbahulkar:
I will work on it.

PR #382, opened 2021-07-06
Fix user role dropdown not collapsing

virajbahulkar:
### Issue - [User role drop down not collapsing properly #177](https://github.com/ToolJet/ToolJet/issues/177)

Loom - 

https://www.loom.com/share/c742e2734d08431b959bd05ad578cfe5

### DIFF
--- a/frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
+++ b/frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
@@ -220,7 +220,7 @@ class ManageOrgUsers extends React.Component {
             {!showNewUserForm && (
               <div className="container-xl">
                 <div className="card">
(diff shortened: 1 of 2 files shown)


frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
```diff
@@ -220,7 +220,7 @@ class ManageOrgUsers extends React.Component {
             {!showNewUserForm && (
               <div className="container-xl">
                 <div className="card">
-                  <div className="card-table table-responsive table-bordered">
+                  <div className="card-table table-responsive table-bordered userTable">
```
Navaneeth-pk, line 223: please use `user-table` instead of `userTable` to be consistent with other class names.

Issue #389, opened 2021-07-09
Version saving status is showing on all 'save' buttons


Navaneeth-pk:
If an app is having multiple versions, all the `save` buttons show the `saving` status when one button is clicked. 

https://www.loom.com/share/8fd38d92f3c848e3863144a2660189fc

keshavbiswa:
@Navaneeth-pk you can assign this to me

PR #390, opened 2021-07-09
Fix version saving status is showing on all 'save' buttons

keshavbiswa:
Fixes: #389

### DIFF
--- a/frontend/src/Editor/SaveAndPreview.jsx
+++ b/frontend/src/Editor/SaveAndPreview.jsx
@@ -52,18 +52,18 @@ class SaveAndPreview extends React.Component {
   };
 
   saveVersion = (versionId) => {
-    this.setState({ isSaving: versionId });
+    this.setState({ savingVersionId: versionId });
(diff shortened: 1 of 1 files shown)


frontend/src/Editor/SaveAndPreview.jsx
```diff
@@ -52,18 +52,18 @@ class SaveAndPreview extends React.Component {
   };
 
   saveVersion = (versionId) => {
-    this.setState({ isSaving: true });
+    this.setState({ isSaving: versionId });
```
Navaneeth-pk, line 55: the name `isSaving` signifies a boolean, can we change the name of the variable to something like `savingVersionId` or `savingVersion`?

Same suggestion for `isDeploying`

keshavbiswa: Sure will make the change right away @Navaneeth-pk

Navaneeth-pk (approved):

Issue #311, opened 2021-06-25
Feature: Support for modifying data on Google sheets


Navaneeth-pk:
ToolJet can read data and add new rows, we need to add the ability to update the data in a sheet.

Sheets API Reference: https://developers.google.com/sheets/api/samples/writing

Name of the operation: 'Update row'

Required options for the update query: 
a) Sheet ID - Id of the Google sheet
b) Sheet name - name of the sheet
c) Row Index - index of the row that needs to be changed
d) Values - JSON object. For example if the sheet headers have name, email and id, and if we need to update name and email, the JSON object will look like: 

```json
{ 
 "name": "new name",
 "email": "new email
}
```

nisusam:
Hey @Navaneeth-pk I'll pick this up

PR #395, opened 2021-07-13
Update google sheet row

alkesh26:
Fixes: #311

### DIFF
--- a/frontend/src/Editor/QueryManager/QueryEditors/Googlesheets.jsx
+++ b/frontend/src/Editor/QueryManager/QueryEditors/Googlesheets.jsx
@@ -47,6 +47,10 @@ class Googlesheets extends React.Component {
     );
   };
 
+  changeRowDetails = () => {
+
+  }
+
   render() {
     const { options } = this.props;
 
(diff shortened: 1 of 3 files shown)


frontend/src/Editor/QueryManager/QueryEditors/Googlesheets.jsx
```diff
@@ -47,6 +47,10 @@ class Googlesheets extends React.Component {
     );
   };
 
+  changeRowDetails = () => {
```
Navaneeth-pk, line 50: We can remove this, right?

frontend/src/Editor/QueryManager/QueryEditors/Googlesheets.jsx
```diff
@@ -136,6 +141,35 @@ class Googlesheets extends React.Component {
         </div>
       )}
 
+      {this.state.options.operation === 'update_row' && (
+        <React.Fragment>
+          <div className="mb-3 mt-2 row">
+            <div className="col">
+              <label className="form-label">Update row number</label>
```
Navaneeth-pk, line 148: `Update row number` should be `Row index`

Navaneeth-pk:
This feature will be handled during the migration of server to JS.

PR #405, opened 2021-07-18
Sets node and npm version

aksh93:
Makes strict versioning for node and npm to be used


server/.node-version
```diff
@@ -0,0 +1 @@
+16.4.2
```
vipulnsward, line 1: Can prefer LTS instead of this?

Navaneeth-pk: `Node.js 16 will be the ‘Current’ release for the next 6 months and then promoted to Long-term Support (LTS) in October 2021`

Wouldn't 16 be better or am I missing something? 🤔

Navaneeth-pk: `Node.js 14 will remain in long-term support until April 2023` - 14 won't be an issue too.

aksh93: I just used the latest stable and had not checked for LTS. Agree, makes sense to stick with version with LTS until.

Issue #398, opened 2021-07-15
Cypress tests for empty state of dashboard


mohinid:
Write Cypress tests for dashboard when there is no existing App, as here you get different view as compared to the case where the user has already created at-least 1 App.
    - Test 'Create your first App' button.
    - Test 'Read Document' button

mohinid:
Fixed in PR #409

PR #407, opened 2021-07-19
Cypress test for empty state of dashboard

mohinid:
All changes are related to issue #398 :
1. Modified all test descriptions in 'dashboard.spec.js' file.
2. Added pre-condition in 'dashboard.spec.js' file to test whether the dashboard is empty. If it's empty then create an app first and then proceed with the tests.
3. Added new test file 'empty-state-dashboard.spec.js'. It tests the options present on the screen when there are no apps.
4. Updated 'cypress.json' file with 'testFiles' which will run tests in the given order.


mohinid:
opening new as I don't want 'Test try git' commit in the PR.

PR #411, opened 2021-07-19
Environment based configuration setup

aksh93:
Changes
- Based on the NODE_ENV set, env files will be picked up from root folder
   - production | development -> `.env`
   -  test -> `.env.test`
- Makes use of ConfigModule to setup config within app from .env file
- Adds custom scripts since typerorm doesn't currently handle [db creation](https://github.com/typeorm/typeorm/issues/809
). Can now do: 
    - `NODE_ENV=enviroment npm run db:create` 
    - `NODE_ENV=enviroment npm run db:migrate` 
    - `NODE_ENV=enviroment npm run db:drop` 
    - `NODE_ENV=enviroment npm run db:reset`


server/scripts/create-database.ts
```diff
@@ -0,0 +1,51 @@
...
+  exec('command -v createdb', (err, stdout, stderr) => {
+    if (err) {
+      console.error(err);
+      return;
+    }
+
+    const createdb =
+      `PGPASSWORD=${process.env.PG_PASS} createdb ` +
+      `-h ${process.env.PG_HOST} ` +
+      `-p ${process.env.PG_PORT} ` +
```
Navaneeth-pk, line 22: Shouldn't there be defaults? `process.env.PG_HOST || 5432` ?

aksh93: Ah yes, missed adding. Will update.

server/ormconfig.ts
```diff
@@ -0,0 +1,54 @@
...
+    type: 'postgres',
+    database: databaseName,
+    port: data.PG_PORT || 5432,
+    username: data.PG_USER,
+    password: data.PG_PASS,
+    host: data.PG_HOST,
+    entities: ['dist/**/*.entity{.ts,.js}'],
+    synchronize: false,
+    migrationsRun: false,
+    logging: data.ORM_LOGGING || false,
```
Navaneeth-pk, line 22: `data.ORM_LOGGING` will be a string, right?

aksh93: Yeah, so logger options [accepts](https://github.com/typeorm/typeorm/blob/master/src/logger/LoggerOptions.ts#L4) either a string or boolean. We'd either specify in those options else default as false. Wdyt?

Navaneeth-pk: Sounds good 👍

PR #412, opened 2021-07-20
chart legend color fix

ajinkyaa:
https://github.com/ToolJet/ToolJet/issues/413


ajinkyaa:
After changes in legend color.
Screenshot 2021-07-20 at 2 13 33 PM

PR #426, opened 2021-07-29
Switch DarkMode Styles according to localStorage.getItem('darkMode')

### LINKED ISSUE #400: Dark mode for map widget
Maps widget should switch to dark styles if in the dark mode. 

Reference: https://developers.google.com/maps/documentation/javascript/examples/style-array#maps_style_array-javascript

xrimpy:
Resource for Styling https://snazzymaps.com/

Fixes #400

### DIFF
--- a/frontend/src/Editor/Components/Map.jsx
+++ b/frontend/src/Editor/Components/Map.jsx
@@ -44,6 +44,88 @@ export const Map = function Map({
     height
   };
 
+  const darkModeStyles = [
+    { elementType: "geometry", stylers: [{ color: "#242f3e" }] },
+    { elementType: "labels.text.stroke", stylers: [{ color: "#242f3e" }] },
+    { elementType: "labels.text.fill", stylers: [{ color: "#746855" }] },
+    {
+      featureType: "administrative.locality",
+      elementType: "labels.text.fill",
+      stylers: [{ color: "#d59563" }],
+    },
+    {
+      featureType: "poi",
+      elementType: "labels.text.fill",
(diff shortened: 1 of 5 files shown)


frontend/src/Editor/Components/Map.jsx
```diff
@@ -110,6 +192,7 @@ export const Map = function Map({
           mapContainerStyle={containerStyle}
           zoom={12}
           options={{
+            styles: localStorage.getItem('darkMode') === 'true' ? darkModeStyles : '',
```
Navaneeth-pk, line 195: Every widget will get `darkMode` as a parameter, we don't need to fetch from local storage. Can you please check `frontend/src/Editor/Components/Chart.jsx` and implement this logic in a similar way?

frontend/src/Editor/Components/Map/Map.jsx
```diff
@@ -44,6 +45,8 @@ export const Map = function Map({
     height
   };
 
+  const darkModeStyles = require("./darkModeStyles.json");
```
Navaneeth-pk, line 48: Can we move the imports to the top of the file?

xrimpy: Extract it out as a JS array as in this example https://www.pluralsight.com/guides/load-and-render-json-data-into-react-components

frontend/src/Editor/Components/Map.jsx
xrimpy: Using  Prop

PR #443, opened 2021-08-03
Database seed setup

aksh93:
- do `npm run db:seed` to create default user
- skips seed if default user already present
- run seed on heroku release


server/scripts/seeds.ts
```diff
@@ -0,0 +1,15 @@
+import { NestFactory } from '@nestjs/core';
+import { AppModule } from '../src/app.module';
+import { SeedsService } from '@services/seeds.service';
+
+async function bootstrap() {
```
aksh93, line 5: Loading the seed service in the application context 
https://docs.nestjs.com/standalone-applications

PR #452, opened 2021-08-03
Refactor google sheets operations for cell range and header check

aksh93:


server/plugins/datasources/googlesheets/operations.ts
```diff
@@ -1,22 +1,53 @@
 import got from 'got';
 
-export async function readDataFromSheet(
+async function googleSheetGetValues(
```
Navaneeth-pk, line 3: Will it be better to name it `getValuesFromSheet`?

aksh93: Agreed. Have updated to make it clearer.

PR #463, opened 2021-08-05
Packer changes for Nestjs

akshaysasidrn:
- Removes rails related setup
- Moves app build within image building step
- Update nginx config for single path as nest will serve frontend as well

### DIFF
--- a/deploy/ec2/setup_app
+++ b/deploy/ec2/setup_app
@@ -1,56 +1,37 @@
-#!/usr/bin/env ruby
+#!/bin/bash
 
-def ensure_db_connectivity(user = ENV.fetch("PG_USER"), pass = ENV.fetch("PG_PASS"), host = ENV.fetch("PG_HOST"))
-  cmd = %{psql -d 'postgresql://#{user}:#{pass}@#{host}' -c 'select now()' > /dev/null 2>&1}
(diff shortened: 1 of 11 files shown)


Navaneeth-pk (approved):
LGTM 👍

deploy/ec2/setup_app
```diff
@@ -1,56 +1,30 @@
...
-ensure_db_connectivity
-install_be_app_deps
-build_fe
-start_services
-
-
-puts "The app will be served at #{ENV.fetch("TOOLJET_HOST")}"
+#!/bin/bash
+
+export $(grep -v '^#' .env | xargs)
```
ashishmax31, line 3: It would be nice to have a check looking for the required env variables for the app to start. If not, throw a message and exit ?

ashishmax31:
Looks good otherwise! Good job @aksh93 :+1:

deploy/ec2/setup_app
aksh93: Yes, should be good to have that. Let me add it as well.

aksh93: had added @ashishmax31

ashishmax31 (approved):
lgtm

Issue #460, opened 2021-08-05
Add update records operation for Airtable


arpitnath:
**Feature Request**: Add a new operation for Airtable integration - update records

arpitnath:
taking this up

PR #468, opened 2021-08-07
Feature/update record operation airtable

arpitnath:
**fixes** : #460 

Update operation for Airtable records

**Loom** :
https://www.loom.com/share/f9b996449cb247a385880405e1b6e4ed

https://tooljet-update-pr.herokuapp.com/

**Screenshot** : `CodeHinter`
Screenshot 2021-08-12 at 6 18 00 PM
Screenshot 2021-08-12 at 6 22 10 PM


Navaneeth-pk:
image

The theme of CodeHinter is not consistent with the other queries. For example, the field for postgres query looks like this: 

image

server/plugins/datasources/airtable/index.ts
```diff
@@ -44,6 +44,25 @@ export default class AirtableQueryService implements QueryService {
...
+          response = await got(`https://api.airtable.com/v0/${baseId}/${tableName}`, { 
+            method: 'patch',   
+            headers: this.authHeader(accessToken),
+            json: {
+              "records": [
+                {
+                   "id": queryOptions['record_id'],
+                   "fields": JSON.parse(queryOptions['body'])
+               }
+           ]
```
Navaneeth-pk, line 58: looks like the code is not indented properly here.

arpitnath:
looking into it

PR #495, opened 2021-08-11
Enable csp directives to allow script load when nest serves frontend

akshaysasidrn:


akshaysasidrn:
This is to enable current heroku deployments. Have opened an [issue](https://github.com/ToolJet/ToolJet/issues/496) to follow up on to improve this setting.

PR #501, opened 2021-08-12
Feature/profile page

### LINKED ISSUE #491: Profile page where the user can change name, password etc
I'm working on this.

sherfin94:
Resolves #491


sherfin94:
https://user-images.githubusercontent.com/20333972/129183555-932b3dde-43fb-47a4-95ce-242445433d03.mp4

server/src/modules/auth/password-revalidate.guard.ts
```diff
@@ -0,0 +1,29 @@
...
+const bcrypt = require('bcrypt')
+
+@Injectable()
+export class PasswordRevalidateGuard implements CanActivate {
+
+  constructor(
+    private usersService: UsersService
+  ) {}
+
+  async validateUser(email: string, password: string): Promise<boolean> {
```
Navaneeth-pk, line 13: Can we re-use the `validateUser` method from auth service?

sherfin94: I considered this, however, the `validateUser` method comes from `AuthModule`, which references `UserModule`, creating a circular dependency. There are ways to circumvent this, however I felt it'd be less complicated to define another `validateUser` method, plus, the method I defined is slightly customized for the requirement at hand.

PR #507, opened 2021-08-13
Fixes docker dev setup

### LINKED ISSUE #504: Fix docker-compose dev setup
### Environment
Docker based dev setup

Should be able to setup dev environment seamlessly following the docs

Following docs for docker dev setup results in error when doing `docker-compose build`

Yes

akshaysasidrn:
Fixes #504


docker-compose.yaml
```diff
@@ -29,13 +28,6 @@ services:
       - /app/node_modules/
     ports:
       - 3000:3000
-    environment:
```
akshaysasidrn, line 32: Values will be picked from `.env` files instead

docs/docs/contributing-guide/setup/docker.md
```diff
@@ -42,19 +43,43 @@ We recommend:
...
+    ORM_LOGGING=error
    ```
 
 4. Build docker images
    ```bash
     docker-compose build
    ```
 
-5. ToolJet server is built using NestJS and the data such as application definitions are persisted on a postgres database. You have to reset the database if building for the first time.
+5. ToolJet server is built using NestJS and the data such as application definitions are persisted on a postgres database. You have to create and migrate the database if building for the first time.
```
akshaysasidrn, line 79: Reverted to use create and migrate explicitly to make intent clear

server/ormconfig.ts
```diff
@@ -35,7 +35,7 @@ function buildConnectionOptions(
     uuidExtension: 'pgcrypto',
     migrationsRun: false,
     logging: data.ORM_LOGGING || false,
-    migrations: ['dist/migrations/**/*{.ts,.js}'],
```
akshaysasidrn, line 38: It Will only pick up if `dist` is generated which won't be the case when setting up the project

PR #516, opened 2021-08-13
[ADD] enable pgcrypto extension

vimoxshah:
This fix handles the gen_random_uuid() when developers run the migration first time. otherwise, Postgres might throw an error saying that `QueryFailedError: function gen_random_uuid() does not exist`

Screenshot 2021-08-13 at 4 58 52 PM

### DIFF
--- a/server/migrations/1625814801416-CreateOrganizations.ts
+++ b/server/migrations/1625814801416-CreateOrganizations.ts
@@ -3,6 +3,9 @@ import { MigrationInterface, QueryRunner, Table, TableIndex } from "typeorm";
 export class CreateOrganizations1625814801416 implements MigrationInterface {
 
   public async up(queryRunner: QueryRunner): Promise<void> {
+    
+    await queryRunner.query('CREATE EXTENSION IF NOT EXISTS pgcrypto;')
+
     await queryRunner.createTable(new Table({
       name: "organizations",
       columns: [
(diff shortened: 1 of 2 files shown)


akshaysasidrn:
@vimoxshah I'm unable to replicate this issue at local. Which postgres version is being used here?
Think typeorm does install extensions by [default](https://github.com/typeorm/typeorm/blob/master/docs/connection-options.md#postgres--cockroachdb-connection-options) 
https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresDriver.ts#L341

vimoxshah:
Hey, @akshaysasidrn I am using Postgresql version 11.12 and macOS 11.5.2. I was facing this issue in my local

akshaysasidrn:
I suppose postgres prior to v13 is not handled. Let's add support for earlier version as well then. ✌️
Could help add this as a separate migration prior to the existing one? Like a zeroth migration `0-MaybeCreateExtension.ts` ?

vimoxshah:
I'll do it. Thanks

vimoxshah:
Hey, @akshaysasidrn  I have made the changes as per your request.

server/migrations/1625814801416-CreateOrganizations.ts
```diff
@@ -3,6 +3,8 @@ import { MigrationInterface, QueryRunner, Table, TableIndex } from "typeorm";
 export class CreateOrganizations1625814801416 implements MigrationInterface {
 
   public async up(queryRunner: QueryRunner): Promise<void> {
+    
+
```
Navaneeth-pk, line 7: Can you please remove the additional line break here?

vimoxshah: @Navaneeth-pk  Done

PR #519, opened 2021-08-16
Feature/qr reader

### LINKED ISSUE #509: QR Reader widget
We need a widget that can

- Use the user's camera to scan and detect QR codes
- Store the data contained in the QR and make it available to be referred from elsewhere

sherfin94:
Resolves #509 

https://www.loom.com/share/cb9d2919420e4c9081d7ee2c981cfe1e


docs/docs/widgets/qr-scanner.md
```diff
@@ -0,0 +1,28 @@
...
+
+# QR Scanner
+Scan QR codes using device camera and hold the data they carry. 
+<img class="screenshot-full" src="/img/widgets/qr-scanner/qr-scanner.jpeg" alt="ToolJet - QR Scanner" height="420"/>
+
+#### Known issue: 
+In IOS, you might have to stick to the Safari browser as camera access had been restricted for third party browsers.
+
+## Properties
+#### active
```
Navaneeth-pk, line 13: Are we hiding the widget when `active` is resolved as `false` or we will still show the widget in a disabled state?

sherfin94: Right now it wont render anything if `active` resolves to false. I'm waiting to learn from the scenarios on which the `active` property will be made use of.

PR #545, opened 2021-08-18
Fix the invite new user button bug

### LINKED ISSUE #467: 'Invite new user' button is unnecessarily displayed on multiple scenarios
**Scenario 1**: Once we click on 'Invite new user' from users page, the 'Invite new user' button still remains there without purpose.

image
 
**Scenario 2**: If my user role is 'Viewer' then I don't have the permission to invite new users. Yet, the 'Invite new user' button is still displayed for me, when I log in as a Viewer.

apoorv1316:
Fixes #467 

- Fixed Scenario 1 - The invite new user button is not visible as soon as the user clicks on the button.
- Fixed Scenario 2 -  The button is not displayed if the user is not admin.
https://www.loom.com/share/acd00949bc964bfc9b28e644b9f671f6

### DIFF
--- a/frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
+++ b/frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
@@ -20,6 +20,7 @@ class ManageOrgUsers extends React.Component {
       newUser: {},
       idChangingRole: null,
       archivingUser: null,
+      currentUserDetails: null,
     };
   }
 
@@ -36,6 +37,9 @@ class ManageOrgUsers extends React.Component {
       this.setState({
         users: data.users,
         isLoading: false,
+        currentUserDetails: data.users.find(
+          (user) => user.role == 'admin' && user.email == this.state.currentUser.email
+        ),
       })
     );
   };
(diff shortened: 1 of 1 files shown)


apoorv1316:
@Navaneeth-pk Please have a look. Here is the deployed link https://invitebuttonbug.herokuapp.com/

frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
```diff
@@ -128,9 +140,14 @@ class ManageOrgUsers extends React.Component {
                   <h2 className="page-title">Users & Permissions</h2>
                 </div>
                 <div className="col-auto ms-auto d-print-none">
-                  <div className="btn btn-primary" onClick={() => this.setState({ showNewUserForm: true })}>
-                    Invite new user
-                  </div>
+                  {!showNewUserForm &&
+                    currentUser &&
+                    currentUserDetails &&
+                    currentUserDetails.email == currentUser.email && (
```
Navaneeth-pk, line 146: what does this condition do? it will always be `true` right?

frontend/src/ManageOrgUsers/ManageOrgUsers.jsx
```diff
@@ -36,6 +37,9 @@ class ManageOrgUsers extends React.Component {
       this.setState({
         users: data.users,
         isLoading: false,
+        currentUserDetails: data.users.find(
```
Navaneeth-pk, line 40: Works only if we fetch all the users in one request. Once we add pagination to this page, it won't work. This solution cannot be accepted.

PR #560, opened 2021-08-20
Add query params option to Go-to-app action

sherfin94:
Adds support for setting query params for the
go-to-app action. When the new app is launched, the passed in
object will be sent as search query params.

https://www.loom.com/share/e0dcac7bcad044a68fa34d8e77717db6

### DIFF
--- a/frontend/src/Editor/ActionTypes.js
+++ b/frontend/src/Editor/ActionTypes.js
@@ -24,7 +24,8 @@ export const ActionTypes = [
     name: 'Go to app',
     id: 'go-to-app',
     options: [
-      { name: 'app', type: 'url', default: 'https://app.tooljet.io/applications/app-id' }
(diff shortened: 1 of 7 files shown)


sherfin94:
Resolves #560

frontend/src/Editor/ActionTypes.js
```diff
@@ -24,7 +24,8 @@ export const ActionTypes = [
     name: 'Go to app',
     id: 'go-to-app',
     options: [
-      { name: 'app', type: 'url', default: 'https://app.tooljet.io/applications/app-id' }
+      { name: 'app', type: 'url', default: 'https://app.tooljet.io/applications/app-id' },
```
Navaneeth-pk, line 27: `url` is not needed, right?

sherfin94: Correct, changed it to `text` and defaulted to `''`.

PR #599, opened 2021-08-24
Fix incorrect tooltip when query editor hidden

### LINKED ISSUE #576: Incorrect tooltip when query editor is hidden
Click on the button to hide the query editor that is present of header currently 

image

On the query editor is hidden and hide/show query editor button is left in hovered state the tooltip must change to `show query editor`

image

apoorv1316:
Fixes #576

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { createRef } from 'react';
 import { datasourceService, dataqueryService, appService, authenticationService } from '@/_services';
 // import { DarkModeToggle } from '@/_components/DarkModeToggle';
 import { DndProvider } from 'react-dnd';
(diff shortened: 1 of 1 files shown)


Navaneeth-pk:
@apoorv1316 
Can you please check why the tooltip for hiding query editor is larger than the other tooltips? ( for example, the tooltip for toggling dark mode ).

apoorv1316:
@Navaneeth-pk I believe it is because we are using different packages at different places for the tooltip. The size of 'hiding query editor' and 'add new query' tooltip is the same because we are using react-tooltip there and their sizes differ from the tooltip for toggling dark mode because we are using react-bootstrap/Tooltip.

Navaneeth-pk:
Can you please check if it can be fixed easily?

apoorv1316:
Here is the updated deployed url https://tooltip-query-editor.herokuapp.com/

Navaneeth-pk:
The size of the button now looks bigger than it was. Please fix.

apoorv1316:
@Navaneeth-pk  I am guessing here you're talking about the hide query editor button. The first image is from develop branch and the second one is from the deployed URL of this PR and buttons in both the images are equal in size. 
imgonline-com-ua-twotoone-q020sqdi4S1N

Navaneeth-pk:
Sorry, looks like I got it wrong somehow. Merging the PR. Thanks!
@apoorv1316

Navaneeth-pk:
@apoorv1316 figured it out, it happens when the button is clicked. See the attached video. It doesn't happen on production.
https://www.loom.com/share/942282b39fcc4700b873f9edb07673e3

PR #600, opened 2021-08-25
Feature/application logger

### LINKED ISSUE #522: Improve application logging for production
Currently, logging is very basic and doesn't help in tracking down or filtering requests through logs.
Make use of Winston (suggested at Nestjs docs) or any other logging library to be added at our application middleware.  
https://docs.nestjs.com/techniques/logger#use-external-logger

akshaysasidrn:
- Makes use of [pino](https://github.com/pinojs/pino/blob/master/docs/web.md#pino-with-nest) for logging

fixes #522


server/package.json
```diff
@@ -83,6 +84,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^3.4.0",
     "jest": "^27.0.6",
+    "pino-pretty": "^6.0.0",
```
Navaneeth-pk, line 87: Should we add `https://www.npmjs.com/package/@types/pino` too?

akshaysasidrn: It's already added as a part of [dependency](https://github.com/ToolJet/ToolJet/pull/600/files#diff-be330df323f72d45d073e311befe92984fad9846bb7e982261d702d805c14053R16260) from `nestjs-pino`

Issue #608, opened 2021-08-26
Feature: Badge/Indicator for a new error log entry


arpitnath:
### Description
Debugger puts a red dot with a number of unread logs next to its icon whenever there is a new error log entry.

PR #616, opened 2021-08-26
Feature: notification badge-indicator [debugger]

arpitnath:
**Fixes**: #608 

**Loom**: https://www.loom.com/share/d5c9ea5a713a422eb2db63e3948c418e

**app**: https://tooljet-update-pr.herokuapp.com/apps/b945fce6-1272-4468-b5c0-1ec3c875255d


Navaneeth-pk:
@arpitnath looks good, please resolve the merge conflicts.

gondar00:
One suggestion, the notification badge ideally should not cut out of the sidebar

arpitnath:
Should I reduce the size? or change the position?

gondar00:
Also, can you add a check if the count increases 1000 then show something like 999+, to handle this edge case 
ref: 
image

gondar00:
> Should I reduce the size? or change the position?

I think height 20px width 20px should be good, the trick would be to reduce the font-weight property to something like 400

arpitnath:
> Also, can you add a check if the count increases 1000 then show something like 999+, to handle this edge case
> ref:
> image

yes, also if the number of errors increases 100 we can paginate the logs

frontend/src/Editor/LeftSidebar/SidebarDebugger.jsx
```diff
@@ -56,20 +57,41 @@ export const LeftSidebarDebugger = ({ darkMode, components, errors }) => {
...
+
+        if(open ===  false && errorLogs.length !== unReadErrorCount.read) {
+            setUnReadErrorCount((prev) => {
+                let copy = JSON.parse(JSON.stringify(prev))
+                copy.unread = unReadErrors
+                return copy
+            })
+        }
+        
+        if(open === true) {
```
gondar00, line 71: I don't think a check is required here it can also be written as `if(open)`

arpitnath:
> > Should I reduce the size? or change the position?
> 
> I think reduce the size to height 20px width 20px should be good, the trick would be to reduce the font-weight property to something like 400

cool, will update in this PR

frontend/src/Editor/LeftSidebar/SidebarDebugger.jsx
```diff
@@ -56,20 +57,41 @@ export const LeftSidebarDebugger = ({ darkMode, components, errors }) => {
...
+    React.useEffect(() => {
+        const unReadErrors = errorLogs.length - unReadErrorCount.read
+
+        if(open ===  false && errorLogs.length !== unReadErrorCount.read) {
+            setUnReadErrorCount((prev) => {
+                let copy = JSON.parse(JSON.stringify(prev))
+                copy.unread = unReadErrors
+                return copy
+            })
+        }
```
gondar00, line 69: `unReadErrors` should be declared inside the if statement where its being used

arpitnath: yes, i was using it in the other block too, but later changed and I guess did not notice, thanks

frontend/src/Editor/LeftSidebar/SidebarDebugger.jsx
arpitnath: Yep, we don't need this, just can put it in the else block and its all good

arpitnath:
Screenshot 2021-08-28 at 4 34 59 PM
Screenshot 2021-08-28 at 4 34 53 PM

arpitnath:
https://www.loom.com/share/e9aca92a8e1c4560aca77e73ab3ef5bc

PR #622, opened 2021-08-27
Fix ESLint

arihant-2310:


akshaysasidrn:
Currently, we don't have a pipeline setup to ensure for all the changes that are merged. Will be doing that afterwards and fix whole of it in one PR. So closing this, thanks.

Issue #644, opened 2021-08-31
Feature: ability to disable widgets


arpitnath:

arpitnath:
working on it

PR #653, opened 2021-08-31
Feature: disable widgets

arpitnath:
**Fixes**: #644 

**App**: https://tooljet-pr-app.herokuapp.com/apps/88112eec-3c96-4ff9-9936-8ceefa9d6814


frontend/src/Editor/Components/Button.jsx
```diff
@@ -21,7 +21,9 @@ export const Button = function Button({
   const backgroundColor = component.definition.styles.backgroundColor.value;
   const color = component.definition.styles.textColor.value;
   const widgetVisibility = component.definition.styles?.visibility?.value || true;
```
gondar00, line 23: `false || true` will always return true 
```suggestion
  const widgetVisibility = component.definition.styles?.visibility?.value ?? true;
```

frontend/src/Editor/Components/Map/Map.jsx
```diff
@@ -107,7 +110,7 @@ export const Map = function Map({
 
 
   return (
-    <div style={{ width, height, display:parsedWidgetVisibility ? '' : 'none' }} onClick={event => {event.stopPropagation(); onComponentClick(id, component)}} className="map-widget">
+    <div disabled={parsedDisableState} style={{ width, height, display:parsedWidgetVisibility ? '' : 'none' }} onClick={event => {event.stopPropagation(); onComponentClick(id, component)}} className="map-widget">
```
gondar00, line 113: <div> elements do not have a disabled attribute

frontend/src/Editor/Components/Button.jsx
arpitnath: yes, I missed it, will fix it.

frontend/src/Editor/Components/Map/Map.jsx
arpitnath: yes, I know. setting it as a new attribute for the div so that can use it as selectors.
Do you have a better way of this?

gondar00: Yes, we can use `data-*` attributes

more info: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

arpitnath: thanks 😁

arpitnath:
@gondar00 up for review

gondar00 (approved):

PR #671, opened 2021-09-01
Customisable font color for string cells

### LINKED ISSUE #670: Customise font color of string table cells
The color of text ina. string type table cell should be customisable. To achieve this feature, the value of the cell should be made available as a dynamic variable.

Navaneeth-pk:
Fixes #670

### DIFF
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -276,10 +276,17 @@ export function Table({
         const cellValue = rowChangeSet ? rowChangeSet[column.name] || cell.value : cell.value;
 
         if (columnType === 'string' || columnType === undefined || columnType === 'default') {
+          
+          const textColor = resolveReferences(column.textColor, currentState, { cellValue }) || [];
+          const cellStyles = {
+            color: textColor
+          }
+
           if (column.isEditable) {
             return (
               <input
                 type="text"
+                style={cellStyles}
                 onKeyDown={(e) => {
                   if (e.key === 'Enter') {
                     if(e.target.defaultValue !== e.target.value) {
@@ -297,7 +304,7 @@ export function Table({
               />
             );
           }
-          return <span>{cellValue}</span>;
+          return <span style={cellStyles}>{cellValue}</span>;
         } if (columnType === 'text') {
           return <textarea 
               rows="1" 
@@ -682,9 +689,9 @@ export function Table({
                       if (componentState.changeSet) {
                         if (componentState.changeSet[cell.row.index]) {
 
-                          const accessor = columnData.find(column => column.id === cell.column.id)?.accessor;
+                          const currentColumn = columnData.find(column => column.id === cell.column.id);
 
-                          if (_.get(componentState.changeSet[cell.row.index], accessor, undefined) !== undefined) {
+                          if (_.get(componentState.changeSet[cell.row.index], currentColumn?.accessor, undefined) !== undefined) {
                             console.log('componentState.changeSet', componentState.changeSet);
                             cellProps.style.backgroundColor = '#ffffde';
                           }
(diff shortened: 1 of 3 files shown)


frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -276,10 +276,17 @@ export function Table({
         const cellValue = rowChangeSet ? rowChangeSet[column.name] || cell.value : cell.value;
 
         if (columnType === 'string' || columnType === undefined || columnType === 'default') {
+          
+          const textColor = resolveReferences(column.textColor, currentState, { cellValue }) || [];
```
gondar00, line 280: Probably have a fallback default color if value is empty?

```suggestion
          const textColor = resolveReferences(column.textColor, currentState, { cellValue }) || '#000';
```

frontend/src/_helpers/utils.js
```diff
@@ -34,15 +34,15 @@ export function resolveAll(data, state) {
 
 }
 
-export function resolveReferences(object, state) {
+export function resolveReferences(object, state, customObjects) {
   if (typeof object === 'string') {
     if (object.startsWith('{{') && object.endsWith('}}')) {
       const code = object.replace('{{', '').replace('}}', '');
```
gondar00, line 40: This can also be written as 
```suggestion
      const code = object.replace(/^{{|}}$/g,'')
```

Navaneeth-pk: Yes, but I think the current version is easier to read.

frontend/src/Editor/Components/Table/Table.jsx
Navaneeth-pk: Right, we also need to consider '#fff' as the default if dark mode.

ckbhatiaredef: It's much better if you create some predefined color constants to be used throughout the app.

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -297,7 +304,7 @@ export function Table({
               />
             );
           }
-          return <span>{cellValue}</span>;
+          return <span style={cellStyles}>{cellValue}</span>;
```
ckbhatiaredef, line 307: Avoid inline-styles

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -682,9 +689,9 @@ export function Table({
                       if (componentState.changeSet) {
                         if (componentState.changeSet[cell.row.index]) {
 
-                          const accessor = columnData.find(column => column.id === cell.column.id)?.accessor;
+                          const currentColumn = columnData.find(column => column.id === cell.column.id);
 
-                          if (_.get(componentState.changeSet[cell.row.index], accessor, undefined) !== undefined) {
+                          if (_.get(componentState.changeSet[cell.row.index], currentColumn?.accessor, undefined) !== undefined) {
```
ckbhatiaredef, line 694: Just directly import `get` and use it here instead of importing everything.

frontend/src/Editor/Inspector/Components/Table.jsx
```diff
@@ -152,6 +152,21 @@ class Table extends React.Component {
             />
           </div>
 
+          {(column.columnType === 'string' || column.columnType === undefined || column.columnType === 'default') &&
```
ckbhatiaredef, line 155: Here too

frontend/src/Editor/Components/Table/Table.jsx
Navaneeth-pk: This is for customising the text colors, so they cannot be predefined as they are computed as per the configuration of the app.

PR #724, opened 2021-09-10
Feature: Apps should only be edited by the admin or developer of the organisation.

### LINKED ISSUE #723: Apps should only be edited by the `admin` or `developer` of the organisation.
### Description

Apps should only be edited by the `admin` or `developer` of the organisation.
Viewers should not be able to edit apps.

Hide the `Edit` button for the apps if the user role is `viewer`.

### Current behaviour
All the users in the organisation can edit the apps. (admin, developer and viewer).
**Loom**: https://www.loom.com/share/14610d5eaeeb4d4cae27d2e306709725

arpitnath:
Fixes #723 

**App**: https://tooljet-pr-app.herokuapp.com/
**Loom**: https://www.loom.com/share/733c179abbdf4fdaae59f585bfc731b4


server/src/entities/user.entity.ts
```diff
@@ -53,11 +53,13 @@ export class User extends BaseEntity {
 
   public isAdmin;
   public isDeveloper;
+  public role;
 
   @AfterLoad()
   computeUserRole(): void {
     this.isAdmin = this.organizationUsers[0].role === 'admin';
     this.isDeveloper = this.organizationUsers[0].role === 'developer';
+    this.role = this.organizationUsers[0].role;
```
akshaysasidrn, line 62: Does it make sense to make do with only with `role` and rid of `isAdmin` and `isDeveloper` as it would be redundant here?

frontend/src/HomePage/HomePage.jsx
```diff
@@ -206,19 +206,21 @@ class HomePage extends React.Component {
                                 <small className="pt-2 app-description">created {moment(app.created_at).fromNow()} ago by {app.user?.first_name} {app.user?.last_name} </small>
                               </td>
                               <td className="text-muted col-auto pt-4">
-                                <Link
+                                {currentUser.role !== 'viewer' && (
```
akshaysasidrn, line 209: This check should also be present at the router level because if the link is available, the viewer can still access it.
Should enforce the same at backend too. Wdyt?

akshaysasidrn (changes requested):

akshaysasidrn:
Please help fix the concerns here in a follow up PR @arpitnath 🙇🏾‍♂️
Let's merge this for now

PR #737, opened 2021-09-13
enable suggestion box for codehinter

gondar00:
This PR enables the preview box for all codehinters.

image

### DIFF
--- a/frontend/src/Editor/LeftSidebar/sidebar-datasources.js
+++ b/frontend/src/Editor/LeftSidebar/sidebar-datasources.js
@@ -9,12 +9,12 @@ export const LeftSidebarDataSources = ({ appId, darkMode, dataSources = [], data
   const [showDataSourceManagerModal, toggleDataSourceManagerModal] = React.useState(false);
   const [selectedDataSource, setSelectedDataSource] = React.useState(null);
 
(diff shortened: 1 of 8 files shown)


frontend/src/Editor/LeftSidebar/sidebar-datasources.js
```diff
@@ -4,20 +4,20 @@ import { LeftSidebarItem } from './sidebar-item';
...
   const [selectedDataSource, setSelectedDataSource] = React.useState(null);
 
-  const renderDataSource = dataSource => {
+  const renderDataSource = (dataSource) => {
     const sourceMeta = DataSourceTypes.find((source) => source.kind === dataSource.kind);
     return (
       <tr
         role="button"
-        key={sourceMeta.kind.toLowerCase()}
+        key={sourceMeta.id}
```
Navaneeth-pk, line 17: `sourceMeta` is just a metadata object, there won't be an `id` key. Can you please check?

gondar00: Yep, that's right, this was complaining on the console saying that 2 elements have same key, when 2 datasources of same kind are added, I have changed the key to index now

PR #740, opened 2021-09-13
fix: striped table when element selected is text/dropdown/multiselect/editable/datepicker

### LINKED ISSUE #741: [Widget]: On selection of tags option in table category type, page crashes
Steps to reproduce: 

1. Drag and drop a table
2. Open widget inspector
3. Select the tags option from the dropdown in column type
4. Page breaks

[tags_error.mov.zip](https://github.com/ToolJet/ToolJet/files/7154333/tags_error.mov.zip)

gondar00:
This PR fixes the table design when combination of a field text + table style striped

Before: 
image

After:
image

-----

Before: 
image
After: 
image

----

Fixes #741 

Incorrect value was provided in tags selection


Navaneeth-pk:
image

Steps to reproduce: 
a) drag and drop a table widget 
b) set the `email` column as editable 

Can you please take a look? @gondar00

gondar00:
I think the above image would've been on a diff. branch

On this branch it looks like the below: 
image

gondar00:
1. Making a column editable
2. Making a column type as dropdown 

These I am working on currently

gondar00:
Making a column editable is fixed, 

pending is making a column type as datepicker

gondar00:
Datepicker fix is done

sherfin94:
Looks like we didn't cover action buttons

https://www.loom.com/share/7340770dfe304a9db8c8cbbdac98effc

gondar00:
@sherfin94 yeah added fix for action button 👍

sherfin94:
There seems to be a strange issue with drop-down, multiselect, badge etc on this branch:
https://www.loom.com/share/3f660dddd2114bb3a508a23a18432e10

A similar issue seems to be there with datepicker also, though its certainly better than what we already have with datepicker :grimacing: .
image

gondar00:
image

```

  td {
    min-height: 35px;
    overflow-x: initial; // so due to overflow being initial the content overflows from the box in case of many fields
    margin: auto;
    ....
}
```

gondar00:
image
```

  td {
    min-height: 35px;
    overflow-x: auto; // i added auto here so that a scrollbar comes, but that has broken select/dropdown 
    margin: auto;
    ....
}
```

gondar00:
I'll change it back to `initial` for now

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -296,19 +297,96 @@ export function Table({
...
+                          column.key || column.name,
+                          e.target.value,
+                          cell.row.original
+                        );
+                      }
+                    }}
+                    className={`form-control-plaintext form-control-plaintext-sm ${!isValid ? 'is-invalid' : ''}`}
+                    defaultValue={cellValue}
+                  />
+                  <div class="invalid-feedback">{validationError}</div>
```
sherfin94, line 369:
```suggestion
                  <div className="invalid-feedback">{validationError}</div>
```

sherfin94:
Went through the code, looks nice!.
There is this one last problem I missed: It also has to support proper stripes when validation text is present
image

Once that's fixed, could you also merge the latest `develop` to this?

sherfin94 (changes requested):

sherfin94 (approved):

PR #743, opened 2021-09-13
Add ability to clone from existing apps

akshaysasidrn:
image


server/src/controllers/apps.controller.ts
```diff
@@ -107,6 +107,25 @@ export class AppsController {
...
   }
 
+
+  @UseGuards(JwtAuthGuard)
+  @Post(':id/clone')
+  async clone(@Request() req, @Param() params) {
+    const existingApp = await this.appsService.find(params.id);
+    const ability = await this.appsAbilityFactory.appsActions(req.user, {});
+
+    if (!ability.can('viewApp', existingApp)) {
```
Navaneeth-pk, line 117: Why the ability is `viewApp`?  🤔
Only admins and developers should be able to clone apps.

server/src/services/apps.service.ts
```diff
@@ -74,6 +76,86 @@ export class AppsService {
...
+
+    const existingDataSources = await manager.find(DataSource, {
+      app: existingApp,
+    });
+
+    for (const source of existingDataSources) {
+      const newSource = manager.create(DataSource, {
+        app: newApp,
+        name: source.name,
+        options: source.options,
```
Navaneeth-pk, line 123: If the data source has encrypted credentials, we should clone those too instead of making reference to the same credential entity and then will have to update the options.

frontend/src/HomePage/HomePage.jsx
```diff
@@ -86,6 +88,22 @@ class HomePage extends React.Component {
     this.setState({ showAppDeletionConfirmation: true, appToBeDeleted: app });
   }
 
+  cloneApp = (app) => {
```
gondar00, line 91: I think there is no parameter `app` passed to this function from the below snippet
```
  <div className="field mb-2">
    <span className="field mb-2" role="button" onClick={() => cloneApp()}>
      Clone app
    </span>
  </div>
```
I think `app` is being used from the global parameter

server/src/controllers/apps.controller.ts
akshaysasidrn: The assumption was if the app is viewable should be able to clone them for modification as well.
Will revise then

PR #762, opened 2021-09-15
feat: add datasource type into individual schema

gondar00:
This PR refactors datasource types in the schema file,


frontend/src/Editor/DataSourceManager/SourceComponents/Api/Graphql.schema.json
```diff
@@ -3,7 +3,22 @@
   "$id": "https://tooljet.io/Graphql.schema.json",
   "title": "Graphql datasource",
   "description": "A schema defining graphql datasource",
-  "type": "object",
+  "type": {
```
akshaysasidrn, line 6: CMIIW, according to JSON schema spec this [type](https://json-schema.org/understanding-json-schema/reference/type.html) would be invalid here, right?
Does it make sense to add this as the definition instead?

akshaysasidrn: Or we can use some other keyword which isn't part of the spec

gondar00:
```
string
number
integer
object
array
boolean
null
```
Yeah true, only above values can exist inside type as per spec, I'll rename it to some other thing with a `$` in front to specify its custom

akshaysasidrn (approved):

PR #765, opened 2021-09-16
Add error to suggestion box in codehinter

gondar00:
This PR adds error block to codehinter when `eval` errors are happening

https://user-images.githubusercontent.com/12490590/133554558-061809ed-5371-431a-ab8a-7086fdbf6304.mov

### DIFF
--- a/frontend/src/_helpers/utils.js
+++ b/frontend/src/_helpers/utils.js
@@ -30,59 +30,78 @@ export function resolve(data, state) {
   }
 }
 
-export function resolveAll(data, state) {
+export function resolveReferences(object, state, defaultValue, customObjects = {}, v2 = false) {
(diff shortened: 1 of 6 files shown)


frontend/src/_helpers/utils.js
```diff
@@ -30,59 +30,78 @@ export function resolve(data, state) {
   }
 }
 
-export function resolveAll(data, state) {
+export function resolveReferences(object, state, defaultValue, customObjects = {}, v2 = false) {
```
Navaneeth-pk, line 33: instead of `v2`, can we name the param to something more readable?

gondar00: ok changed to `withError`

Issue #789, opened 2021-09-20
Unable to access user invitation link on Local


mohinid:
Getting below error on clicking the invitation link. Please check below SS.

<img width="1791" alt="Screenshot 2021-09-20 at 2 43 26 PM" src="https://user-images.githubusercontent.com/5313625/133979937-8be44541-d708-4c40-88ca-3f39a5e405cd.png">
<img width="1786" alt="Screenshot 2021-09-20 at 2 43 42 PM" src="https://user-images.githubusercontent.com/5313625/133979944-4fa23e3f-93be-4a2b-ab9d-e7e8ab4b88a9.png">
<img width="1789" alt="Screenshot 2021-09-20 at 2 43 58 PM" src="https://user-images.githubusercontent.com/5313625/133979949-aead72d4-b92f-4e94-92cf-1ec7f12058a1.png">

akhilgkrishnan:
@Navaneeth-pk picking up this issue

PR #799, opened 2021-09-20
Fix: User Invitation URL in email is not clickable

### LINKED ISSUE #795: Invitation URL in email should be clickable.
The invitation URL in the email should be clickable(hyperlink). It’s plain text for now.

Screenshot 2021-09-20 at 11 15 43 AM

akhilgkrishnan:
Fixed #795 , #789

### DIFF
--- a/server/src/services/email.service.ts
+++ b/server/src/services/email.service.ts
@@ -82,6 +82,7 @@ export class EmailService {
   async sendOrganizationUserWelcomeEmail(to: string, name: string, sender: string, invitationtoken: string) {
 
     const subject = 'Welcome to ToolJet';
+    const inviteUrl = `${this.TOOLJET_HOST}/invitations/${invitationtoken}`
     const html = `
       <!DOCTYPE html>
       <html>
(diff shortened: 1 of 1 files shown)


akhilgkrishnan:
@Navaneeth-pk please review

server/src/services/email.service.ts
```diff
@@ -94,7 +95,7 @@ export class EmailService {
           ${sender} has invited you to use ToolJet. Use the link below to set up your account and get started.
           </span>
           <br>
-          <a href="<%= @url %>">${`${this.TOOLJET_HOST}/invitations/${invitationtoken}`}</a>
+          <a href="${inviteUrl}">${inviteUrl}</a>
```
Navaneeth-pk, line 98: Can we change `href="${inviteUrl}"` to `href={inviteUrl}`?

akhilgkrishnan: No. It takes it as a string. because the whole content is enclosed in a back tick.

PR #860, opened 2021-09-28
get queryOptions from url provided and populate the params tab

### LINKED ISSUE #855: Simple GET API call fails with error
### Version Information
latest / v0.7.3

### Environment
Docker

Preview the JSON of the API call result
the same API returns with cURL and it is a valid JSON object

Error:
"Query could not be completed: Unexpected token < in JSON at position 0"

1. Add a REST API source
2. Type GET
3. Source "https://en.wikipedia.org/w/api.php?action=query&format=json&generator=search&gsrnamespace=0&gsrlimit=5&gsrsearch=API"

### Screenshots or Screencast

api_fail

not sure, not much troubleshooting info on https://docs.tooljet.io/

### Any possible solutions?
no

no

no

arpitnath:
Fixes #855


arpitnath:
App: https://tooljet-pr-app.herokuapp.com/apps/67d183cf-5f19-45a5-8805-11dc3813af01

arpitnath:
Screencast:

https://user-images.githubusercontent.com/67645175/135306328-491cec9f-c3a5-4bad-840d-1a0843978e9e.mov

Navaneeth-pk (approved):

PR #861, opened 2021-09-28
button text fix

ajinkyaa:


ajinkyaa:
Work in progress, please do not merge.

frontend/src/Editor/Components/Button.jsx
```diff
@@ -51,7 +49,7 @@ export const Button = function Button({ id, width, height, component, onComponen
         onComponentClick(id, component);
       }}
     >
-      {text}
+      {getDynamicText(text, currentState)}
```
sherfin94, line 52: You can directly invoke `resolveReferences` here, instead of calling a separate function to do so.
```suggestion
      {resolveReferences(text, currentState)}
```

frontend/src/_helpers/utils.js
```diff
@@ -227,3 +227,18 @@ export function validateWidget({ validationObject, widgetValue, currentState, cu
     validationError,
   };
 }
+
+export const getDynamicText = (text, currentState) => {
```
sherfin94, line 231: The `resolveReferences` function will take care of resolving every code expression inside the string. There is no need to separate them out from scratch. I don't feel this new function is required.

sherfin94 (changes requested):

Issue #898, opened 2021-10-05
Add support for setting color of star widget 


gondar00:
Is your proposal related to a problem?
Presently, the star widget can have only one label color: black.

Describe the solution you'd like
Add a color selector in inspector for star label

Hardcoded value can be found [here](https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/Editor/Components/StarRating/index.js#L101)

virajbahulkar:
@sherfin94  picking this up

PR #900, opened 2021-10-05
Set label color to star widget

virajbahulkar:
Fixes - #898


virajbahulkar:
@Navaneeth-pk Heroku link added. please check

PR #903, opened 2021-10-05
Feature: Googlesheet update operation

### LINKED ISSUE #915: Update Operation for Google Sheet Datasource
### Description
- updating a row
- able to update cell value(s) of a particular row
- using a unique column and the corresponding value, identify the single row to update

arpitnath:
Fixes #915 

**Loom**:  https://www.loom.com/share/0ec488e18927457d85c1a100cea60ce2
**App**:  https://tooljetapp-pr.herokuapp.com/apps/c01fcf8d-03dc-43b4-bf28-717993a9650a


sherfin94:
@arpitnath Can we display the saved values of `WHERE` and `update value` inputs?

image

sherfin94:
There seems to be some issue with how the values are parsed. when I supply some values from the table to the query and click on preview it works. However, if I change the value in the table and run the query(not preview), the value seems to be updated to the initial value that it parsed and not the new value. I'm suspecting that what's being stored is the parsed value and not the value that contains code. That is , `{{1+1}}` might be stored as `2` instead of `{{1+1}}`. My screencast is a bit big for github, I can share in some other platform if you'd like.

PR #918, opened 2021-10-06
Feature: User access management

akshaysasidrn:
Adds support for user management with group permissions
Changes:
- Roles usage will be deprecated
- By default there will be two group `all_users` and `admin`
- Users in `admin` group will have all access and can have users added to it
- Users in `all_users` group can be edited for app permissions but cant have users removed
- New group created should have unique group name within the organization
- Apps added to groups can have either view or edit permission
- If user is part of multiple groups, most permissible group is taken

Group creation:
image

Group list:
image

Group view:
image
image

ERD:
image


akshaysasidrn:
@gondar00 have updated, kindly re-review

PR #977, opened 2021-10-11
Small fixes

donno2048:
Remove unnecessary variable, use `var` not `let` as `x` and `y` will not remain in their default values, don't check for `!isTesting` in the else part of a trinary statement with `isTesting` as `if` condition


arpitnath:
@donno2048 
Thanks for contributing to ToolJet again. Would you mind filling up our Hacktoberfest form to confirm your swags? 
https://app.tooljet.io/applications/hacktoberfest-2021-form

donno2048:
OK

PR #1060, opened 2021-10-14
Tackle JS-0032

### LINKED ISSUE #1055: (JS-0032) Found `await` inside of loops
### Description
Performing an operation on each element of an iterable is a common task.

### Occurrences

https://deepsource.io/gh/arpitnath/ToolJet/issue/JS-0032/occurrences

radiantly:
Fixes #1055


radiantly:
In that case, the original code is completely valid and the detection was a false positive. I've added a comment on the line to skip detection.

PR #1069, opened 2021-10-14
[Fix]: #1057 (JS-0443) fixed comparison of objects

### LINKED ISSUE #1057: (JS-0443) usage of `setState` in `componentDidUpdate`
### Description
Updating the setState inside componentDidUpdate can create an infinite loop because there's no break condition.

### Occurrences
https://deepsource.io/gh/arpitnath/ToolJet/issue/JS-0443/occurrences

Example
```
componentDidUpdate(previousProps, previousState) {
    if (previousProps.data !== this.props.data) {
        this.setState({/*....*/})
    }
}
```

keshav-bohr:
This PR fixes #1057 issue

Objects were compared with equal sign, so fixed the object comparison using lodash `frontend/src/Editor/DataSourceManager/DataSourceManager.jsx`

And the second instance of componentDidUpdate seems fine to me. Please check

### DIFF
--- a/frontend/src/Editor/DataSourceManager/DataSourceManager.jsx
+++ b/frontend/src/Editor/DataSourceManager/DataSourceManager.jsx
@@ -42,7 +42,7 @@ class DataSourceManager extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    if (!_.isEqual(prevProps.selectedDataSource, this.props.selectedDataSource)) {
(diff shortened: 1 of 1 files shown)


arpitnath:
Yes, you are correct.
```
{
  "appId": "4ca17070-2098-467e-a714-7ceb9ee6e77e",
  "showDataSourceManagerModal": false,
  "darkMode": true,
  "selectedDataSource": null
}
```
a simple check to validate if `selectedDataSource` is not null can be enough. I think

PR #1118, opened 2021-10-16
fix(folders): add no folder present messaging in dropdown

jyash97:
No search message:
image

Cannot share screenshot for No folders message since there is no way to delete folders


jyash97:
@Navaneeth-pk there isn't any issue created.

Issue #1137, opened 2021-10-18
"Ago" appears twice in application info in homepage


Jayden-Chiu:
### Environment

Prod environment (https://app.tooljet.io/)

### What is the expected behaviour?

Application creation date time stamp should only say "ago" once

### What is the current behaviour?

Application creation time stamp says "created X ago ago by X"

### How to reproduce the issue?

1. Go to app home page
2. Click create new application
3. Go back to home page
4. Under the application you just created, it should say ago twice

### Screenshots or Screencast

![image](https://user-images.githubusercontent.com/57676066/137814811-0dc65924-baf6-4000-beb5-9e7404ea1c47.png)

### Any possible solutions?
 pass `true` as argument to `.fromNow()` as described in [moment docs](https://momentjs.com/docs/) to remove suffix "ago"

### Can you identify the location in the driver source code where the problem exists?
`frontend > src > HomePage > HomePage.jsx > line 255`

### If the bug is confirmed, would you be willing to submit a PR?
Yes

PR #1141, opened 2021-10-19
remove ago suffix from app creation date

Jayden-Chiu:
Fixes [#1137](https://github.com/ToolJet/ToolJet/issues/1137)

[Heroku deployment](https://tooljet-ago-appears-twice.herokuapp.com/)


Navaneeth-pk:
@Jayden-Chiu 
Yes, the form is actually an application built using ToolJet. We might've used the wrong regex to validate the field. Please submit the form without the hyphen, we will fix it from our end.

@arpitnath 
Can you please take a look at the application?

PR #1265, opened 2021-10-25
Fixed the SEO issues

saintmalik:
So the conical urls and alternate urls, Open graph url showing for the docs page are broken due to an error in docusaurus config file.

## Before

sdfgdf
UUGIIF

## After 

fdgfdgdf
gdsffg


saintmalik:
> We are not able to reproduce this issue at our end and none of the other users has reported this error. Can you please try debugging the issue if you have time? Meanwhile please send your details to `hello@tooljet.io`

Did you guys get the mail?

PR #1274, opened 2021-10-25
Feature: File Picker Widget

### LINKED ISSUE #1271: Feature: File Picker Widget
### Description
- Create a new widget in the widget ecosystem -  File Picker.
- User should be able to pick a file(s) or use drag and drop 

Properties: 
```
height
width
enableDropzone - bool
enablePicker - bool (input picker)
exposeVariable  : { name, content,  type }
maxFileCount - 2 (e.g) // ! hide and show max file count when enableMultiple is true/false
enableMultiple - bool
fileType - [ '.png', '.json'] (types of file that it could accept)
maxSizeLimit (Bytes)
minSizeLimit (Bytes)
visibility
disabled 
event - 'fileSelected' (fires an event "fileSelected" )

```

arpitnath:
Fixes #1271 

### App:


arpitnath:
update: https://www.loom.com/share/906550dac4b0468585ef3d3af7e8a164

frontend/src/Editor/Components/FilePicker.jsx
```diff
@@ -0,0 +1,257 @@
+import React, { useEffect, useMemo } from 'react';
+import { useDropzone } from 'react-dropzone';
+import { resolveWidgetFieldValue } from '@/_helpers/utils';
+import { toast } from 'react-toastify';
+
+export const FilePicker = ({ width, height, component, currentState, onComponentOptionChanged, onEvent, darkMode }) => {
+  //* property definations
```
akshaysasidrn, line 7: nitpick: definitions*

akshaysasidrn (approved):
@arpitnath Looks good. Need to decide if and how we should handle case wherein user wouldn't want to set a predefined type.

frontend/src/Editor/Components/FilePicker.jsx
```diff
@@ -0,0 +1,257 @@
...
+  const enableDropzone = component.definition.properties.enableDropzone?.value ?? true;
+  const enablePicker = component.definition.properties?.enablePicker?.value ?? true;
```
akshaysasidrn, line 9: we can also expect properties here as well?

Navaneeth-pk (approved):

akshaysasidrn:
> @arpitnath Looks good. Need to decide if and how we should handle case wherein user wouldn't want to set a predefined type.

On further discussion, found this is already handled if we put empty values for accept fields. 👍🏾

arpitnath:
Will add the details in the docs with a follow up PR

PR #1289, opened 2021-10-25
fix(editor): avoid empty slug in share URL

### LINKED ISSUE #1288: Empty slug in share application URL
### Environment

ToolJet Cloud

Slug id should not be empty

Slug id can be empty

1. Open share modal in editor
2. Add spaces in slug id
3. Copy the URL 

### Screenshots or Screencast

Empty Slug Id

### Any possible solutions?
Prevent spaces to be considered as valid

yes

jyash97:
Closes #1288

### DIFF
--- a/frontend/src/Editor/ManageAppUsers.jsx
+++ b/frontend/src/Editor/ManageAppUsers.jsx
@@ -104,7 +104,7 @@ class ManageAppUsers extends React.Component {
   };
 
   handleSetSlug = (event) => {
-    const newSlug = event.target.value || this.props.app.id;
+    const newSlug = event.target.value.trim() || this.props.app.id;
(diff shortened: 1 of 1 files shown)


jyash97:
@Navaneeth-pk  Facing issues with heroku deploy. Maybe screen recording will help: 
Empty slug url

Navaneeth-pk:
If the slug is empty, the field should be shown empty but when the user click on the copy icon, we will copy the URL that includes the id of the application?
@jyash97

jyash97:
Hey @Navaneeth-pk  👋

Yes in case of empty slug (entered by user) , on copying the url will contain application id, else it just goes to `/application`

Navaneeth-pk:
closing as this change is not required now.

PR #1332, opened 2021-10-29
Reset to default value when option values gets an update

### LINKED ISSUE #1331: Reset dropdown value to default when its option values are changed
Current value should reset to default when option values are changed

If the options values are changed, the current value does not get cleared.

### Screenshots or Screencast

https://user-images.githubusercontent.com/67645175/139456183-1a932e8e-3583-4369-8f61-83c50896b372.mov

arpitnath:
Fixes #1331 

### Screenshots or Screencast

https://user-images.githubusercontent.com/67645175/139457888-e0aa998c-c829-491a-8f4e-d1605c3ad29f.mov


frontend/src/Editor/Components/DropDown.jsx
```diff
@@ -63,6 +63,19 @@ export const DropDown = function DropDown({
   const value = currentValueProperty ? currentValueProperty.value : '';
   const [currentValue, setCurrentValue] = useState('');
 
+  const [optionValues, setOptionValue] = useState(() => selectOptions);
+
+  useEffect(() => {
```
sherfin94, line 68: We have to reset the value to default value. Here instead it seems like we're resetting to an arbitrary value: `''`. 

If the default value is not present in the latest set of options, then it might be a good idea to reset to `''`.

arpitnath: yeah if the default value is not present and we should reset it to `' '`

arpitnath:
https://user-images.githubusercontent.com/67645175/140286894-9326be3d-5c91-4980-98c0-491f84f8b001.mov

sherfin94 (approved):

Navaneeth-pk (approved):

PR #1356, opened 2021-11-02
fix: card styles, add border and box shadow

gondar00:
This PR makes the comment sidebar css under a parent class, as card css were affecting across all cards


gondar00:
closing as made same changes in #1357

PR #1359, opened 2021-11-02
Fixed : Show tooltip on comments tab which says indicates that the app needs to be saved to open comments

### LINKED ISSUE #1357: Show tooltip on comments tab which says indicates that the app needs to be saved to open comments

Presently, a new user will get confused why they can't open the comments tab, it is unclear that at least one version of the app needs to be saved for the comments section to be activated.

Show a tooltip on hover of the inactive comments section icon, which indicates that at least one version of the app needs to be saved.

saisumanthkumar:
Fixes #1357 

heroku-deploy : https://tooljet-s.herokuapp.com/apps/75fc49fd-06b5-4f95-acc0-8d4d6f883321

### DIFF
--- a/frontend/src/Editor/LeftSidebar/SidebarComment.jsx
+++ b/frontend/src/Editor/LeftSidebar/SidebarComment.jsx
@@ -7,15 +7,17 @@ export const LeftSidebarComment = ({ toggleComments, appVersionsId }) => {
   return (
     <LeftSidebarItem
       commentBadge
-      tip="toggle comments"
+      tip={appVersionsId ? 'toggle comments' : 'At least one version of the app should be deployed'}
       icon={`comments`}
       className={cx(`left-sidebar-item sidebar-zoom position-relative`, {
         disabled: !appVersionsId,
         active: isActive,
       })}
       onClick={() => {
-        toggleActive(!isActive);
-        toggleComments();
+        if (appVersionsId) {
+          toggleActive(!isActive);
+          toggleComments();
+        }
       }}
     />
   );
--- a/frontend/src/Editor/LeftSidebar/SidebarComment.jsx
+++ b/frontend/src/Editor/LeftSidebar/SidebarComment.jsx
@@ -7,7 +7,7 @@ export const LeftSidebarComment = ({ toggleComments, appVersionsId }) => {
   return (
     <LeftSidebarItem
       commentBadge
-      tip={appVersionsId ? 'toggle comments' : 'At least one version of the app should be deployed'}
+      tip={appVersionsId ? 'toggle comments' : 'Comments section will be available once you save this application'}
       icon={`comments`}
       className={cx(`left-sidebar-item sidebar-zoom position-relative`, {
         disabled: !appVersionsId,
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/LeftSidebar/SidebarComment.jsx
```diff
@@ -7,15 +7,17 @@ export const LeftSidebarComment = ({ toggleComments, appVersionsId }) => {
   return (
     <LeftSidebarItem
       commentBadge
-      tip="toggle comments"
+      tip={appVersionsId ? 'toggle comments' : 'At least one version of the app should be deployed'}
```
sherfin94, line 10: Can we change the tooltip to:
```suggestion
      tip={appVersionsId ? 'toggle comments' : 'Comments section will be available once you save this application'}
```

frontend/src/Editor/LeftSidebar/SidebarComment.jsx
```diff
@@ -7,15 +7,17 @@ export const LeftSidebarComment = ({ toggleComments, appVersionsId }) => {
...
       icon={`comments`}
       className={cx(`left-sidebar-item sidebar-zoom position-relative`, {
         disabled: !appVersionsId,
         active: isActive,
       })}
       onClick={() => {
-        toggleActive(!isActive);
-        toggleComments();
+        if (appVersionsId) {
```
sherfin94, line 17: I feel this condition is not required here as its functionality is already achieved by `disabled: !appVersionsId` on line number 13.

sherfin94 (changes requested):

frontend/src/Editor/LeftSidebar/SidebarComment.jsx
saisumanthkumar: @sherfin94 If the app is disabled ,the tooltip will not show

frontend/src/Editor/LeftSidebar/SidebarComment.jsx
saisumanthkumar: @sherfin94 yeah sure

frontend/src/Editor/LeftSidebar/SidebarComment.jsx
gondar00: @saisumanthkumar Actually the click event works because of adding the pointer events css[here](https://github.com/ToolJet/ToolJet/pull/1359/files#diff-1ff9cafa7affbc78d3367c1918f48b0cce543785adfe5064010262f68c30e7c9R23)

Any particular reason to add that css property?

saisumanthkumar: @gondar00 yeah by default the pointer event is none , so even if we hover nothing happens , so I set it to visible

gondar00: image
https://github.com/react-bootstrap/react-bootstrap/issues/1588

Hey @saisumanthkumar, I have found few references around why the tooltip wasn't working on disabled elements, 

Instead of adding pointer events: all/visible; recommended way is to wrap the disabled element around a <div />

gondar00 (approved):

frontend/src/Editor/LeftSidebar/SidebarComment.jsx
saisumanthkumar: @gondar00 thanks for the ref.

Navaneeth-pk (approved):

PR #1384, opened 2021-11-08
Broadcast only to clients connected to a particular app id

gondar00:
This PR adds filter to the broadcast method where broadcast only to clients connected to a particular app id

### DIFF
--- a/server/src/events/events.gateway.ts
+++ b/server/src/events/events.gateway.ts
@@ -18,7 +18,7 @@ export class EventsGateway implements OnGatewayConnection, OnGatewayDisconnect {
   handleDisconnect(client: any): void {}
 
   broadcast(data) {
-    this.server.clients.forEach((client) => {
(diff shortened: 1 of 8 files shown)


server/src/events/events.gateway.ts
```diff
@@ -18,12 +18,20 @@ export class EventsGateway implements OnGatewayConnection, OnGatewayDisconnect {
   handleDisconnect(client: any): void {}
 
   broadcast(data) {
-    this.server.clients.forEach((client) => client.send(data));
+    this.server.clients.forEach((client) => {
+      if (client.appId === data.appId) client.send(data.message);
```
akshaysasidrn, line 22: @gondar00 Get error on this line when compiling. Could you check?

gondar00: Yes, I have pushed the fix should work fine now 👍

akshaysasidrn (approved):

PR #1415, opened 2021-11-16
Read file as dataURL or textString

### LINKED ISSUE #1405: File Picker Incorrectly encodes binary data
### Environment
Docker

Being able to decode binary payloads

reader.readAsText breaks binary files.

Use filepicker.file as value in post request.

### Any possible solutions?
Use base64 encoding or provide option to send file as form data.

Proposed crude fix in Editor/Components/FilePicker.jsx
```
reader.readAsDataURL(acceptedFile); # line 113 currently reads as text

content: zippedResults[0].srcElement.result.split(',')[1], # line 118 split data:binary data url part
```

Yes.

arpitnath:
Fixes #1405 

app: https://tooljetpr-app.herokuapp.com/apps/96d4ea14-6223-4127-b751-d1184e2e8c36

Screenshots/Screencast:
Screenshot 2021-11-16 at 4 53 51 PM
Screenshot 2021-11-16 at 4 59 20 PM

https://www.loom.com/share/4f51052e88de44e9a4917c45ff1cf672

### DIFF
--- a/frontend/src/Editor/Components/FilePicker.jsx
+++ b/frontend/src/Editor/Components/FilePicker.jsx
@@ -7,8 +7,6 @@ export const FilePicker = ({ width, height, component, currentState, onComponent
   //* properties definitions
   const enableDropzone = component.definition.properties.enableDropzone?.value ?? true;
   const enablePicker = component.definition.properties?.enablePicker?.value ?? true;
-  const readAsDataURL = component.definition.properties?.readAsDataURL?.value ?? true;
-  const readAsText = component.definition.properties?.readAsText?.value ?? true;
   const maxFileCount = component.definition.properties.maxFileCount?.value ?? 2;
   const enableMultiple = component.definition.properties.enableMultiple?.value ?? false;
   const fileType = component.definition.properties.fileType?.value ?? 'image/*';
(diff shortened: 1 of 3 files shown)


frontend/src/Editor/Components/FilePicker.jsx
```diff
@@ -95,6 +96,47 @@ export const FilePicker = ({ width, height, component, currentState, onComponent
...
+    const readFileAsText = await getFileData(file);
+
+    // * readAsDataURL
+    const readFileAsDataURL = await getFileData(file, 'readAsDataURL');
+
+    return {
+      name: file.name,
+      type: file.type,
+      content: readFileAsText,
+      data: readFileAsDataURL,
```
akshaysasidrn, line 136: I'm thinking about whether it will be better to expose this field as dataUrl instead of data itself such that we don't split result when doing using `readAsDataURL` method. The reasoning being users can use it directly without any preprocessing. Wdyt?

frontend/src/Editor/Components/components.js
```diff
@@ -1166,7 +1166,7 @@ export const componentTypes = [
       disabledState: { type: 'code', displayName: 'Disable' },
     },
     exposedVariables: {
-      file: [{ name: [], content: [], type: [] }],
+      file: [{ name: [], content: '', data: '', type: '' }],
```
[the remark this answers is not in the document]
arpitnath: yeah name should be string 👍🏻

frontend/src/Editor/Components/FilePicker.jsx
arpitnath: yeah, dataURL makes sense

akshaysasidrn (approved):
Looks good!

arpitnath:
Have removed the app

Navaneeth-pk (approved):

PR #1446, opened 2021-11-23
feat: add restore shortcut when appDefinition components are changed

gondar00:
This PR adds a shortcut i.e `⌘Z` for undo-ing a change made to a component
This PR adds a shortcut i.e `⌘+shift+Z` for redo-ing a change made to a component

### Define change made to a component? 
Change can be user activities such as resizing, deleting, moving the element on the canvas, add-edit component to canvas. 

https://user-images.githubusercontent.com/12490590/143017412-14e950e4-0c81-449d-b47e-c61afc4f2fd2.mov

----

### What is Immer patches?
To know more about immer please visit their [website](https://immerjs.github.io/immer/patches/). 
Immer brings in a feature called [Patches](https://immerjs.github.io/immer/patches/). That allows us to implement many cool features – one of them being Undo-Redo.

Immer records all the changes that are performed on the draft object and generates an array of JSON objects indicating what has changed. These arrays are called Patches. Immer also produces inverse patches, which are patches that need to be applied if you want to go back to the previous state.

### Implementation

appDefinition is the place where all the components including their layout changes are stored. Whenever a change is made to appDefinition: `appDefinitionChanged` is the callback function invoked every-time any changes to `appDefinition` are made.

In this callback function have added: immer function i.e

```
    produce(
      this.state.appDefinition,
      (draft) => {
        draft.components = newDefinition.components;
      },
      this.handleAddPatch
    );
```

Next, `handleAddPatch` function handles 5 things: 

1. Updating the **currentVersion**
2. Add the redo and undo keys to the **currentVersionChanges** object
3. Sets `canUndo` & `canRedo` properties
4. When a new patch is added it deletes the redo version i.e. `delete currentVersion + 1`
5. Maintains the **max** number of version as a lot of versions can effect the browser perf. currently set to **100**.  i.e if the currentVersion is 100 then it starts deleting the 0th version, 101 then it deletes the 1st version and so on, such that at any given point in time the maximum number of component versions stored inside `currentVersionChanges` are 100 to preserve browser performance due to the deeply nested nature of `appDefinition` the `currentVersionChanges` can be a very large object.

```
  handleAddPatch = (patches, inversePatches) => {
    if (isEmpty(patches) && isEmpty(inversePatches)) return;
    this.currentVersion++;
    this.currentVersionChanges[this.currentVersion] = {
      redo: patches,
      undo: inversePatches,
    };

    this.canUndo = this.currentVersionChanges.hasOwnProperty(this.currentVersion - 1);
    this.canRedo = this.currentVersionChanges.hasOwnProperty(this.currentVersion + 1);

    delete this.currentVersionChanges[this.currentVersion + 1];
    delete this.currentVersionChanges[this.currentVersion - this.noOfVersionsSupported];
  }
```

### Things to keep in mind
 1. When we receive an undoable action – we can always undo but cannot redo anymore.
 2. Whenever you perform an undo – you can always redo and keep doing undo as long as we have a patch for it.
 3. Whenever you redo – you can always undo and keep doing redo as long as we have a patch for it.


Navaneeth-pk:
Please fix the merge conflicts. 
@gondar00

gondar00:
@Navaneeth-pk fixed merge conflict 👍

sherfin94 (commented):
Adding an event handler, and undoing a couple of times seems to crash the app:

undo-after-event-handler

sherfin94:
> Adding an event handler, and undoing a couple of times seems to crash the app:
> 
> undo-after-event-handler

Actually, now I think the crash happens whenever the inspector panel is open.

1. Add any component
2. Open its inspector panel
3. Undo several times

@gondar00

gondar00:
Thanks @sherfin94 

I was able to reproduce this bug, its related to inspector panel

https://user-images.githubusercontent.com/12490590/145200483-d83206ad-cd2f-4135-b9e4-165437faa710.mov

1. click on widget
2. open inspector 
3. undo, undo, undo
4. app crashes

gondar00:
image
One observation is layouts get added to appDefinition after opening the inspector, this creates a undoable action as appDefinition changed, will fix this one in a different PR

gondar00:
@sherfin94 added the fix

akshaysasidrn:
Apologies for late question, do we also have to add some buttons as well for undo/redo?

frontend/src/Editor/Editor.jsx
```diff
@@ -30,11 +29,16 @@ import { WidgetManager } from './WidgetManager';
...
 import queryString from 'query-string';
+import toast from 'react-hot-toast';
+import { cloneDeep, isEqual, isEmpty } from 'lodash';
+import produce, { enablePatches, setAutoFreeze, applyPatches } from 'immer';
 import Logo from './Icons/logo.svg';
 import EditIcon from './Icons/edit.svg';
 import MobileSelectedIcon from './Icons/mobile-selected.svg';
 import DesktopSelectedIcon from './Icons/desktop-selected.svg';
 
+setAutoFreeze(false);
```
akshaysasidrn, line 40: What happens if we set this, it should be a safeguard to not allow any modifications?  🤔

server/migrations/1638255797809-SetLoadingStateToFalseForExistingTables.ts
```diff
@@ -0,0 +1,46 @@
+import { AppVersion } from '../src/entities/app_version.entity';
```
akshaysasidrn, line 1: Will have to update for the template definitions as well

frontend/src/Editor/Editor.jsx
```diff
@@ -161,6 +166,17 @@ class Editor extends React.Component {
     });
   };
 
+  // 1. When we receive an undoable action – we can always undo but cannot redo anymore.
+  // 2. Whenever you perform an undo – you can always redo and keep doing undo as long as we have a patch for it.
+  // 3. Whenever you redo – you can always undo and keep doing redo as long as we have a patch for it.
+  initComponentVersioning = () => {
+    this.currentVersion = -1;
```
akshaysasidrn, line 173: nitpick: will this be confusing later b/w app's current version and immer patch's current version? ́😅

frontend/src/Editor/Editor.jsx
```diff
@@ -302,13 +318,70 @@ class Editor extends React.Component {
...
+
+      if (!appDefinition) return;
+      this.setState({
+        appDefinition,
+      });
+    }
+  };
+
   appDefinitionChanged = (newDefinition) => {
+    produce(
```
akshaysasidrn, line 372: Nice!

sherfin94:
> @sherfin94 added the fix

Hey @gondar00, looks like the crash still happens in certain scenarios:

1. Add a button first
2. Then add a table
3. Click on table handle (so that inspector opens)
4. Undo twice

undo-issue

frontend/src/Editor/Editor.jsx
```diff
@@ -161,6 +166,17 @@ class Editor extends React.Component {
     });
   };
 
+  // 1. When we receive an undoable action – we can always undo but cannot redo anymore.
```
akshaysasidrn, line 169: What is an undoable action? Switching between app versions?

frontend/src/Editor/Editor.jsx
gondar00: Actually what immer does is it uses Object.freeze and gives out an immutable object, we are using immer for the `patches` functionality currently, 

this disables the immutability feature as introducing that at current stage is a slightly bigger change

server/migrations/1638255797809-SetLoadingStateToFalseForExistingTables.ts
gondar00: 👍 

Added. them in this commit : https://github.com/ToolJet/ToolJet/pull/1446/commits/09978dda850256badebfe73de98aee9cc193d3b4

frontend/src/Editor/Editor.jsx
gondar00: Actually app version can be version 1, but the immer patch's current version can be anywhere between a range of 1-100, 

But thanks for bringing this up, one issue I found is this doesn't reset on changing the `app` version, I feel on change of app version calling `initComponentVersioning` should suffice

frontend/src/Editor/Editor.jsx
gondar00: > > @sherfin94 added the fix
> 
> Hey @gondar00, looks like the crash still happens in certain scenarios:
> 
> 1. Add a button first
> 2. Then add a table
> 3. Click on table handle (so that inspector opens)
> 4. Undo twice
> 
> undo-issue

This I've added a check it seems to work fine now, thanks ^^

gondar00:
> > @sherfin94 added the fix
> 
> Hey @gondar00, looks like the crash still happens in certain scenarios:
> 
> 1. Add a button first
> 2. Then add a table
> 3. Click on table handle (so that inspector opens)
> 4. Undo twice
> 
> undo-issue

This I've added a check it seems to work fine now, thanks ^^

frontend/src/Editor/Editor.jsx
gondar00: Any change made to appDefinition is an undo/redo action 

resizing, deleting, moving the element on the canvas, add-edit component to canvas.

yes switching b/w app versions works too

sherfin94 (changes requested):
Looks like a great PR @gondar00  👏🏼 👏🏼 

One small thing, unrelated to the main purpose of this PR that we'll have to fix is, the new type of toasts doesn't show different colors when different types of alert is selected (success, failure etc).
toast

sherfin94 (approved):

frontend/src/Editor/Editor.jsx
akshaysasidrn: Oh I see, yeah by reading immer docs I got that impression. Thought we can leverage immutability here as well. Sure can do later, just curious.

frontend/src/Editor/Editor.jsx
akshaysasidrn: I meant we store `currentVersionId` for apps at db, was speaking in terms of that.

frontend/src/Editor/Editor.jsx
akshaysasidrn: got it

akshaysasidrn:
Great work @gondar00 🙌🏽

akshaysasidrn (approved):

PR #1470, opened 2021-11-30
Improvement:  widget components from existing implementation to the new API

arpitnath:
Related #1475 
### Widgets
- [x] text input 
- [x] number input
- [x] checkbox
- [x] radio button
- [x] dropdown
- [x] Button


sherfin94 (changes requested):
@arpitnath Could you look into these things:

- The default value for number input is not working, I've made some changes on text input, refer to that so that its easier for you to get default value working here as well
- Radio button component still uses `currentState` and `component` props. Ensure that this is changed to new API, you can refer to the changes I made to text input
- Merge with latest develop

sherfin94 (approved):

Issue #1417, opened 2021-11-17
Incorrect placement for delete button on text input


gondar00:
![image](https://user-images.githubusercontent.com/12490590/142168296-d826eac0-03b0-4422-a420-93124bc2ae3b.png)

As per the image above the delete button (on hover) overlaps the text input, 

Possible solution: 
delete button should behave consistently as in other widgets, ex: on table widget appears on the widget drag handle

gondar00:
![image](https://user-images.githubusercontent.com/12490590/142168707-abdf7f35-59e9-46c5-9e72-7ab2d84692d7.png)

Also, observed for button

singh-vikas-m:
@gondar00 are you still working on this ?

gondar00:
Nope, I'm not. Would you like to take this up?

singh-vikas-m:
sure ✅

ivelin-web:
Can I work on this?

PR #1479, opened 2021-12-01
fix: Incorrect placement for delete button on text input 

singh-vikas-m:
fix for issue - https://github.com/ToolJet/ToolJet/issues/1417

The issue is with not having the default min-width of the ConfigHandle.jsx.
Added changes to theme.css now work with all small widgets.


arpitnath (changes requested):
Would you please revert the changes and use the recommended linter setup?

singh-vikas-m:
Sure !

Navaneeth-pk:
Closing a there is no update.

PR #1539, opened 2021-12-08
Chore: Setup pipeline

### LINKED ISSUE #746: Setup pipeline using github actions
Setup Github actions workflow to run build, lint and test stages for both client and server.
We should enable for public fork pull request runs as well.

akshaysasidrn:
Fixes #746


server/test/controllers/users.e2e-spec.ts
```diff
@@ -35,45 +37,8 @@ describe('users controller', () => {
...
-      expect(response.statusCode).toBe(403);
-
-      await user.reload();
-
-      expect(user.password).toEqual(oldPassword);
-    });
-  });
-
   afterAll(async () => {
+    await getConnection().close();
```
akshaysasidrn, line 41: Had to specify this to get the suite working, without this change app fails with [test suite](https://github.com/ToolJet/ToolJet/runs/4462789088?check_suite_focus=true#step:9:65) not being able to run.
But now with this change, I get error on the suite [suite](https://github.com/ToolJet/ToolJet/runs/4466987549?check_suite_focus=true#step:9:72).

Navaneeth-pk (approved):

PR #1558, opened 2021-12-10
[Bug Fix #1521] Home Page Pagination

### LINKED ISSUE #1521: Pagination need to support larger index
Pagination should handle large indexes

image

gsmithun4:
Related to https://github.com/ToolJet/ToolJet/issues/1521, 
Fixed `there is an extra page with no apps if we add a new app and delete the (21st app)`


akshaysasidrn:
Could you fix the lint error, please?

Navaneeth-pk (approved):

PR #1570, opened 2021-12-13
Fixes z index issue of date picker widget

Navaneeth-pk:
Bug: The datepicker widgets gets overlapped by other widgets once the month is changed.


frontend/src/Editor/Editor.jsx
```diff
@@ -784,7 +784,11 @@ class Editor extends React.Component {
               <div
                 className={`canvas-container align-items-center ${!showLeftSidebar && 'hide-sidebar'}`}
                 style={{ transform: `scale(${zoomLevel})` }}
-                onClick={() => this.switchSidebarTab(2)}
+                onClick={(e) => {
+                  if (e.target.className === 'real-canvas') {
```
sherfin94, line 788: This is preventing me from adding components inside modal, as the inspector panel wont close:
inspector-not-closing-inside-modal

PR #1575, opened 2021-12-14
[WIP] Widget: List view

Navaneeth-pk:
Adds a new widget to create lists of widgets.


frontend/src/Editor/Box.jsx
```diff
@@ -102,12 +106,24 @@ export const Box = function Box({
...
-
+  const resolvedProperties = resolveProperties(component, currentState, null, customResolvables);
+  const resolvedStyles = resolveStyles(component, currentState, null, customResolvables);
   resolvedStyles.visibility = resolvedStyles.visibility !== false ? true : false;
 
-  const exposedVariables = currentState?.components[component.name] ?? {};
+  let exposedVariables = {};
+
+  if (component.parent) {
+    const parentComponent = components[component.parent];
```
sherfin94, line 116: The variable `components` is not defined here. And I think the reason why its not causing an error is because the flow of execution is not reaching this area of the code, because, `component.parent` is always undefined on line 115. I'm not able to figure out why.

frontend/src/Editor/Box.jsx
```diff
@@ -102,12 +106,24 @@ export const Box = function Box({
...
-  const resolvedStyles = resolveStyles(component, currentState);
-
+  const resolvedProperties = resolveProperties(component, currentState, null, customResolvables);
+  const resolvedStyles = resolveStyles(component, currentState, null, customResolvables);
   resolvedStyles.visibility = resolvedStyles.visibility !== false ? true : false;
 
-  const exposedVariables = currentState?.components[component.name] ?? {};
+  let exposedVariables = {};
+
+  if (component.parent) {
```
sherfin94, line 115: This condition never passes when I try to run. No component's parent is getting correctly set on ListView. Probably because of that, the radio button `exposedVariables` are not being calculated correct. Thereby disabling it from correctly working.

Navaneeth-pk: Yes, this logic seems to be flawed. Working on it now.

PR #1587, opened 2021-12-15
Add error boundary for components

### LINKED ISSUE #1586: Add error boundary to components

Presently if some logic inside a component throws an exception, the entire platform crashes.

Make it so, that only the crashing component falls back to a fallback UI and the platform stays intact.

sherfin94:
Resolves #1586 

Refer to https://reactjs.org/docs/error-boundaries.html for code review

### DIFF
--- a/frontend/src/Editor/ErrorBoundary.jsx
+++ b/frontend/src/Editor/ErrorBoundary.jsx
@@ -19,7 +19,7 @@ class ErrorBoundary extends Component {
   render() {
     if (this.state.hasError) {
       // You can render any custom fallback UI
-      return this.props.showFallback ? <h1>Something went wrong.</h1> : <div></div>;
+      return this.props.showFallback ? <h2>Something went wrong.</h2> : <div></div>;
     }
 
     return this.props.children;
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/ErrorBoundary.jsx
```diff
@@ -0,0 +1,29 @@
...
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(error) {
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true };
+  }
+
+  componentDidCatch(error, errorInfo) {
```
akshaysasidrn, line 14: I think we should also consider logging in to sentry, if APM for sentry is enabled via env vars.
https://docs.sentry.io/platforms/javascript/guides/react/usage/#capturing-errors

gondar00 (approved):

frontend/src/Editor/ErrorBoundary.jsx
sherfin94: Will do this as a separate issue @akshaysasidrn .

sherfin94: https://github.com/ToolJet/ToolJet/issues/1619

Issue #1567, opened 2021-12-13
Ability to search apps on dashboard


akshaysasidrn:
### Is your proposal related to a problem?
Apps can't be searched on the homepage. We need to add a search for this.

### Describe the solution you'd like

We can add a basic search bar on the Homepage. This might need a bit of revision once the design is ready therefore a simple search bar would do.
![image](https://user-images.githubusercontent.com/11629675/145789215-2a273481-45ee-4fe4-adf3-a6fd14c875f8.png)

PR #1603, opened 2021-12-17
Home Page Application search box

gsmithun4:
Related to https://github.com/ToolJet/ToolJet/issues/1567

### DIFF
--- a/server/src/services/folders.service.ts
+++ b/server/src/services/folders.service.ts
@@ -31,8 +31,8 @@ export class FoldersService {
     );
   }
 
-  async all(user: User, searchKey: string): Promise<Folder[]> {
-    if (!searchKey && (await this.usersService.hasGroup(user, 'admin'))) {
+  async allFolders(user: User): Promise<Folder[]> {
+    if (await this.usersService.hasGroup(user, 'admin')) {
       return await this.foldersRepository.find({
         where: {
           organizationId: user.organizationId,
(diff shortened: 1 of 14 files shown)


server/src/services/folders.service.ts
```diff
@@ -106,16 +165,21 @@ export class FoldersService {
...
-      .andWhere('app_group_permissions.read = :value', { value: true })
-      .orWhere(
-        '(viewable_apps.is_public = :value AND viewable_apps.organization_id = :organizationId) ' +
-          'OR viewable_apps.user_id = :userId',
-        {
-          value: true,
-          organizationId: user.organizationId,
-          userId: user.id,
-        }
+      .where('LOWER(viewable_apps.name) like :searchKey', { searchKey: `%${searchKey && searchKey.toLowerCase()}%` })
```
akshaysasidrn, line 168: Can conditionally search based on the presence of the search key? Doing `like` queries always will be slower.

server/src/services/folders.service.ts
```diff
@@ -31,8 +31,8 @@ export class FoldersService {
     );
   }
 
-  async all(user: User): Promise<Folder[]> {
-    if (await this.usersService.hasGroup(user, 'admin')) {
+  async all(user: User, searchKey: string): Promise<Folder[]> {
```
akshaysasidrn, line 34: Can we extract out logic for getting folders when search is involved onto a separate function? This seems complex as of now.

akshaysasidrn (changes requested):

akshaysasidrn:
Found two bugs while testing out:
1. Folders not being listed when signed in as a non admin user having permission to view public apps.
2. Searching when having a folder selected will show results having `All applications` selected.

akshaysasidrn:
Also, please update e2e specs with search. Need to add test cases for search results to not show another organization apps and non-permissible ones as well.

akshaysasidrn (approved):
Looks good. Nice work.

Issue #1644, opened 2021-12-27
Tags widget


sherfin94:
### Is your proposal related to a problem?
Presently we don't have a tags widget.

### Describe the solution you'd like
Add a tags widget.

PR #1659, opened 2021-12-28
Feature/tags widget

shah21:
Related to #1644


sherfin94:
Lets make the tags have the same height as button component

image

sherfin94:
If the grid of tags goes outside the height of the widget, then there should be horizontal scroll:

image

sherfin94:
Lets have a little space between vertically stacked tags
image

sherfin94:
Each tag should have separate text color also, just like tag color.

image

sherfin94:
Finally, currently when a tag is created, it says tag1 tag2 etc, instead of that, lets have 4 tags that goes: 'success'(green), info (blue), warning (yellow) and danger(red)

sherfin94 (changes requested):
I've added the required changes as comments..

shah21:
> Finally, currently when a tag is created, it says tag1 tag2 etc, instead of that, lets have 4 tags that goes: 'success'(green), info (blue), warning (yellow) and danger(red)

You mean the default tags, right?

sherfin94:
image

Hey @shah21 , some changes are required,

- The scrollbar is unnecessarily shown, check screenshot, show it only if necessary
- The bottom corners of the tags are rounded, whereas the top corders are sharp, like a rectangle: Make it so, that all corners are rounded
- The tags now expand as we increase the height of the widget, lets just have the tags take the same height that they come in https://preview.tabler.io/docs/badges.html
- The initial default colors that we apply when tags widget is created, lets select from the colors shown in the page: https://preview.tabler.io/docs/badges.html, with the same labels that you've already set.

sherfin94 (changes requested):
@shah21 Two more items to go,

- Since we're accepting text color via data itself, the text color property on style section is not useful any more. Let's remove that.
- Specify in documentation that text color can also be set through data

sherfin94 (dismissed):

shah21:
> @shah21 Could you merge develop locally into this branch and push? There are are conflicts.

Okay,I will do that

Navaneeth-pk (approved):

PR #1735, opened 2022-01-04
Added different border shapes for image widget

### LINKED ISSUE #1734: Add different border shapes to an image

Now the image can only see in rectangle shape, So adding more border shapes is good

Should able to change the border shape to rounded, circle etc

Yes

shah21:
Resolves #1734

### DIFF
--- a/frontend/src/Editor/Components/components.js
+++ b/frontend/src/Editor/Components/components.js
@@ -748,7 +748,7 @@ export const componentTypes = [
         type: 'select',
         displayName: 'Border type',
         options: [
-          { name: 'None', value: 'rounded-0' },
+          { name: 'None', value: 'none' },
           { name: 'Rounded', value: 'rounded' },
           { name: 'Circle', value: 'rounded-circle' },
           { name: 'Thumbnail', value: 'img-thumbnail' },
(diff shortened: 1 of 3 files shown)


sherfin94 (changes requested):
Let's also write a migration that initiates the border type to default value.

`server/migrations/1633370361564-SetShowBulkSelectorToTrue.ts` is a good example you can refer to. It set a particular property for all instances of table in all applications. Instead, sets the style property you've defined to its initial value.

To create a new migration, go to the server directory, and run `npm run -- typeorm migration:create -n yourMigrationName`

frontend/src/Editor/Components/components.js
```diff
@@ -744,6 +744,16 @@ export const componentTypes = [
       onClick: { displayName: 'On click' },
     },
     styles: {
+      borderType: {
+        type: 'select',
+        displayName: 'Border type',
+        options: [
+          { name: 'None', value: 'rounded-0' },
```
sherfin94, line 751:
```suggestion
          { name: 'None', value: 'none' },
```

Let's go with `none` itself as the value for no border type. `rounded-0` is a bit confusing.

You might have to add a condition on the component's code to determine border type and will also have to update the migration for this.

Navaneeth-pk (approved):

PR #1741, opened 2022-01-05
Added radius style property to change the radius of action buttons in the table widget

### LINKED ISSUE #1740: Add a 'action button radius' style property to the table widget

Currently, we cannot change the radius of the action button in the table widget 

Add a style field to change the radius value of action buttons 

Yes

shah21:
Resolves #1740


sherfin94 (changes requested):
Could you also add documentation for this?

Documentation can be accessed by doing `npm install` and `npm start` on the `docs` directory.

Just add an entry to this table on `Table`'s doc page
image

sherfin94 (approved):

PR #1743, opened 2022-01-05
Bugfix - Import button style

### LINKED ISSUE #1742: Import button is not properly clickable
### Environment
MacOS

If we click anywhere on the Import button, it should open

We need to click on the "Import" text on the button to open

abhinabaadhikari:
Resolves #1742


sherfin94 (changes requested):
Thanks!. Could you also resolve the lint issues?

sherfin94 (approved):

PR #1748, opened 2022-01-06
Feature: Force logout user from session when archived

### LINKED ISSUE #1698: Force logout users when archived or group permissions changes
As of now, if there are changes in group permission or if the user gets archived this wont affect the currently logged in user as the permissions and status is only checked on login and will be cached in the browser.

Figure out an approach that either polls periodically or subscribes to these changes such that the cached information is updated or simply force logout user.

Yes

akshaysasidrn:
Fixes #1698 for archived user session termination
archive

### DIFF
--- a/frontend/src/App/App.jsx
+++ b/frontend/src/App/App.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import config from 'config';
 import { Router, Route } from 'react-router-dom';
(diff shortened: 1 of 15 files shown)


akshaysasidrn:
I'm closing this as we are handling the case with #1749. The session won't be terminated in real-time, but the user will get logged out in subsequent requests.

PR #1761, opened 2022-01-08
Add upper & lower limit to number input widget

### LINKED ISSUE #1760: Support Upper & Lower limit on Number input widget

Presently the number input widget can go from negative infinity to positive infinity, and there’s no way to set a limit.

The widget should have properties for maximum and minimum.

Yes.

abhinabaadhikari:
Resolves #1760

### DIFF
--- a/docs/docs/widgets/number-input.md
+++ b/docs/docs/widgets/number-input.md
@@ -10,6 +10,8 @@ Numbers can be changed by using the arrow keys.
 
 #### Properties
 
-| properties      | description |
-| ----------- | ----------- |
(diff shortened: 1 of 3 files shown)


sherfin94:
@abhinabaadhikari Could you also add documentation for this?

You just need to add these two properties to `ToolJet/docs/docs/widgets/number-input.md`

Documentation can be run by these following commands:

`cd docs`
`npm install`
`npm start`

Remember to stop the server while starting docs server.

sherfin94 (approved):

PR #1763, opened 2022-01-08
Removed class text-muted from Library card text

### LINKED ISSUE #1134: Card text in Library is not legible in dark mode
Current state:
image
Expected state:
image

ivelin-web:
Resolves #1134

I remove class `text-muted` from Library card text, because text is not legible in dark mode. When class `text-muted` is removed everything looks right

With class `text-muted` :

Without class `text-muted` :


Navaneeth-pk:
Thanks for contributing @ivelin-web.

Navaneeth-pk (approved):

PR #1774, opened 2022-01-10
Toggle buttons work even if we click far outside the input

### LINKED ISSUE #1775: Toggle buttons work even if we click far outside the input
### Version Information
Latest
### Environment

MacOs

Toggle buttons should only work if the user clicks on the button.

Toggle buttons work even if we click far outside the input
### Screenshots or Screencast

https://user-images.githubusercontent.com/18279004/148727759-659a1b1b-aea4-4de1-94af-e8c0d55b3746.mov
Yes

shah21:
Fixes #1775


sherfin94 (approved):
Nice!

Issue #1786, opened 2022-01-10
[Home Page Design Changes] Update icon API


gsmithun4:
### Is your proposal related to a problem?
As per the [New Design](https://www.figma.com/file/kviZGq1lsML1Mi10OHYftU/1st-cycle-(Nov-jan)?node-id=486%3A8446), we have to build API to update icon

### Describe the solution you'd like
Create PUT API : `/api/apps/:id/icons`

### If the feature is approved, would you be willing to submit a PR?
Yes

PR #1787, opened 2022-01-10
[Home Page Design Changes] Update icon API

gsmithun4:
Resolves https://github.com/ToolJet/ToolJet/issues/1786

### DIFF
--- a/frontend/src/HomePage/Folders.jsx
+++ b/frontend/src/HomePage/Folders.jsx
@@ -84,7 +84,7 @@ export const Folders = function Folders({
               </div>
             )}
           </div>
-          {folders && folders.length < 0 ? (
+          {folders && folders.length > 0 ? (
             folders.map((folder, index) => (
               <a
                 key={index}
--- a/frontend/src/HomePage/Folders.jsx
+++ b/frontend/src/HomePage/Folders.jsx
@@ -80,25 +80,30 @@ export const Folders = function Folders({
             <div className="folder-info">Folders</div>
             {canCreateFolder && (
               <div className="folder-create-btn" onClick={() => setShowForm(true)}>
-                + Create New Folder
+                + Create new folder
               </div>
             )}
           </div>
-          {folders.map((folder, index) => (
-            <a
-              key={index}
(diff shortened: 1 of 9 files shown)


frontend/src/HomePage/Folders.jsx
```diff
@@ -80,25 +80,30 @@ export const Folders = function Folders({
             <div className="folder-info">Folders</div>
             {canCreateFolder && (
               <div className="folder-create-btn" onClick={() => setShowForm(true)}>
-                + Create New Folder
+                + Create new folder
```
Navaneeth-pk, line 83: 👍

server/src/controllers/apps.controller.ts
```diff
@@ -275,9 +274,7 @@ export class AppsController {
 
   @UseGuards(JwtAuthGuard)
   @Put(':id/versions/:versionId')
-  async updateVersion(@Request() req, @Param() params) {
-    const definition = req.body['definition'];
-
+  async updateVersion(@Request() req, @Param() params, @Body('definition') definition) {
```
akshaysasidrn, line 277: Nice, didn't know we could infer params from body as such.

server/src/controllers/apps.controller.ts
```diff
@@ -288,4 +285,18 @@ export class AppsController {
     const appUser = await this.appsService.updateVersion(req.user, version, definition);
     return decamelizeKeys(appUser);
   }
+
+  @UseGuards(JwtAuthGuard)
+  @Put(':id/icons')
+  async updateIcon(@Request() req, @Param() params, @Body('icon') icon) {
```
akshaysasidrn, line 291: Can't we do this update in `PUT /api/apps/:id`? Do we need to add a separate endpoint for this?

gsmithun4: Since there is a plan to expand this feature, and enable custom icons which user can upload, I think its better to go with separate end point

akshaysasidrn: Alright then

akshaysasidrn (approved):

PR #1798, opened 2022-01-11
Bugfix: resolve missing query associations on importing app

### LINKED ISSUE #1797: Run-query event associations are not preserved while importing app
### Environment

Mac OS

When an app is imported, the run query actions should keep their associations with queries.

The queries are not associated properly when an app is imported.

sherfin94:
Resolves #1797


akshaysasidrn (approved):

akshaysasidrn:
Looks good.

PR #1812, opened 2022-01-13
fix: Query manager panel doesn't resize smoothly 

### LINKED ISSUE #1811: Resizing of the query manager panel is not smooth
### Version Information
Latest

### Environment

MacOs

Resizing of query panel should be smooth 

 - Now the panel is resizing with a jumping start 
 - Also need to click outside of the panel's border to enable resize pointer

1. Try to resize the height of query manager m]panel

### Screenshots or Screencast

https://user-images.githubusercontent.com/18279004/149288558-d71b2195-5dcd-434f-b6b3-84dfd186f36b.mov

### Any possible solutions?
Add a transition effect to the container

Editor component 

Yes

shah21:
Fixes #1811

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -72,6 +72,7 @@ class Editor extends React.Component {
       allComponentTypes: componentTypes,
       isQueryPaneDragging: false,
       queryPaneHeight: 70,
+      isTopOfQueryPane: false,
       isLoading: true,
       users: null,
       appId,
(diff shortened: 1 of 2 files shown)


sherfin94 (changes requested):
When I tried, the query panel is moving only when mouse drag is complete, also its very slow. Let's make it so, that the panel is dragged along with cursor. It might help to sync up with @gondar00 , he has very good context on this area.

https://user-images.githubusercontent.com/20333972/149303246-eca5fe9b-d929-4eee-8211-7be942360204.mov

shah21:
> When I tried, the query panel is moving only when mouse drag is complete, also its very slow. Let's make it so, that the panel is dragged along with cursor. It might help to sync up with @gondar00 , he has very good context on this area.
> 
>  Screen.Recording.2022-01-13.at.2.54.14.PM.mov

Oh, that was because of the CSS transition effect. will remove it

Navaneeth-pk (approved):

PR #1824, opened 2022-01-14
Fix: window title should change to app name when the user is on the editor page

### LINKED ISSUE #1823: Change the window title to app name when the user is on the editor page

Currently, when opening the app editor, we can see that the title shows 'Toojet - Dashboard'. So instead of that show the current app name 
  
Change the document title name

Yes

shah21:
Fixes #1823

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -105,6 +105,10 @@ class Editor extends React.Component {
     };
   }
 
+  setWindowTitle(name) {
+    document.title = name ? `Tooljet - ${name}` : `Tooljet - Untitled App`;
+  }
+
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


frontend/src/Editor/Editor.jsx
```diff
@@ -105,6 +105,10 @@ class Editor extends React.Component {
     };
   }
 
+  setWindowTitle(name) {
+    document.title = name ? `Tooljet - ${name}` : `Tooljet - Untitled App`;
```
Navaneeth-pk, line 109: Can we reverse the order? `<app name> - ToolJet`?

shah21: Sure

Navaneeth-pk (approved):

PR #1834, opened 2022-01-14
Feature : statistics widget

### LINKED ISSUE #1833: Statics widget addition
There is no statistics widget currently

Add new statistics widget
Properties included could be values and label of the item ex : weight , 63kg
style change in colour for fields can be included

stepinfwd:
closes #1833


Navaneeth-pk:
image
image

The width of the widget should fit the bounding box. 
@stepinfwd

Navaneeth-pk:
image

Shouldn't the text `-2.85%` be in red?

sherfin94:
Aparat from the issues that Navaneeth pointed out, there are some more issues @stepinfwd , I tried to create a statistic that shows how my weight went down to 115 from 118.5

image

As you can see, weight is expressed in $, there is no way for me to show that it's weight, and not money. **Therefore, remove the mandatory dollar sign. If the user wants it, they can attach it on the text itself.**

Similarly, you're attaching `+` or `-`  and `%` in the secondary value, **here also do not attach anything, if the user wants, they can type it on the text itself.**

sherfin94:
This might probably be a miscommunication from my part, the widget name was supposed to be `Statistic` widget, not `Static` widget. Please rename it on the UI, and also on the variable names.

sherfin94:
> image image
> 
> The width of the widget should fit the bounding box. @stepinfwd

@stepinfwd I think there might have been a miscommunication in our conversation. The widget(the white container that covers the set of text elements) always has to expand to the bounding box(blue border box), the content inside the widget (the set of text elements, in this case), should be centered horizontally and vertically.

We can discuss this on a call if you need more clarification

sherfin94 (changes requested):
Apart from the comments that were left, please also, add support for dark mode, and locally merge the develop branch into this branch and push as there are conflicts.

frontend/src/Editor/Components/Statistics.jsx
```diff
@@ -0,0 +1,72 @@
...
+        fontWeight: '700'
+    }
+
+    return (
+        <div className="" style={baseStyle}>
+            <p style={{ ...letterStyle, color: darkMode?'#FFFFFC' :PrimaryLabelColour }}>{primaryValueLabel}</p>
+            <h2 style={priceStyle} >{primaryvalue}</h2>
+            <div>
+                <div className="d-flex flex-row ">
+                    {secondarysigndisplay == 'positive' ? < img src='../../../assets/images/icons/widgets/upstatics.svg' style={{ ...marginStyle, marginRight: '6.5px' }} /> : <img src='../../../assets/images/icons/widgets/downstatics.svg' style={{ ...marginStyle, marginRight: '6.5px' }} />}
```
Navaneeth-pk, line 65: Can we use `` instead of `../../assets` ?

stepinfwd: sure will update

frontend/src/Editor/Components/Statistics.jsx
```diff
@@ -0,0 +1,72 @@
+import React, { useEffect, useState } from 'react';
+export const Statistics = function Statistics({ height, properties, styles, darkMode }) {
+    const { primaryValueLabel, primaryvalue, secondaryValueLabel, secondaryvalue, secondarysigndisplay } = properties;
```
Navaneeth-pk, line 3: Please use consistent naming conventions. `secondaryvalue, secondarysigndisplay` -> `secondaryValue, secondarySignDisplay`

Navaneeth-pk:
image

This needs to be a `codehinter` field instead of a dropdown. Otherwise, we won't be able to programmatically set if the sign should be negative or positive.

Navaneeth-pk (approved):

PR #1859, opened 2022-01-18
Fix: datasource Icons doesn't show on templates model

### LINKED ISSUE #1858: Datasource icons of each template aren't showing on the template model
### Version Information
Latest

### Environment

MacOs

Each template tab should show a data source icon that the template uses. 

Now the implementation had been done. but icons don't visible

1. Open template model
2. Switch between templates

### Screenshots or Screencast

Screenshot 2022-01-18 at 1 33 29 PM

### Any possible solutions?
Check if the path is incorrect or not

Yes

shah21:
Fixes #1858

### DIFF
--- a/frontend/src/HomePage/TemplateLibraryModal/TemplateDisplay.jsx
+++ b/frontend/src/HomePage/TemplateLibraryModal/TemplateDisplay.jsx
@@ -1,15 +1,9 @@
 import React from 'react';
 import { Container, Row, Badge } from 'react-bootstrap';
-import { allSvgs } from '@tooljet/plugins/client';
+import { getSvgIcon } from '@/_helpers/appUtils';
 
 export default function TemplateDisplay(props) {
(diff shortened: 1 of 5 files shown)


frontend/src/HomePage/TemplateLibraryModal/TemplateDisplay.jsx
```diff
@@ -1,8 +1,16 @@
 import React from 'react';
 import { Container, Row, Badge } from 'react-bootstrap';
+import { allSvgs } from '@tooljet/plugins/client';
 
 export default function TemplateDisplay(props) {
   const { id, name, description, sources } = props?.app ?? {};
+
+  const getSvgIcon = (key, height = 50, width = 50) => {
```
Navaneeth-pk, line 8: Looks like there are other instances of the same function within the app. Can we move it to utils to avoid duplication?

Navaneeth-pk (approved):

PR #1899, opened 2022-01-23
Allow dropdown option to be selected on Enter key press (Query Manager)

### LINKED ISSUE #1663: [Query manager] Select dropdown option on enter pressed
image

Expected: 
When enter/return key is pressed the hovered option should get selected

marboleda:
Should resolve #1663

The dropdown menu that does not select on Enter key when hovering was the SelectSearch component from the react-select-search package. 

Since its implementation is in a separate codebase, I could not make a code change to it.

Instead, I replaced it with a dropdown menu from a different package already in the project (react-select) which does allow for selecting by hovering and pressing Enter key.


Navaneeth-pk:
@marboleda can you please screenshots since this is a UI change?

marboleda:
image

Navaneeth-pk:
@marboleda please fix the lint errors.

Navaneeth-pk (approved):

Navaneeth-pk:
Please ignore my last comment. The lint issues were not from this PR. Thanks for contributing!

PR #1944, opened 2022-01-25
Added an organization 'role' field to the page that redirects after the email confirmation

shah21:


Navaneeth-pk:
@shah21 please check the failing tests.

PR #1948, opened 2022-01-26
Bugfix/Fixing text overflow issues in table for long text

### LINKED ISSUE #1945: Long text overflow from cell in table widget
text to be wrapped or be scrollable in cell
new option to be added to include scroll, text wrapping so that text does not overflow.

long text / sentences overflows in table.

1.drag and drop table component
2.write long  text in any of the fields 
3.text can be seen overflowing from the table cell

### Screenshots or Screencast
Screenshot 2022-01-26 at 1 51 54 AM

### Any possible solutions?

option/property can be included as a dropdown to set behaviour of text with options like wrap , scroll etc.

stepinfwd:
closes #1945


frontend/src/Editor/Inspector/Components/Table.jsx
```diff
@@ -544,6 +561,8 @@ class Table extends React.Component {
     const columns = this.props.component.component.definition.properties.columns;
     const column = columns.value[index];
 
+    console.log('col', column);
```
Navaneeth-pk, line 564: Please remove.

frontend/src/_styles/theme.scss
```diff
@@ -1447,7 +1447,12 @@ border-radius: 4px;
   .has-actions {
     margin: 0;
   }
-
+.text-wrapper{
```
Navaneeth-pk, line 1450: Indentation looks incorrect here.

sherfin94:
The text wrap option is shown for column types for which it does not apply, show it only for default and string

image

sherfin94:
I think the name of the property should be `overflow` and options should be `wrap`, `hide` and `scroll`

sherfin94:
While using the scroll option, a scroll bar should be shown when that cell is hovered and the cell is overflown. Right now only mac users will be able to use the scroll, correct me if I'm wrong.

sherfin94:
Using the `hide` option destroys the table's structure: 
image

sherfin94 (changes requested):
Commented some issues.

stepinfwd:
issues resolved

sherfin94:
The scroll and wrap options work correctly, however the hide option hides the entire column instead of just hiding the part of the cell that overflew, please check:


https://user-images.githubusercontent.com/20333972/154009124-4e7131e4-87a0-484b-85bf-e3baa643fbef.mov

sherfin94:
Also, the `Text` column type does not require overflow options, it is specifically designed to contain big volumes of text

sherfin94 (changes requested):
Few changes requested, otherwise everything looks fine.

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -760,6 +772,7 @@ export function Table({
             ),
             Cell: ({ row }) => (
               <div className="d-flex flex-column align-items-center">
+                ]
```
sherfin94, line 775: @stepinfwd Dude what's this 😅 

image

sherfin94: Other than this I don't see any issue, remove this and merge the latest develop. And then let's merge it.

stepinfwd: updated

sherfin94 (approved):

PR #1961, opened 2022-01-27
Bugfix/table searchbox getting removed from ui on disabling and disabling seachbox on server side pagination

### LINKED ISSUE #1951: The search box and total count in table widget should be removed if server side pagination selected
The search box and total count in table widget is not removed when server side pagination selected

The search box and total count in table widget should be removed if server side pagination selected

1.drag and drop table widget
2.go to properties and toggle server side pagination  option
3.observe behaviour

### Any possible solutions?

updated to hide search box on if server side pagination is true

### LINKED ISSUE #1952: Show Search Box options gets removed if disabled in table widget
Show Search Box options gets must not get removed from ui upon toggling it.

Show Search Box options gets removed on disable option.

1.drag and drop table widget
2.disable Show Search Box option from options.
3.observe option gets removed from ui.

stepinfwd:
closes #1951 
closes #1952

### DIFF
--- a/frontend/src/Editor/Inspector/Components/Table.jsx
+++ b/frontend/src/Editor/Inspector/Components/Table.jsx
@@ -728,11 +728,10 @@ class Table extends React.Component {
 
     options.map((option) => renderOptions.push(renderCustomElement(option)));
 
-    const conditionalOptions = [
-      { name: 'clientSidePagination', condition: !serverSidePagination },
-      { name: 'loadingState', condition: true },
-    ];
+    const conditionalOptions = [{ name: 'clientSidePagination', condition: !serverSidePagination }];
     !serverSidePagination && conditionalOptions.push({ name: 'displaySearchBox', condition: true });
+    conditionalOptions.push({ name: 'loadingState', condition: true });
+
     conditionalOptions.map(({ name, condition }) => {
       if (condition) renderOptions.push(renderCustomElement(name));
     });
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/Inspector/Components/Table.jsx
```diff
@@ -730,7 +730,7 @@ class Table extends React.Component {
 
     const conditionalOptions = [
       { name: 'clientSidePagination', condition: !serverSidePagination },
-      { name: 'displaySearchBox', condition: displaySearchBox },
```
gsmithun4, line 733: Shall we hide `Show search box` option if server side pagination is selected?

frontend/src/Editor/Inspector/Components/Table.jsx
```diff
@@ -730,7 +730,7 @@ class Table extends React.Component {
...
+      !serverSidePagination && { name: 'displaySearchBox', condition: true },
```
gsmithun4, line 733: Let's keep same type of object in the array, this will add `false` as second element of the array, if server side pagination is enabled.
`!serverSidePagination && conditionalOptions.push({ name: 'displaySearchBox', condition: true })` Conditionally push the value.

gsmithun4 (dismissed):
LGTM

stepinfwd:
> LGTM

havent fixed this there is still some issue ,

PR #1970, opened 2022-01-27
Add clone component functionality

gondar00:
This PR adds a keyboard shortcut `cmd + d` to allow a component to be cloned __when__ the widget **inspector** is open.


sherfin94:
Few issues that I've noticed:

- If same component is cloned twice or more, all clones has the same name
- If cloned inside container, component comes outside container
- Putting anything inside a modal crashes the modal, but I'm not sure if this issue is only in this PR
- If we clone a container, widgets inside the container are not cloned

PR #2006, opened 2022-01-28
[Bugfix] component names should be generated uniquely

### LINKED ISSUE #2005: Widget component name should be unique
Widget components can have the same name, e.g., a button widget and datepicker widget can have the same name `button1`

1. Add two widgets from the widget manager
2. Edit or update  one of the widget's name to another

Screenshot 2022-01-29 at 1 49 20 AM

arpitnath:
Fixes #2005 

### Loom
https://www.loom.com/share/b2f4d062ee9f462a88477030d4c87b33


sherfin94 (changes requested):
I'm still able to create widgets with same name: 

https://user-images.githubusercontent.com/20333972/152330121-a59a50e1-bb91-45dc-a803-571f9c61e471.mov

EDIT: ignore as it will be resolved in separate PR

sherfin94:
Works fine. I think its better to show the alert only when the user presses `enter` or `onBlur`

arpitnath:
> I'm still able to create widgets with same name:
> 
>  Screen.Recording.2022-02-03.at.4.27.44.PM.mov

this is resolved, will fix it in a separate pr as it is a different issue

frontend/src/_hooks/use-Focus.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const useFocus = () => {
```
sherfin94, line 3: Nice 🙌🏼

sherfin94 (approved):

Navaneeth-pk (approved):

PR #2059, opened 2022-02-01
Fx button for code hinter fields that will flip them to corresponding UI components

### LINKED ISSUE #2067: Add support for managing any widget property field using code hinter

Presently, there are several property fields for widgets that are simply not programmable.
For instance, there is no way to dynamically set the color of buttons.

Add an `Fx` button to every property field, clicking on which, converts it into a code field.

Yes

sherfin94:
Resolves #2067


frontend/src/Editor/Inspector/Elements/Color.jsx
```diff
@@ -16,6 +16,8 @@ export const Color = ({ param, definition, onChange, paramType, componentMeta })
   const paramMeta = componentMeta[paramType][param.name] || {};
   const displayName = paramMeta.displayName || param.name;
 
+  console.log({ paramMeta });
```
Navaneeth-pk, line 19: Please remove this line.

Navaneeth-pk (approved):

PR #2120, opened 2022-02-03
Fix: Autosave definition on version release

### LINKED ISSUE #2119: Releasing version right after upating definition loses last change
The updated changes on the canvas must be saved as the app definition before release

Autosave has a debounce of 3s seconds and thus when user releases updated canvas within 3s the changes done will be lost

1. Open editor
2. make change on the editor
3. click on release quickly

### Any possible solutions?
Perform explicit save before releasing version

Yes

akshaysasidrn:
Fixes #2119

### DIFF
--- a/frontend/src/Editor/ReleaseVersionButton.jsx
+++ b/frontend/src/Editor/ReleaseVersionButton.jsx
@@ -9,11 +9,13 @@ export const ReleaseVersionButton = function DeployVersionButton({
   isVersionReleased,
   fetchApp,
   onVersionRelease,
+  saveEditingVersion,
 }) {
   const [isReleasing, setIsReleasing] = useState(false);
 
   const releaseVersion = (editingVersion) => {
     setIsReleasing(true);
+    saveEditingVersion();
     appService
       .saveApp(appId, {
         name: appName,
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/ReleaseVersionButton.jsx
```diff
@@ -9,11 +9,13 @@ export const ReleaseVersionButton = function DeployVersionButton({
   isVersionReleased,
   fetchApp,
   onVersionRelease,
+  saveEditingVersion,
 }) {
   const [isReleasing, setIsReleasing] = useState(false);
 
   const releaseVersion = (editingVersion) => {
     setIsReleasing(true);
+    saveEditingVersion();
```
Navaneeth-pk, line 18: Should there be an `await` ?

akshaysasidrn: Cant make use await here, as that function comprises of setState calls which don't return promises.
We would have to change this up to handle with multiple callbacks to ensure it runs serially. 

But if I'm not wrong having this run concurrently is fine as one API call is to update the current definition on the canvas and the other to update the current version id and there isn't any dependency on their order of execution.

Navaneeth-pk (approved):

PR #1947, opened 2022-01-26
[improvement] Query Panel - new design

arpitnath:
Resolves #1949 

Todo: 
- [x] Handle responsiveness related to query pane 
- [x] Add images/svgs for the selection component list
- [x] Handle/Fix dark theme for selection component
- [x] Update all other datasources' queries UI (should be consistent with Airtable query UI)
- [x] Fix minor bugs

PR #2135, opened 2022-02-04
[Chore] query pane header title to lowercase

arpitnath:
Related #1947


arpitnath:
Not related after discussion

PR #2182, opened 2022-02-07
Bugfix :: wraping all text in statistics widget

### LINKED ISSUE #2089: Bugfix :: ui bugs in statistics widget
test to be wrapped for every value and label in widget
improvement : primary value can be made number only
dark theme border colour to be changed 

Text is not wrapped , it overflows outside
Primary value  is string
Border colour in dark theme not as specified in design

1.Write long text inside statistics widget
2.Write string inside primary value
3.Observe border colour in dark theme

### Screenshots or Screencast
Screenshot 2022-02-01 at 11 31 39 AM

stepinfwd:
closes #2089


Navaneeth-pk:
image
The primary value is being cut off.

sherfin94:
The arrow icon which was properly positioned, does not look like its properly positioned anymore.

before: 
image

after: 
image

sherfin94 (changes requested):
I've commented one issue, otherwise looks good 👍

stepinfwd:
updated @sherfin94

sherfin94 (approved):

PR #2287, opened 2022-02-18
adds the not equals and not matches filters to the table widget

kfelter:
This is my first nodejs PR so any suggestions would be greatly appreciated!


Navaneeth-pk:
Please sign the CLA so that we can proceed with merging the PR.
@kfelter

kfelter:
Signed

PR #2337, opened 2022-02-24
Integration couchDB

### LINKED ISSUE #2336: Feature:: integration of couchDB
 Currently   couchDB is not supported by Tooljet.

Build plugin to integrate couchDB

**Operations**

- Update document
- Delete document
- Get document by ID
- Create document
- List records
- Find
- Get view

**Authentication options**

Use basic auth

Fields

- Host
- Port
- User
- Password
- Databse

Yes

stepinfwd:
resolves #2336


docs/docs/data-sources/couchdb.md
```diff
@@ -0,0 +1,11 @@
+
+# couchdb
+
+ToolJet can connect to CouchDB databases to read and write data. 
+
+- [Connection](#connection)
+- [Getting Started](#querying-couchdb)
+
+## Connection
```
akshaysasidrn, line 9: Please update documentation here

plugins/packages/couchdb/lib/index.ts
```diff
@@ -0,0 +1,177 @@
+import {
+  QueryError,
+  QueryResult,
+  QueryService,
+  ConnectionTestResult,
+} from "@tooljet-plugins/common";
+import { SourceOptions, QueryOptions } from "./types";
+import got from "got";
+// const NodeCouchDb = require('node-couchdb');
```
akshaysasidrn, line 9: can remove if not used

akshaysasidrn (changes requested):

plugins/packages/couchdb/lib/index.ts
```diff
@@ -0,0 +1,177 @@
...
+      status: "ok",
+      data: result,
+    };
+  }
+
+  async testConnection(
+    sourceOptions: SourceOptions
+  ): Promise<ConnectionTestResult> {
+    const client = await this.getConnection(sourceOptions);
+    // const couch = new NodeCouchDb();
```
gsmithun4, line 154: Remove commented codes

plugins/packages/couchdb/lib/index.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    const client = got(`${host}:${port}/`);
+    // const couchAuth = new NodeCouchDb({
+    //   auth: {
+    //       user: 'login',
+    //       pass: 'secret'
+    //   }
+    // });
+
+    // return couchAuth;
```
gsmithun4, line 175: Should we return something?

plugins/packages/couchdb/lib/manifest.json
```diff
@@ -0,0 +1,110 @@
...
+    "name": "CouchDB",
+    "kind": "couchdb",
+    "exposedVariables": {
+      "isLoading": false,
+      "data": {},
+      "rawData": {}
+    },
+    "options": {
+      "username": {
+        "type": "string"
```
gsmithun4, line 16: option types are not used anymore, but should keep the encryption issue, 
```
"options": {
   "password": {
     "encrypted": true
   },
}
```
this is enough

plugins/packages/couchdb/lib/index.ts
```diff
@@ -0,0 +1,177 @@
...
+          result = JSON.parse(response.body);
+          break;
+        }
+
+        case "create_record": {
+          response = await got(`${protocol}://${host}:${port}/${database}`, {
+            method: "post",
+            headers: authHeader(),
+            json: {
+              records: JSON.parse(queryOptions.body),
```
gsmithun4, line 72: will be good if we use [JSON5](https://json5.org/) to parse user entered json string

gsmithun4 (changes requested):

plugins/packages/couchdb/lib/index.ts
```diff
@@ -0,0 +1,172 @@
...
+
+        case "create_record": {
+          response = await got(`${protocol}://${host}:${port}/${database}`, {
+            method: "post",
+            headers: authHeader(),
+            json: {
+              records: JSON5.parse(queryOptions.body),
+            },
+          });
+          result = JSON5.parse(response.body);
```
gsmithun4, line 78: No need to parse response using `json5` parse user inputs like `queryOptions.body`. Create a private function to parse json
refer https://github.com/ToolJet/ToolJet/blob/develop/plugins/packages/bigquery/lib/index.ts

gsmithun4 (changes requested):

gsmithun4 (dismissed):
LGTM

stepinfwd:
> @stepinfwd @gsmithun4 The `List records` query seems to be returning the documents in this format described in their documentation: image
> 
> It is not returning any additional data that I attach while creating record using the `create record` operation or from the dashboard. That is, it is returning the document that I create, but it is not returning any data that I attach to the document, other than the reserved fields like `id`, `key` and `value`(inside which there is only `rev`).
> 
> I'm not exactly sure if this is an issue or if it is indeed the behaviour this operation is supposed to work.



> @stepinfwd @gsmithun4 The `List records` query seems to be returning the documents in this format described in their documentation: image
> 
> It is not returning any additional data that I attach while creating record using the `create record` operation or from the dashboard. That is, it is returning the document that I create, but it is not returning any data that I attach to the document, other than the reserved fields like `id`, `key` and `value`(inside which there is only `rev`).
> 
> I'm not exactly sure if this is an issue or if it is indeed the behaviour this operation is supposed to work.

https://pouchdb.com/api.html#batch_fetch
pouchDB is similar to couchDB alldocs does not returning any attachments inherently ,also its not mentioned in docs of couchDB.in pouchDB docs a key "include_docs" is mentioned to return the attachments its working for couchDB too (default is false).

or we can use  this for getting attachments : https://docs.couchdb.org/en/3.2.0/api/database/bulk-api.html#db-design-docs

sherfin94 (approved):

PR #2338, opened 2022-02-24
Fix: Client entrypoint

### LINKED ISSUE #983: Add base URL setting for better reverse proxy support

My production server has been given a subdomain but not *.subdomain. I can't control that, so my reverse proxy forwards requests to various apps based on subdirectory. Tooljet doesn't have base url setting so I can't use it in prod properly.  

Add a base URL setting. For example that's how sonarr(a popular self-hosted PVR for series and tv shows) does it:
image

akshaysasidrn:
Fixes #983 partly
- Fixes react-router to honour base path when client is deployed in a domain subpath


frontend/webpack.config.js
```diff
@@ -9,7 +9,10 @@ const API_URL = {
   development: 'http://localhost:3000',
 };
 
+const ASSET_PATH = process.env.ASSET_PATH || '/';
+
 module.exports = {
+  entry: './src/index.jsx',
```
gondar00, line 15: I think this is not required as `HtmlWebpackPlugin` would handle injected of the `main.js` script file

akshaysasidrn: have updated @gondar00

gondar00 (approved):

Navaneeth-pk (approved):

PR #2349, opened 2022-02-25
Hotfix/fix action for dropdown

### LINKED ISSUE #2347: Actions are not triggered from Dropdown widget
The dropdown widget should trigger the action `On select`  when an option is selected.

No action is invoked for the dropdown `On select` event.
e.g., if we add event(s) for `On select` to trigger a query or fire up an alert when an option is selected.

1. Add Dropdown widget from the widget manager
2. Attach a query to the `On select` event in the Event handler or add an event to trigger a toast on selecting any options
3. Select an option from the dropdown

### Screenshots or Screencast

https://user-images.githubusercontent.com/67645175/155648544-4ae4c2b1-1976-44ae-b154-afcc750e45ef.mp4

arpitnath:
Fixes #2347 

https://user-images.githubusercontent.com/67645175/155649631-406e13e6-6eb0-4338-bd93-e9db7ec4ad5f.mp4

### DIFF
--- a/frontend/src/Editor/Components/DropDown.jsx
+++ b/frontend/src/Editor/Components/DropDown.jsx
@@ -153,8 +153,7 @@ export const DropDown = function DropDown({
             onChange={(selectedOption, actionProps) => {
               if (actionProps.action === 'select-option') {
                 setCurrentValue(selectedOption.value);
-                setExposedVariable('value', selectedOption.value);
-                fireEvent('onSelect');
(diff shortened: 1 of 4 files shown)


arpitnath:
` handleOnComponentOptionChanged` from editor.jsx could be returning the original promise and not the new promise returned by .then() as this is resolved to undefined as the function inside is not explicitly returning anything.
if we return the promise, solves all the cases. like the same issue is with the radio button widget

sherfin94 (approved):

PR #2380, opened 2022-02-28
[Hotfix] resolves crash for table cell types expecting input type array

arpitnath:
Fixes app crash for cell type multi-select, dropdown, badge, radio and badges of table widget which expects input type of array.
Also resolves app crash for existing field values for the same.


arpitnath:
Todo: move the function **validateInputType** to `/components/Table.jsx`

arpitnath:
Closing this.
Refer: #2507

PR #2381, opened 2022-02-28
fix alert when email already exists from

### LINKED ISSUE #2375: Incorrect error message when user email already exists
If we try to signup using an existing email id, the message shown is - invalid email, while it should be like - email already exists. 

image

image

Sync271:
Fixes #2375 
When the email already exists I should say "Email already exists" instead of "Invalid email"

### DIFF
--- a/frontend/src/SignupPage/SignupPage.jsx
+++ b/frontend/src/SignupPage/SignupPage.jsx
@@ -42,7 +42,7 @@ class SignupPage extends React.Component {
         this.setState({ isLoading: false, signupSuccess: true });
       },
       () => {
-        toast.error('Invalid email', {
+        toast.error('Email already exists', {
           position: 'top-center',
         });
         this.setState({ isLoading: false });


frontend/src/SignupPage/SignupPage.jsx
```diff
@@ -42,7 +42,7 @@ class SignupPage extends React.Component {
         this.setState({ isLoading: false, signupSuccess: true });
       },
       () => {
-        toast.error('Invalid email', {
+        toast.error('Email already exists', {
```
Navaneeth-pk, line 45: I think we need to display the error message in the response from server. The request can fail for many reasons.

Sync271: > I think we need to display the error message in the response from the server. The request can fail for many reasons.

Yeah, but how'd you check the error tho? I am not sure how to go about this, let me know if you have any suggestions

Navaneeth-pk: `server/src/services/auth.service.ts:61` - you might need to add some logic here to respond with a meaningful message if the email already exists.

Navaneeth-pk:
Just noticed that this issue is already assigned to someone. Please create PRs only for the issues that are assigned to you. 
@Sync271

PR #2393, opened 2022-03-01
Adding HELM charts for installation in Kubernetes

replicadse:
In this PR, I added some HELM charts I have used to bootstrap a DEV environment on a local Kubernetes cluster (I use `kind`). I thought it might be valuable for others, too. This chart is currently pinned to v1.2.2 - it might be a good idea to make this part of a CI pipeline if accepted.
Note: This is not necessarily a config that production grade and it comes with it's own PostgreSQL SS. Also, the database is not seeded when deployed. I did this by going into the pod with my shell and running `npm run db:seed`

Feedback to this PR is very welcome!


akshaysasidrn:
Hey @replicadse , thanks for the helm setup. But may I ask how exactly would this benefit running k8s locally for development? 🤔 
We do have a docker-compose setup to help devs get started with running the app locally with ease.

As I understand, helm charts would benefit if we were to deploy on k8s and have a bunch of configs to manage within our manifests. But I'd think it would be better to keep it simple for now with manually tweaking our manifests and then introduce helm charts when stuff gets complicated.

deploy/helm/templates/tooljet/service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: tooljet
+spec:
+  type: ClusterIP
```
akshaysasidrn, line 6: @replicadse Does it make sense to change this to a load balancer such that it can be publicly available?

replicadse: @akshaysasidrn I see what you mean.
I do not think (and in fact really hate it) when the default is a LoadBalancer. Why? In any cloud environment, this will generate a new static IP for some kind of cloud managed LB. This adds extra cost plus works around other things you have already set up maybe. Many people use traefik as reverse proxy and it handles TLS whereas a service directly pinned to an LB does not.
But: One thing we usually encounter is that the service type is configurable from the `values.yaml` file. Maybe we don't need to make that decision but let the user decide which service type he/she wants? Update follows...

replicadse: Update in commit [ac90f18](https://github.com/ToolJet/ToolJet/pull/2393/commits/ac90f182c5e41791315a26ced854dbd9bdc748f2)

akshaysasidrn (approved):

akshaysasidrn:
Thank you for the contribution @replicadse 🙌🏽

Issue #2372, opened 2022-02-28
Text alignment for dropdown list


abhinabaadhikari:
### Describe the solution you'd like

Text alignment feature for dropdown labels, similar to what we have on Text widget, where the selected label & the full list of labels would be aligned according to the type of alignment we select.

PR #2406, opened 2022-03-02
Feature/2372 - add text alignment option to dropdown

mslourens:
Add `justifyContent` option to `components.js` to justify the selected value and the list of the dropdown. Implements #2372 

image


sherfin94:
Hey @mslourens , we have to add default value for the property as well, otherwise when we create a new dropdown, the alignement style will be empty: 

image

You can add it to the `definition` section of the component's object in `components.js`

sherfin94 (changes requested):
Please set default value for alignment. EDIT: ignore, I will add it.

sherfin94 (approved):

PR #2418, opened 2022-03-03
FIX #2375

### LINKED ISSUE #2375: Incorrect error message when user email already exists

PiperGuy:
Incorrect error message when user email already exists

### DIFF
--- a/frontend/src/SignupPage/SignupPage.jsx
+++ b/frontend/src/SignupPage/SignupPage.jsx
@@ -41,8 +41,9 @@ class SignupPage extends React.Component {
         };
         this.setState({ isLoading: false, signupSuccess: true });
       },
-      () => {
-        toast.error('Invalid email', {
+      (e) => {
+        console.log(e);
+        toast.error(e?.error || 'Something went wrong!', {
           position: 'top-center',
         });
         this.setState({ isLoading: false });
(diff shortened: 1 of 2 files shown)


frontend/src/SignupPage/SignupPage.jsx
```diff
@@ -41,8 +41,9 @@ class SignupPage extends React.Component {
         };
         this.setState({ isLoading: false, signupSuccess: true });
       },
-      () => {
-        toast.error('Invalid email', {
+      (e) => {
+        console.log(e);
```
Navaneeth-pk, line 45: Please remove.

server/src/services/auth.service.ts
```diff
@@ -66,7 +72,16 @@ export class AuthService {
 
     const { email } = params;
     const organization = await this.organizationsService.create('Untitled organization');
-    const user = await this.usersService.create({ email }, organization, ['all_users', 'admin']);
+    let user;
+    try {
+      user = await this.usersService.create({ email }, organization, ['all_users', 'admin']);
```
Navaneeth-pk, line 77: I think it might be better to check for existence of an email before trying to create a new user and then throw an error than relying on error code.

Navaneeth-pk: @akshaysasidrn wdyut?

PiperGuy: > I think it might be better to check for existence of an email before trying to create a new user and then throw an error than relying on error code.

Actually, this makes more sense. I have patched this in the latest commit.

Navaneeth-pk (approved):

PR #2524, opened 2022-03-15
[ Enhancement ] :: Tabs widget bug fixes

### LINKED ISSUE #2523: [ Enhancement ] :: Tabs widget bugs 🐛

Tab widget enhancements
When we change the default tab field, the change should be  immediately reflected on the widget.
default tab field only accepts string value: make it reflect number values as well
expose the currentTab of tabs widget

When we change the default tab field, the change is not immediately reflected on the widget.
Default tab field only accepts string value.
CurrentTab/defaultTab of tabs widget is not exposed.

1.Drag and drop Tabs widget.
2.Observe above issues.

frontend/src/Editor/Components/Tabs.jsx

Yes 🚀

stepinfwd:
resolves #2523

### DIFF
--- a/frontend/src/Editor/Components/Tabs.jsx
+++ b/frontend/src/Editor/Components/Tabs.jsx
@@ -12,7 +12,6 @@ export const Tabs = function Tabs({
   currentState,
   removeComponent,
   setExposedVariable,
-  properties,
 }) {
   const widgetVisibility = component.definition.styles?.visibility?.value ?? true;
   const disabledState = component.definition.styles?.disabledState?.value ?? false;
--- a/frontend/src/Editor/Components/Tabs.jsx
+++ b/frontend/src/Editor/Components/Tabs.jsx
@@ -51,7 +51,6 @@ export const Tabs = function Tabs({
   };
 
   const parentRef = useRef(null);
-  console.log('parsed tab', parsedDefaultTab);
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/Components/Tabs.jsx
```diff
@@ -42,18 +51,27 @@ export const Tabs = function Tabs({ id, component, width, height, containerProps
   };
 
   const parentRef = useRef(null);
-
+  console.log('parsed tab', parsedDefaultTab);
```
Navaneeth-pk, line 54: Please remove this line.

frontend/src/Editor/Components/components.js
```diff
@@ -1491,7 +1491,7 @@ export const componentTypes = [
       visibility: { type: 'toggle', displayName: 'Visibility' },
       disabledState: { type: 'toggle', displayName: 'Disable' },
     },
-    exposedVariables: {},
+    exposedVariables: { defaultTab: '' },
```
Navaneeth-pk, line 1494: `defaultTab` is the initial tab. I don't think e should change it when the tab is changed.

stepinfwd: updated @Navaneeth-pk

sherfin94 (approved):

PR #2536, opened 2022-03-16
Feature: Added refresh token functionality to Rest API OAuth flow

### LINKED ISSUE #2532: Improvement: Add refresh token functionality to Rest API OAuth flow

Yes. Currently, the OAuth flow doesn't have any refresh token functionality. Because of that related queries will return error when the access token expires

Add refresh token functionality in Rest API data source 
Yes

shah21:
Resolves #2532


plugins/packages/restapi/lib/index.ts
```diff
@@ -61,20 +61,31 @@ export default class RestapiQueryService implements QueryService {
...
     if (requiresOauth) {
       const tokenData = sourceOptions['tokenData'];
 
       if (!tokenData) {
         const tooljetHost = process.env.TOOLJET_HOST;
-        const authUrl = `${sourceOptions['auth_url']}?response_type=code&client_id=${sourceOptions['client_id']}&redirect_uri=${tooljetHost}/oauth2/authorize&scope=${sourceOptions['scopes']}`;
+        const authUrl = `${sourceOptions['auth_url']}?response_type=code&client_id=${
+          sourceOptions['client_id']
+        }&redirect_uri=${tooljetHost}/oauth2/authorize&scope=${sourceOptions['scopes']}&access_type=offline${
+          customParams['prompt'] ? '&prompt=consent' : ''
```
akshaysasidrn, line 87: I think rather than hardcoding by checking for presence, it will be better to simply append all custom auth params as query params?

akshaysasidrn: But in Ui we don't currently have a separation b/w what's sent as json body and query params. But that can be taken up to be fixed separately, and invalid params would be ignored. Might need to check that.

shah21: So now go with append custom auth params right. we are also providing the custom params to JSON body in fetchOauthToken function. maybe it will ignore as you said.

akshaysasidrn: Yeah. Let's check it against google oauth endpoint once. We can iterate upon the UI to fix this separately if so.

akshaysasidrn:
@shah21 Some e2e tests are failing. Could you check them please?

akshaysasidrn (approved):

akshaysasidrn:
@shah21 I think we should separate JSON and query params to prevent any errors on API endpoints that whitelists/validate params. And ensuring backward compatibility. 
Could you open up an issue for this and help pick it up? Let's put this PR on hold until then.

shah21:
> @shah21 I think we should separate JSON and query params to prevent any errors on API endpoints that whitelists/validate params. And ensuring backward compatibility. Could you open up an issue for this and help pick it up? Let's put this PR on hold until then.

Yeah sure

akshaysasidrn:
@shah21 please update with the latest develop

akshaysasidrn (approved):

PR #2551, opened 2022-03-16
Feature: port multi select to react select lib

### LINKED ISSUE #2550: Improvement: Change multi-select to react-select

Port multi-select to react select component
- Also this widget should contain an optional option for selecting/unselecting all options

Replace current select lib with react-select 

Yes

shah21:
Resolves #2550


Moode09 (dismissed):

sherfin94:
The exposed variable `values` presently is an array of objects, whereas it used to be an array of values of the selected options. Please change it back to that, for backward compatibility

image

sherfin94:
The close button width seems to be a bit out of bounds
image

sherfin94:
Please make the height of the selected options smaller and fixed, so that even if there are many selected options, it can be properly fit (given that the developer has given enough height for the multiselect).

That is, it should be able to look like this, if there are many selected options:
image

sherfin94:
Would it be possible to not include `All` as a selection on the multiselect when it is selected and only show it on the options? To deselect all, we already have the `X` button on the right side.

image

sherfin94 (changes requested):
I've mentioned some changes on the comments.

Moode09 (approved):

stepinfwd:
updated changes @sherfin94

sherfin94 (approved):

PR #2562, opened 2022-03-17
Feature: Added a feature to see raw query responses

### LINKED ISSUE #2561: Add a feature to show raw query responses

Currently, We can only see JSON responses, what if a request returns other than JSON like HTML, Raw text, etc. 

Add tab or another option to see raw responses. Users should be able to switch from JSON to raw and vice versa for valid responses.

Yes

shah21:
Resolves #2561

### DIFF
--- a/frontend/src/Editor/QueryManager/Preview.jsx
+++ b/frontend/src/Editor/QueryManager/Preview.jsx
@@ -1,0 +1,71 @@
+import React, { useEffect } from 'react';
+import { JSONTree } from 'react-json-tree';
+import { Tab, ListGroup, Row } from 'react-bootstrap';
+
+const Preview = ({ previewPanelRef, previewLoading, queryPreviewData, theme, darkMode }) => {
+  const [key, setKey] = React.useState('raw');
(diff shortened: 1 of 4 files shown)


sherfin94 (changes requested):
Few things to be done @shah21 ,

1. With this feature, we need to be able to query URLs that does not respond in JSON format as well, for example, we need to be able to hit `https://www.google.com/`
2. The raw data preview text is not copiable at the moment, it needs to be copiable
3. If the response is not json, make it available as plain text on `queries.queryName.data` itself.

shah21:
> Few things to be done @shah21 ,
> 
> 1. With this feature, we need to be able to query URLs that does not respond in JSON format as well, for example, we need to be able to hit `https://www.google.com/`
> 2. The raw data preview text is not copiable at the moment, it needs to be copiable
> 3. If the response is not json, make it available as plain text on `queries.queryName.data` itself.

The first one is properly working I think. It can hit the non-JSON URLs and show the raw HTML content instead.

sherfin94 (approved):

PR #2577, opened 2022-03-19
fix: ignore OAuth hd field when SSO_RESTRICTED_DOMAIN not set

### LINKED ISSUE #2563: Google SSO fails for users without Google hosted domain
### Version Information

1.4.0

### Environment

Docker

Google SSO should succeed for any Google Account, assuming that configuration variables are set correctly and the OAuth consent screen is configured as "External" in the Google Cloud Platform console.

Google SSO fails when the `hd` key of the OAuth response is omitted, which is the case for any user not belonging to a hosted domain. From [https://developers.google.com/identity/protocols/oauth2/openid-connect](https://developers.google.com/identity/protocols/oauth2/openid-connect): "`hd`: The hosted G Suite domain of the user. Provided only if the user belongs to a hosted domain.".

1. Configure Tooljet for Google SSO with "External" user type. Ensure that environment variable `SSO_RESTRICTED_DOMAIN` is unset.
2. Attempt to log in to Tooljet with Google SSO using a non-GMail, non-Google Workspace account. (In my case, something akin to `example [at] brentsch.com`, which is not a Google Workspace account.)
3. Login will fail with "Invalid Email or Password" message.

This error generates no logs, which is an issue of its own.

### Any possible solutions?

When `SSO_RESTRICTED_DOMAIN` is not set, the `OauthService.#isValidDomain` check can be short-circuited. See our version of [server/ee/services/oauth/oauth.service.ts#L26-L29](https://github.com/scholarship-junkies/ToolJet/blob/d997aa93b9fa18f45fd1d4d38e26f0ba0b1014b5/server/ee/services/oauth/oauth.service.ts#L26-L29).

See link in previous section.

Yes.

brentschroeter:
fixes #2563

### DIFF
--- a/server/ee/services/oauth/oauth.service.ts
+++ b/server/ee/services/oauth/oauth.service.ts
@@ -25,11 +25,13 @@ export class OauthService {
   #isValidDomain(domain: string): boolean {
     const restrictedDomain = this.configService.get<string>('SSO_RESTRICTED_DOMAIN');
 
+    if (!restrictedDomain) {
+      return true;
+    }
     if (!domain) {
       return false;
     }
     if (
-      restrictedDomain &&
       !restrictedDomain
         .split(',')
         .filter((e) => !!e)


gsmithun4:
@brentschroeter can you please fix the lint errors?

Navaneeth-pk (approved):

Navaneeth-pk:
It seems the lint errors are from develop. Merging this PR.
@gsmithun4

PR #2580, opened 2022-03-19
[Bugfix] Fixes re-rendering of querypanel when canvas gets any update

### LINKED ISSUE #2579: Query Manager receives unnecessary renders resulting in wasted renders
Re-render only the components that are directly affected by that specific change in the editor. 
Possibly configuring query manager component to only render and differentiate when it is necessary.

Any update to the canvas, for example, adding a new button or resizing or repositioning a widget results in Query Manager to re-render the component even if the props/state does not receive any update.

1. Click on add icon or **create query** button in the query manager component
2. Select a datasource, e..g REST API
3. Add a new widget to the canvas (without creating or saving the query)
4. Create a new query and reposition or resize the widget

### Screenshots or Screencast

https://user-images.githubusercontent.com/67645175/159140209-3fa8ae3e-7e42-42fe-bd82-9a5df9d07c91.mov

arpitnath:
Resolves #2579

### DIFF
--- a/frontend/src/Editor/QueryManager/QueryManager.jsx
+++ b/frontend/src/Editor/QueryManager/QueryManager.jsx
@@ -57,7 +57,6 @@ let QueryManager = class QueryManager extends React.Component {
         currentState: props.currentState,
         selectedSource: source,
         dataSourceMeta,
-        isSourceSelected: props.isSourceSelected,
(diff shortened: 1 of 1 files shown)


gondar00 (approved):

arpitnath:
https://user-images.githubusercontent.com/67645175/159140355-c3670bf0-beed-4358-8751-bba16ea593ff.mp4

PR #2595, opened 2022-03-21
Improvement: Exposed dark mode as mode

### LINKED ISSUE #2594: Improvement: Add dark mode into globally exposed data

Currently, there is no way to access dark mode globally in the Tooljet application. So by exposing dark mode as mode, developers can use that in their design conditions and more use cases. 

Yes

shah21:
Resolves #2594

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -108,7 +108,7 @@ class Editor extends React.Component {
         components: {},
         globals: {
           currentUser: userVars,
-          mode: { value: props.darkMode ? 'dark' : 'light' },
+          mode: props.darkMode ? 'dark' : 'light',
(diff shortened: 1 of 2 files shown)


sherfin94 (changes requested):
One major issue and one minor issue,

Major issue: the value is not available on viewer (when we preview or launch the app)
Minor issue: please set the value on `globals.mode` instead of `globals.mode.value`

Please also remove the `console.log`s

shah21:
> 

For the second issue, currently, each global resolves as an object. So just setting the value directly will cause this error
Screenshot 2022-03-23 at 4 06 49 PM

sherfin94:
> > 
> 
> regarding the second issue, currently, each global resolves as an object in code hinter suggestions. So just setting the value directly will cause this error [Screenshot]
> 
> @sherfin94 So should I fix that in this issue? I mean the code hinter issue

In that case, change it to `globals.theme.name`

PR #2597, opened 2022-03-21
Feature/update timer behaviour

### LINKED ISSUE #2549: Timer widget : the start button should be disabled for "Count down" option with the default value is "00:00:00:000"
- If the timer type is set to "Count down" , the default value is or set to "00:00:00:000", the start button should be disabled as we can't start the timer anyway.
- Also there should be a different default value for the “count down” type f.e 00:00:10:000

With timer type is set to "Count down"  with the default value being "00:00:00:000". the start button is enabled. 

1. Add a timer widget from the widget manager
2. Select **Count down** from the timer type dropdown
3. Set the default value to "00:00:00:000" and click start

felipeTht:
Resolves #2549 

Update the json on component.js for **Timer**, I added a new prop under:

definition: {
   ...,
   defaults: [
        {
          type: _selectValue_,
          value: _defaultValue_
          paramName: _nameOfTheValue,_
        },
}.

The default values works for all select changes that has a defaults value under the component definition.

### DIFF
--- a/frontend/src/Editor/Components/Timer.jsx
+++ b/frontend/src/Editor/Components/Timer.jsx
@@ -18,9 +18,14 @@ export const Timer = function Timer({ height, properties = {}, styles, setExpose
   const [state, setState] = useState('initial');
   const [intervalId, setIntervalId] = useState(0);
 
+  const TimerType = {
+    COUNTDOWN: 'countDown',
+    COUNTUP: 'countUp',
+  };
+
   useEffect(() => {
     if (
-      properties.type === 'countDown' &&
+      properties.type === TimerType.COUNTDOWN &&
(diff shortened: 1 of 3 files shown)


arpitnath (changes requested):
@felipeTht Please revert the changes which are not related to this issue.
It will be great if you can follow the repo's eslint rules, it's easier to review 😅

felipeTht:
> 

I'm sorry it was setted the default auto saving with prettier conf. I pushed the file reverted.

gsmithun4 (commented):
Timer changes looks good, @arpitnath please take a look into components and inspector

arpitnath (approved):

PR #2619, opened 2022-03-23
Adding react-datepicker library instead of react-datetime library in datepicker widget

### LINKED ISSUE #2618: using react-datepicker

In the Datepicker component, we are using datetime library which is not comprehensive enough to meet the specifications of our date picker component. It has several bugs, and usage is difficult.

 We can replace the existing library with react-datepicker library

Yes, Willing to raise a PR

manishkushare:
resolve #2618

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -30,6 +30,7 @@
         "bootstrap": "^4.6.0",
         "classnames": "^2.3.1",
         "css-loader": "^6.5.1",
+        "date-fns": "^2.28.0",
         "dompurify": "^2.2.7",
         "draft-js": "^0.11.7",
         "draft-js-export-html": "^1.4.1",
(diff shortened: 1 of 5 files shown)


sherfin94 (changes requested):
Hey @manishkushare , there are a few things we need to fix,

1. The component's style doesn't change for dark mode. There is a parameter called `mode` that gets passed to the component which you can use to style the component accordingly for dark mode
2. The component's width and height should be equal to the width and height set by the user by clicking and dragging the blue container around it. This width and height is passed in as parameters to the React component that you're working on, you just have to apply it
3. The component seems to not respect border radius style, which is by default `0`. Check the style tab on the widget inspector to find out

sherfin94:
Border on the right side seems to go over the blue selector box: (should be something to do with margins or padding:
image

sherfin94:
If `enable date selection` is set to `false`, the widget goes to `Something went wrong`, ideally if its set to false, and time is set to true, the widget should remain as a time selector.

sherfin94:
It is possible to change months with one click, but there seems to be no way to jump years with this picker, please see if it can be enabled: 
image

sherfin94 (approved):

PR #2624, opened 2022-03-24
Chore: Update docker files to use buster to be used with oracledb

akshaysasidrn:
Fixes server crashing with missing dependencies on alpine image

### DIFF
--- a/plugins/packages/oracledb/package.json
+++ b/plugins/packages/oracledb/package.json
@@ -19,7 +19,7 @@
   "dependencies": {
     "@tooljet-plugins/common": "file:../common",
     "knex": "^0.95.14",
-    "oracledb": "^5.3.0",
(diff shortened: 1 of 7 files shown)


plugins/packages/oracledb/package.json
```diff
@@ -19,7 +19,7 @@
   "dependencies": {
     "@tooljet-plugins/common": "file:../common",
     "knex": "^0.95.14",
-    "oracledb": "^5.3.0",
```
Navaneeth-pk, line 22: This change may not be required.

akshaysasidrn: We can directly use the prebuilt binary so that for arm machines to run locally it can work out.
https://oracle.github.io/node-oracledb/INSTALL.html#-2-quick-start-node-oracledb-installation

PR #2636, opened 2022-03-25
[Feature] Implemented SSL certificate option in PostgreSQL data-source 

### LINKED ISSUE #2635: Feature: Add SSL Certificates option to PostgreSQL datasource

Normally, We can connect the PostgreSQL database with authority-issued SSL certificates or self-signed certificates. This will help us to establish a good secured connection between the database and tooljet. So Adding this feature to our application will be a good option

So we should add a field for certificate content and the user should be able to choose different types of certificate methods 

Yes

shah21:
Resolves #2635


gsmithun4 (approved):
LGTM

akshaysasidrn (approved):

PR #2667, opened 2022-03-29
Added a field for sending JSON body with Rest Api Queries 

### LINKED ISSUE #2666: Feature: Add a field for sending JSON body with Rest Api Queries

Yes. Currently, We don't have an option to send JSON strings or objects along with Rest API queries  

Add a field for that. And also need a toggle to show json body field instead of the body form  

Yes

shah21:
Resolves #2666


frontend/src/Editor/QueryManager/QueryEditors/Restapi/TabBody.jsx
```diff
@@ -9,21 +9,32 @@ export default ({
...
+  onJsonBodyChange,
   componentName,
+  bodyToggle,
+  setBodyToggle,
 }) => {
   return (
     <>
-      <GroupHeader addNewKeyValuePair={addNewKeyValuePair} paramType={'body'} descText={'Body Parameters'} />
-
+      <div style={{ display: 'flex' }}></div>
```
Navaneeth-pk, line 19: Can we use the `d-flex` class instead of inline CSS?

Navaneeth-pk:
If the body is a JSON array, it won't work.
Eg: `{{[1,2,3]}}`

Also, instead of trying to convert the JSON object to key value pairs, can we send the object as it is to the server? and based on `raw body` flag, the server can just use the object or proceed with existing flow?

@shah21

Nandinisaha13:
<html>
<body>

Test Case | Steps | Result
-- | -- | --
Should be able to provide JSON body with Rest API queries. | Click on Datasource & select Rest API. | Selected Rest API
  | Enter the URL(pipedream) and click save. | Entered URL and saved it.
  | Toggle the json option in body, enter the body in json format. | The body should be visible in json format.
  | Click on preview. | Send the queries to pipedream.
  | View the ouput on piepedream, | It matches the input given above.
Should be able to authenticate with OAuth 2.0 | Click on Datasource & select Rest API. | Selected Rest API
  | Enter the details as per documentation & click on save. | The details should be entered & saved.
  | Create a query for get method & enter header click on preview. | It asks the account to get selected.
  | Select the account & close the window. | It shows the account detail in json.


</body>
</html>

akshaysasidrn (approved):
Nice! LGTM.

PR #2694, opened 2022-03-31
Feature: Add data transfer object layer to validate and sanitize user inputs

akshaysasidrn:
- Adds data transfer object to sanitize and validate API params
- Resolves for 
   - #2749 
   - #2750 
- Update existing specs for DTO changes


server/src/helpers/utils.helper.ts
```diff
@@ -29,3 +30,11 @@ export async function getCachedConnection(dataSourceId, dataSourceUpdatedAt): Pr
     }
   }
 }
+
+export function sanitizeInput(value: string) {
+  return sanitizeHtml(value, {
+    allowedTags: [],
```
gondar00, line 36: This would not allow any tags to be entered, might conflict with the html widget, probably add a default list here?

gondar00 (changes requested):

gsmithun4 (approved):
LGTM

server/src/helpers/utils.helper.ts
akshaysasidrn: What do you suggest? There would be HTML as input?

gondar00 (dismissed):

server/src/helpers/utils.helper.ts
gondar00: Yeah, like you said since appDef is not being affected we can ignore it for now,

PR #2722, opened 2022-04-03
fix: add transition to search input

### LINKED ISSUE #2717: Transition animation for app search field on the dashboard
image
When this field is clicked, the width is increased. It would be better if we add `ease-in-out` transition for width change. 

Reference: https://stackoverflow.com/questions/7861648/animate-an-elements-width-from-0-to-100-with-it-and-its-wrapper-being-only-a

Olaleye-Blessing:
resolve #2717

### DIFF
--- a/frontend/src/_ui/Search/Search.jsx
+++ b/frontend/src/_ui/Search/Search.jsx
@@ -93,9 +93,8 @@ export const SearchBox = ({ onChange, ...restProps }) => {
           type="text"
           value={searchText}
           onChange={handleChange}
-          className="form-control"
+          className="form-control animate-width-change"
           placeholder={placeholder ?? 'Search'}
-          style={{transition: "width 0.2s ease-in-out"}}
         />
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


frontend/src/_ui/Search/Search.jsx
```diff
@@ -95,6 +95,7 @@ export const SearchBox = ({ onChange, ...restProps }) => {
           onChange={handleChange}
           className="form-control"
           placeholder={placeholder ?? 'Search'}
+          style={{transition: "width 0.2s ease-in-out"}}
```
Navaneeth-pk, line 98: Can we use a class instead of inline CSS? If we can abstract these styles into a class like `animate-width-change`, we can  reuse it for other input fields in future if we want to.

Olaleye-Blessing:
@Navaneeth-pk I tried creating a class but I couldn't find the base css file.

Olaleye-Blessing:
@Navaneeth-pk I've seen the [base styles folder](https://github.com/ToolJet/ToolJet/tree/develop/frontend/src/_styles). Should create an util file and add the class to it?

Navaneeth-pk (approved):

PR #2731, opened 2022-04-04
added instruction text property  in file picker widget to add custom instructions

### LINKED ISSUE #2730: Add "instruction text" prop in Date picker widget to display custom instructions for the widget
### Environment
Development

The user should be able to provide the instruction text for the file picker widget

We have default text that is shown on dragging the file picker widget onto the editor something like "drag the file..."

1. Drag the file Picker widget in the editor
2. You will see the default message - Drag the file .....

### Any possible solutions?
Add instruction text property in the file picker widget and that should be displayed on the file picker widget

manishkushare:
Resolve #2730

### DIFF
--- a/frontend/src/Editor/Components/FilePicker.jsx
+++ b/frontend/src/Editor/Components/FilePicker.jsx
@@ -14,6 +14,7 @@ export const FilePicker = ({
   styles,
 }) => {
   //* properties definitions
+  const instructionText = component.definition.properties.instructionText.value ?? '';
   const enableDropzone = component.definition.properties.enableDropzone.value ?? true;
   const enablePicker = component.definition.properties?.enablePicker?.value ?? true;
   const maxFileCount = component.definition.properties.maxFileCount?.value ?? 2;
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/Components/FilePicker.jsx
```diff
@@ -14,6 +14,7 @@ export const FilePicker = ({
   styles,
 }) => {
   //* properties definitions
+  const instructionText = component.definition.properties.instructionText.value ?? '';
```
sherfin94, line 17: Hey @manishkushare , file picker components in existing apps don't have the instructionText property, so it is `undefined` for them, this is causing crashes for existing apps. Could you make this change and make sure that it is not crashing existing file picker widgets?

```suggestion
  const instructionText = component.definition.properties.instructionText?.value ?? '';
```

sherfin94 (changes requested):
One small changed requested. Otherwise looks fine!

sherfin94:
@manishkushare We'll need to write a migration that will set the instructionText on all existing file pickers to `Drag and Drop some files here, or click to select files`. Otherwise it does not show up on the widget inspector of existing file pickers, it confuses users and if they accidentally click on the instruction text field on the inspector, it goes to blank.

To generate new migration, you can use this command from the `server` folder:

`npm run typeorm -- migration:create -n YourMigrationName`

@arpitnath, please provide Manish with a simple migration example.

sherfin94:
Hey @manishkushare , why do we need to create/update `plugins/packages/woocommerce/lib/operations.json` for filepicker?

sherfin94 (changes requested):
PR looks fine, also tested the migration. Finally please resolve the conflict by merging develop to this branch.

PR #2322, opened 2022-02-22
[Improvement] migrates to react-select: selection dropdown component

arpitnath:
Resolves #2321 

- **Migrates to `react-select` lib from `react-select-search`**
- **Default common styles for the selection dropdown** to avoid duplications
- **Updates selection dropdown component** :
      - widget inspector [widget definations]
      - dynamic form - queries
      - dynamic form - datasource modal
      - datasource: oauth and authentication dropdowns (source component)
      - query pane - query selector dropdown
      - query pane - rest API methods selector

PR #2745, opened 2022-04-04
added react-select library to make dropdown while creating version

### LINKED ISSUE #2744: Add react-select library to create dropdown in create version modal
### Environment
Development

While creating the version inside the application, we may have a proper react dropdown using react-select in the modal

We have browser dropdown, which does not look consistent with the application

1. Create a new application or go inside the existing application
2. On the header, click App version
3. Dropdown will appear, which will show versions and the option to create a new version
4. Click on create a version
5. A modal will open to create a version
6. Click on Create Version from the input field, and you will see a dropdown, which is a browser dropdown

### Any possible solutions?
We may use a react-select library to create a dropdown

manishkushare:
Resolved  #2744

### DIFF
--- a/frontend/src/Editor/AppVersionsManager.jsx
+++ b/frontend/src/Editor/AppVersionsManager.jsx
@@ -3,6 +3,7 @@ import Modal from '../HomePage/Modal';
 import { toast } from 'react-hot-toast';
 import { appVersionService } from '@/_services';
 import { Confirm } from './Viewer/Confirm';
+import Select from 'react-select';
 
 export const AppVersionsManager = function AppVersionsManager({
   appId,
@@ -240,6 +241,11 @@ const CreateVersionModal = function CreateVersionModal({
       createVersion(versionName, createAppVersionFrom);
     }
   };
+
+  const options = appVersions.map((version) => {
+    return { ...version, label: version.name };
+  });
+
   return (
     <Modal
       show={showModal || showCreateVersionModalPrompt}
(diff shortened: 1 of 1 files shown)


frontend/src/Editor/AppVersionsManager.jsx
```diff
@@ -264,16 +270,15 @@ const CreateVersionModal = function CreateVersionModal({
...
-                {version.name}
-              </option>
-            ))}
-          </select>
+          <Select
+            options={options}
+            maxMenuHeight={100}
+            menuPlacement="auto"
+            onChange={(version) => setCreateAppVersionFrom(version)}
+          />
```
gondar00, line 281: How about adding a defaultValue prop here?
```suggestion
          <Select
            options={options}
            maxMenuHeight={100}
            menuPlacement="auto"
            defaultValue={options[options.length - 1]}
            onChange={(version) => setCreateAppVersionFrom(version)}
          />
```

manishkushare:
Sure, on it

frontend/src/Editor/AppVersionsManager.jsx
sherfin94: Yes, presently, even though the default selected version is `v1`, it will create version from the editing version instead of `v1`, unless we drop it down and click on it. @manishkushare

sherfin94 (changes requested):
Please also apply brand colors, like the dropdown widget: 
image

Currently the colors are these: 
image

arpitnath:
@manishkushare Putting this PR on hold.
Till #2322 gets resolved (merged)

sherfin94:
> @manishkushare Putting this PR on hold. Till #2322 gets resolved (merged)

Hey @arpitnath , given #2322 is merged, what should we do about this PR?

Issue #2775, opened 2022-04-08
Making canvas background colour programmable


stepinfwd:
### Is your proposal related to a problem?

Currently there is no option for us to give different colours based on theme , something like 
{{globals.theme.name === 'dark' : 'cyan' : 'blue'}}

### Describe the solution you'd like
Add codehinter feature same in editor for changing background colour of canvas

### If the feature is approved, would you be willing to submit a PR?
Yes

PR #2807, opened 2022-04-13
Bugfix :: Canvas background made programmable and fix for alpha not working

### LINKED ISSUE #2776: Alpha value in colour picker  for changing canvas doesn't work
User must be able to adjust alpha value

Alpha value is always set to 100

1.Click on setting icon in left sidebar
2.Open Background color of canvas
3.Try changing alpha value

Yes

stepinfwd:
#resolves #2776 #2775


sherfin94 (changes requested):
Few major issues:

- Whatever code that we type using Fx, gets applied immediately and does not get applied when the app is running. For eg, we should be able to set the value of the color from an input widget, and if the user edits the text on the widget while the app is running, background color should change.
- When I press the popout button for the Fx field, the area that I can type on the portal that appears is small, there is something buggy thats going on. Not fully sure if its related to this PR

https://user-images.githubusercontent.com/20333972/163809169-433b1c25-1ec1-40af-abe7-4830f76c1745.mov

stepinfwd:
Link : https://canvas-toool.herokuapp.com/
@sherfin94

frontend/src/Editor/LeftSidebar/SidebarGlobalSettings.jsx
```diff
@@ -23,6 +29,19 @@ export const LeftSidebarGlobalSettings = ({
...
+    setRealState(currentState);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [currentState.components]);
+
+  React.useEffect(() => {
+    console.log('check', realState);
+    backgroundFxQuery &&
+      globalSettingsChanged('canvasBackgroundColor', resolveReferences(backgroundFxQuery, realState));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [realState]);
```
sherfin94, line 43: How about this to make the background color change as soon as the value computed by backgroundFxQuery changes?
```suggestion
  }, [JSON.stringify(resolveReferences(backgroundFxQuery, realState))]);
```

sherfin94 (changes requested):

sherfin94:
Its not working on viewer @stepinfwd, please check.

stepinfwd:
issue fixed @sherfin94

PR #2826, opened 2022-04-14
[Bugfix] On dispatching Undo ⌘Z or Redo ⌘+shift+Z actions to a component on the editor should also update the internal state or props for the children.

### LINKED ISSUE #2825: On dispatching **Undo** `⌘Z` or **Redo** `⌘+shift+Z` actions to a component on the editor should also update the internal state or props for the children.
### Description
On dispatching **Undo** `⌘Z` or **Redo** `⌘+shift+Z` actions to a component on the editor should also update the internal state or props for the children.
Similarly renaming a widget, should also start a new render pass. 

Perform **Undo** `⌘Z` or **Redo** `⌘+shift+Z` actions or rename a widget should update the internal state **currentState** (which is passed down as props to the children)

No updates are received to children until the next render, on performing the above actions in the editor.

Case 1: Performing **Undo-ing** or **Redo-ing** action
1. Create a new app from the Dashboard 
2. Click on the **Inspector** from the left sidebar to open the popup. (expand the **component** node )
3. Add a new widget to the canvas from the widget manager. (The widget is added to the inspector popup: in the **component** node )
4. Delete the widget from the editor. The widget should also be removed from the inspector popup
5. Perform an **Undo** action by doing `⌘` + `Z`. The widget (recently deleted) should be back in the editor.
6. Check the inspector popup, the recently deleted widget should be added back to the component node

Case 2: Renaming a widget from the widget inspector
1. Create a new app from the Dashboard 
2. Click on the **Inspector** from the left sidebar to open the popup. (expand the **component** node )
3. Add a new widget to the canvas from the widget manager. For example: a Button widget. Default name is **button1** (The widget is added to the inspector popup: in the **component** node )
```json
component: {
 ...,
 button1: {...}
}
```
4. Click on the **config handler** of the widget to open the widget inspector.
5. Rename the widget to `renamedBtn`. And Press enter. The widget gets the update.
6. Check the inspector popup, the component node should also receive the update and the widget should be renamed to **renamedBtn** from **button1**.

### Screenshots or Screencast

https://user-images.githubusercontent.com/67645175/163368216-cad73017-9aa5-4476-8dca-b6dba17d7b81.mov

arpitnath:
Resolves #2825

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -506,6 +506,7 @@ class Editor extends React.Component {
       this.handleAddPatch
     );
     setStateAsync(_self, newDefinition).then(() => {
+      computeComponentState(_self, _self.state.appDefinition.components);
       this.autoSave();
     });
   };
@@ -882,9 +883,6 @@ class Editor extends React.Component {
     } = this.state;
 
     const appVersionPreviewLink = editingVersion ? `/applications/${app.id}/versions/${editingVersion.id}` : '';
-
-    console.log('Render App:: EDITOR ==> currentState', JSON.stringify(this.state.currentState));
-    console.log('Render App:: EDITOR ==> Full State', JSON.stringify(this.state));
 
     return (
       <div className="editor wrapper">
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -144,6 +144,12 @@ class Editor extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.appDefinition !== this.state.appDefinition) {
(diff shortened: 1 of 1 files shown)


Navaneeth-pk (approved):

frontend/src/Editor/Editor.jsx
```diff
@@ -144,6 +144,12 @@ class Editor extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.appDefinition !== this.state.appDefinition) {
```
gondar00, line 148: @arpitnath this doesn't do a deep equality check, 

```
var a = {x: { a: 2 }}
var b = {x: { a: 2 }}

a === b // returns false
```

Can we use lodash's `isEqual` here?

PR #2844, opened 2022-04-18
Fix key-value input increment when no items are added

### LINKED ISSUE #2738: Rest api headers keep incrementing when no items are added to either key or value
Header fields should not increment when no value is entered

Header fields are incrementing when no value is entered

Same issue observed with params and body tabs too
### Screenshots or Screencast

https://user-images.githubusercontent.com/12490590/161534761-100c91d2-cdfa-4614-b80f-c5eab1c7bb44.mov

eyuelberga:
Resolves #2738


gsmithun4:
deployed app: https://tj-2738.herokuapp.com/

gondar00 (approved):

Issue #2840, opened 2022-04-18
When pagination icons are disabled the cursor should change to disabled on hover


gondar00:
### Version Information
Latest

### Environment
Tooljet Cloud

### What is the expected behaviour?
When hovered on pagination icon the cursor should change to disabled

### What is the current behaviour?
The cursor doesn't turn to disabled on hover

### How to reproduce the issue?

1. Add a table
2. Hover over the disabled icon
3. Observe the cursor 

### Screenshots or Screencast
<img width="1440" alt="Screenshot 2022-04-18 at 12 50 49 PM" src="https://user-images.githubusercontent.com/12490590/163769751-b9142996-40e6-41f0-8c3a-99eed5fb5fb6.png">

### Any possible solutions?
Please add css classname `cursor-not-allowed` [here](https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/Editor/Components/Table/Pagination.jsx#L54)

### If the bug is confirmed, would you be willing to submit a PR?

Olaleye-Blessing:
Can I work on this?

I would like to know what the two circled buttons do? Can they also be affected by `cursor-not-allowed` class?

![Screenshot (354)_LI](https://user-images.githubusercontent.com/70102539/163812432-17cb16d8-d4ce-40be-a48d-dd6a666d11a4.jpg)

gondar00:
Sure @Olaleye-Blessing assigning the issue to you, thanks for contributing. 

The two icons take you either to the first page or the last page, ideally they should also be disabled if its the first page/last page, if possible feel free to disable them and add the `cursor-not-allowed` class

PR #2852, opened 2022-04-18
bugfix :: disable cursor when pagination icons are disabled

Olaleye-Blessing:
Resolves: #2840 

## Quick question:
Is it okay to put those common button className in a variable? All buttons have `btn btn-light btn-sm`?
It could be:

> let btnClassName = "btn btn-light btn-sm";
> let prevBtnClassName = `${btnClassName} ${pageIndex === 1 ? "cursor-not-allowed" : ""}`;
> let nextBtnClassName = `${btnClassName} ${!autoCanNextPage && !serverSide ? "cursor-not-allowed" : ""}`;

Then other classNames could be added, which is `mx-2` at the moment.


Navaneeth-pk:
Please fix the lint errors. 
@Olaleye-Blessing

PR #2862, opened 2022-04-19
added HTML widget

### LINKED ISSUE #2861: Feature request - HTML viewer widget
Feature Request : 
Current scenario - There is no widget, which will show an HTML page/ window upon providing raw HTML data

What we can do to add the above functionality - 
We can add a feature where users can provide raw HTML and can see the output in the application itself using the HTML viewer widget

manishkushare:
Resolve #2861


sherfin94:
@manishkushare You seem to have pushed package-lock.json without having installed any new packages. Please remove these changes.

image

frontend/src/Editor/Components/components.js
```diff
@@ -2021,4 +2021,43 @@ export const componentTypes = [
...
+      value: {},
+    },
+    definition: {
+      others: {
+        showOnDesktop: { value: '{{true}}' },
+        showOnMobile: { value: '{{false}}' },
+      },
+      properties: {
+        rawHtml: {
+          value: '',
```
sherfin94, line 2054: Lets set some simple sample HTML as default value here.

sherfin94:
Is it possible to also add another property called `CSS` , where we can supply css that will be applied to the HTML that is rendered?

sherfin94 (changes requested):
I have requested some changes. Additionally, please also add documentation and resolve conflicts with `develop` branch.

Widget works well by the way 😄

manishkushare:
Sure, on it

PR #2903, opened 2022-04-25
added react-select library to create app version

### LINKED ISSUE #2744: Add react-select library to create dropdown in create version modal

manishkushare:
Resolve #2744

### DIFF
--- a/frontend/src/Editor/AppVersionsManager.jsx
+++ b/frontend/src/Editor/AppVersionsManager.jsx
@@ -3,7 +3,8 @@ import Modal from '../HomePage/Modal';
 import { toast } from 'react-hot-toast';
 import { appVersionService } from '@/_services';
 import { Confirm } from './Viewer/Confirm';
-
+import Select from '../_ui/Select';
+import defaultStyle from '../_ui/Select/styles';
 export const AppVersionsManager = function AppVersionsManager({
   appId,
   editingVersion,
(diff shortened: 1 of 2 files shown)


manishkushare:
@sherfin94 and @gondar00, I am facing an issue with the custom style.
Rest everything working fine, I have also added custom styles, but while the component is getting injected in the dom, a class that is responsible for dark theme display is injected in both dark and light theme, that's the reason, we get the dark background of the dropdown option every time.

Kindly watch the video for reference, In the beginning, I have added custom styles to the Select component, and it provides dark theme background color of options in the dropdown, when I remove the custom style, it is working fine.

https://www.loom.com/share/6bfb6d7fce7d4abd9305e37c2d92be7f


Below is the image, when no custom Style is applied so following class gets injected into the light theme in the dom
Screenshot 2022-04-26 at 12 54 29 AM


Below is the image, when the custom style is applied and the class, which has a background color of a dark theme, is injected into dom, even in the light theme.

Screenshot 2022-04-26 at 12 43 31 AM

frontend/src/Editor/AppVersionsManager.jsx
```diff
@@ -241,6 +242,37 @@ const CreateVersionModal = function CreateVersionModal({
       createVersion(versionName, createAppVersionFrom);
     }
   };
+  const options = appVersions.map((version) => {
+    return { ...version, label: version.name, value: version.name };
+  });
+  const width = '100%';
+  const height = 32;
+  const darkMode = localStorage.getItem('darkMode');
```
arpitnath, line 250: @manishkushare Here the `darkMode` is a string ('true').

arpitnath (changes requested):
`const darkMode = localStorage.getItem('darkMode') === 'true';`

frontend/src/Editor/AppVersionsManager.jsx
manishkushare: @arpitnath applied the suggested change, it is working fine now, thank you.

akshaysasidrn (approved):

akshaysasidrn (approved):

PR #2919, opened 2022-04-27
Added GH issue template for cypress tests

ajith-k-v:
Fixes: #2918


sherfin94 (changes requested):
I think it would be better to not mention `Cypress` on the template, instead just mention `Test`, and ask a question to specify the kind of test (the default answer to this could be `Cypress integration test`) cc: @Navaneeth-pk

ajith-k-v:
@sherfin94 Updated the changes. Please review

Navaneeth-pk (approved):

PR #2920, opened 2022-04-27
feat: add user avatar

gondar00:
fixes: #3018 

Deployed on: https://useravatar.herokuapp.com


frontend/src/SettingsPage/SettingsPage.jsx
```diff
@@ -101,16 +116,35 @@ function SettingsPage(props) {
...
+                      <input type="text" className="form-control" name="email" value={email} readOnly disabled />
+                    </div>
+                  </div>
+                  <div className="col">
+                    <div className="mb-3">
+                      <div className="form-label">Avatar</div>
+                      <input
+                        onChange={(e) => {
+                          const file = e.target.files[0];
+                          console.log(file);
```
akshaysasidrn, line 133: missed to remove

server/src/controllers/files.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+} from '@nestjs/common';
+import { Readable } from 'stream';
+import { Response } from 'express';
+import { FilesService } from '../services/files.service';
+import { UpdateFileDto } from '../dto/update-file.dto';
+import { JwtAuthGuard } from 'src/modules/auth/jwt-auth.guard';
+
+@Controller('files')
+@UseInterceptors(ClassSerializerInterceptor)
+export class FilesController {
```
akshaysasidrn, line 21: If we aren't using `#findAll` and `#update`, could we remove them?

akshaysasidrn: Also, let's try to stick with current naming conventions ie #index, #show etc when possible within a controller?

server/src/controllers/users.controller.ts
```diff
@@ -37,6 +48,14 @@ export class UsersController {
     };
   }
 
+  @Post('avatar')
+  @UseGuards(JwtAuthGuard)
+  @UseInterceptors(FileInterceptor('file'))
+  async addAvatar(@Req() req, @UploadedFile() file: Express.Multer.File) {
+    if (!file || !file?.buffer) throw new BadRequestException('File cannot be empty');
```
akshaysasidrn, line 55: could use dto directly for this to validate here and then pass it onto the service

akshaysasidrn (changes requested):

server/src/controllers/files.controller.ts
gondar00: Cool, I have removed the unused endpoints and renamed the fn

akshaysasidrn (approved):

akshaysasidrn (approved):

PR #2942, opened 2022-05-01
Fix error in create version modal

### LINKED ISSUE #2941: Cannot create new app versions
### Version Information

v1.11.0

### Environment

Docker with image `tooljet/tooljet-ce:v1.11.0`

The "Create version" modal should create a new version from an existing version.

- An error "Oops, something went wrong" is shown when clicking the Create Version button.
- This happens the second time a new version is created

It looks like the issue is that frontend call to the "versions" API (`api/apps/[APP_ID]/versions`) is missing the `versionFromId` parameter. This does not happen the first time a new version is created -- only the second time

1. Create a new app
2. Create a new version "v2" from "v1" --> this is successful
3. Create a new version "v3" from "v2" --> this fails

### Screenshots or Screencast

image
image

### Any possible solutions?

When the `versionFromId` is included in the request, the version seems to be created correctly. I assume this value does not get correctly set somewhere in the modal setup.

I think the API calling code is [here](https://github.com/ToolJet/ToolJet/blob/16ce6f1f4201b7ebcbac999449c28ec200222ff0/frontend/src/Editor/AppVersionsManager.jsx#L70-L98)

Yes: https://github.com/ToolJet/ToolJet/pull/2942

dlebech:
This commit fixes a bug that occurs in the app version modal popup when
an existing version is chosen from the dropdown menu.

Fixes #2941

### DIFF
--- a/frontend/src/Editor/AppVersionsManager.jsx
+++ b/frontend/src/Editor/AppVersionsManager.jsx
@@ -243,7 +243,7 @@ const CreateVersionModal = function CreateVersionModal({
     }
   };
   const options = appVersions.map((version) => {
-    return { ...version, label: version.name, value: version.name };
+    return { ...version, label: version.name, value: version };
   });
   const width = '100%';
   const height = 32;


frontend/src/Editor/AppVersionsManager.jsx
```diff
@@ -304,6 +304,11 @@ const CreateVersionModal = function CreateVersionModal({
             options={options}
             defaultValue={options[options.length - 1]}
             onChange={(version) => {
+              // If the changed value is a simple string, find the version
```
akshaysasidrn, line 307: Hey @dlebech, we can fix this by changing the value of the dropdown selection to be the version object rather than the version name value which is causing this bug.

https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/Editor/AppVersionsManager.jsx#L246

In that `return` we can change it as 
``` return { ...version, label: version.name, value: version };``` 
Could you please make that change?

dlebech: Thanks @akshaysasidrn , beautiful solution. I've updated the PR.

dlebech: I have rebased the PR as well.

akshaysasidrn:
Thanks for your contribution 🙇🏾‍♂️

akshaysasidrn (approved):

PR #2944, opened 2022-05-02
[Feature] Added an option to save edited queries to avoid losing values by state changes  🚀 

### LINKED ISSUE #2943: Feature to save edited queries to avoid losing values by state changes

Currently, if a user will lose edited & unsaved query values while,
1. Switching to another query
2. Click on the new query icon
3. Click inside the canvas (Like select a widget)

So, to avoid these problems, we can show a popup modal to save the unsaved queries

Auto-save 

Yes

shah21:
Resolves #2943

Heroku https://query-save.herokuapp.com/


frontend/src/Editor/QueryManager/QueryEditors/Restapi/index.jsx
```diff
@@ -17,6 +17,7 @@ class Restapi extends React.Component {
     );
     this.state = {
       options,
+      arrayValuesChanged: false,
```
akshaysasidrn, line 20: Is this state being used directly? Looks like we are only setting it within options to be used with.

frontend/src/Editor/QueryManager/QueryManager.jsx
```diff
@@ -109,9 +117,27 @@ let QueryManager = class QueryManager extends React.Component {
...
+      } else if (!isQueryChanged && this.state.selectedQuery.kind === 'restapi' && this.state.restArrayValuesChanged) {
+        this.setState({ showSaveConfirmation: true, nextProps });
+        return;
+      }
+    }
     this.setStateFromProps(nextProps);
   }
 
+  removeRestKey = (options) => {
+    delete options.arrayValuesChanged;
```
akshaysasidrn, line 137: This would modify the object and delete the key in the original object.
Better to copy and remove to compare?

sherfin94:
Looks like the issue where the query gets discarded when the panel is dragged is back

https://user-images.githubusercontent.com/20333972/166435477-4b3988fd-7298-44ac-ae49-9088acbd176a.mov

sherfin94 (changes requested):
One change requested

sherfin94:
@shah21  Right now, if I click on `+` and start creating a new query, and add a widget to the canvas while doing this, then the confirmation dialogue is shown. I understand that the panel already resets in this scenario, even without this confirmation box. So is that something we can fix in this PR itself, or is it a separate issue that should be fixed separately?

shah21:
> @shah21 Right now, if I click on `+` and start creating a new query, and add a widget to the canvas while doing this, then the confirmation dialogue is shown. I understand that the panel already resets in this scenario, even without this confirmation box. So is that something we can fix in this PR itself, or is it a separate issue that should be fixed separately?

Yeah, it is the expected behavior of the query manager, if the user does something in the canvas area while editing a query, the popup will be shown.

sherfin94 (approved):

PR #2948, opened 2022-05-02
[Feature] Added textSize style to the text widget

### LINKED ISSUE #2947: [Feature] Add a style property to change text size of a text widget

Currently, there is no option to change the text size of a text widget 

Add a style property to change the text size of a text widget

Yes

shah21:
Resolves #2947


sherfin94 (changes requested):
Hey @shah21 , lets also write a migration to set the textSize of every existing text widget so that it wont say `0` like this: 
image

sherfin94 (approved):

PR #2950, opened 2022-05-02
exposed variable of timer is set to default value after reset

### LINKED ISSUE #2949: On Resetting the Timer widget, exposed variables are not updated
### Environment
Develop

In the Timer Widget, When we reset the timer, the exposed value of the timer in the inspector should get reset

The exposed value is not getting reset, in the inspector section

manishkushare:
Resolve #2949


gsmithun4 (approved):
LGTM

PR #2961, opened 2022-05-03
fixed addition of application to folder more than once

### LINKED ISSUE #2916: Able to add the same application in a folder multiple times.
### Version Information
all versions

Should not be able to add the same application in a folder multiple times.

Able to add the same application in a folder multiple times

1. Add an app to a folder
2. Add it again
3. Observe the app count of folders are increasing 

### Screenshots or Screencast
[Video](https://www.loom.com/share/cb94c69231ac4bbaa889bed93d19b6f0)

### Any possible solutions?
Back-end validation if app exists in folder

No

asprazz:
Fixes #2916 

Please review and let me know if any changes are required.

## Tests 
- [x] Manually Tested on local
- [x] Test Case written

## Screenshots
test

## Approach
I have used repository to find if record already exists by querieng before creating folderApp record, please let me know 
if any factory method available which can be used to find if app id is already referenced with folder id which can be used.

### DIFF
--- a/server/src/services/folder_apps.service.ts
+++ b/server/src/services/folder_apps.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@nestjs/common';
+import { BadRequestException, Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
 import { FolderApp } from '../../src/entities/folder_app.entity';
@@ -11,6 +11,14 @@ export class FolderAppsService {
   ) {}
 
   async create(folderId: string, appId: string): Promise<FolderApp> {
+    const existingFolderApp = await this.folderAppsRepository.findOne({
+      where: { appId, folderId },
+    });
+
+    if (existingFolderApp != null && existingFolderApp !== undefined) {
+      throw new BadRequestException('App has been already added to the folder');
+    }
+
     const newFolderApp = this.folderAppsRepository.create({
       folderId,
       appId,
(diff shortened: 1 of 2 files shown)


Navaneeth-pk:
Please fix the lint errors.
@asprazz

asprazz:
@Navaneeth-pk Please find it done.

server/src/services/folder_apps.service.ts
```diff
@@ -11,6 +11,14 @@ export class FolderAppsService {
   ) {}
 
   async create(folderId: string, appId: string): Promise<FolderApp> {
+    const oldFolderApp = await this.folderAppsRepository.findOne({
```
gsmithun4, line 14: shall we rename the variable something like `existingFolderApp`? `old` will be confusing.

server/src/services/folder_apps.service.ts
```diff
@@ -11,6 +11,14 @@ export class FolderAppsService {
...
+      where: { appId, folderId },
+    });
+
+    if (oldFolderApp != null) {
```
gsmithun4, line 18: `if (!oldFolderApp)`
will be good, we can capture undefined value also

gsmithun4 (changes requested):

server/src/services/folder_apps.service.ts
asprazz: Yes! it should be more informative

server/src/services/folder_apps.service.ts
asprazz: I have added the check. Please check.

gsmithun4 (approved):
LGTM

gsmithun4:
deployed https://folder-bug.herokuapp.com/

PR #2977, opened 2022-05-05
Feature :: Button group widget

### LINKED ISSUE #2876: Feature :: support for button group widget
Currently tooljet doesn't have a button group widget.

Add button group widget.

Yes.

stepinfwd:
closes #2876


sherfin94 (changes requested):
Widget works fine!, few things:

Since you're using an array as exposed variable for both single selection and multiple selection, please ensure that the initially exposed value is `[1]` and not `1`.

Also please add documentation, sync with @withshubh if that helps.

stepinfwd:
updated with required changes @sherfin94

sherfin94:
Can we make it so, that we can supply default selected an actual array instead of a string that will be parsed as an array?
image

sherfin94:
When the `Enable multiple selection` property is toggled on or off, the widget's state (the buttons that are displayed as selected, and the exposed value) should reset to the `default selected` value. Presently, the widget stays as if no buttons are selected and the exposed value remains the same.

frontend/src/Editor/Components/components.js
```diff
@@ -2122,6 +2182,71 @@ ReactDOM.render(<ConnectedComponent />, document.body);`,
...
+    exposedVariables: {
+      selected: [1],
+    },
+    definition: {
+      others: {
+        showOnDesktop: { value: '{{true}}' },
+        showOnMobile: { value: '{{false}}' },
+      },
+      properties: {
+        label: { value: `Button` },
```
sherfin94, line 2232:
```suggestion
        label: { value: `Button group` },
```

sherfin94 (changes requested):
Suggested few changes, otherwise looks fine.

sherfin94 (changes requested):
Final change, there seems to be unnecessary padding on the top when button group label is removed:

image

If header is removed, this padding is not required.

sherfin94 (approved):

PR #2999, opened 2022-05-06
Avoid creation of system group names

### LINKED ISSUE #2969: Able to create group name with system group names
### Version Information
all versions

Should not allow to create group name with `Admin` and `All Users` since they are reserved for system groups

Able to create 

1. Try to create group with name `Admin`
2. Group will be created

### Screenshots or Screencast
Screenshot 2022-05-04 at 4 52 37 PM

### Any possible solutions?
Restrict both group names in back end 

No

tanmoyopenroot:
#### Checklist
- [x] Includes tests
- [ ] Update documentation

#### Changes proposed in this pull request:
Check for possible values like `'All Users', 'all_users', 'Admin', 'admin'`

#### Reviewers should focus on:
Currently hard cording all the possible values in the create method in `GroupPermissionsService`
```javascript
    const reservedGroups = ['All Users', 'all_users', 'Admin', 'admin'];
```

### DIFF
--- a/server/src/services/group_permissions.service.ts
+++ b/server/src/services/group_permissions.service.ts
@@ -27,11 +27,17 @@ export class GroupPermissionsService {
     private appRepository: Repository<App>,
 
     private usersService: UsersService
-  ) {}
+  ) { }
 
   async create(user: User, group: string): Promise<GroupPermission> {
     if (!group || group === '') {
       throw new BadRequestException('Cannot create group without name');
+    }
+
+    const reservedGroups = ['All Users', 'all_users', 'Admin', 'admin'];
+
+    if (reservedGroups.includes(group)) {
+      throw new BadRequestException('Cannot create group with system reserved group name');
     }
 
     const groupToFind = await this.groupPermissionsRepository.findOne({
--- a/server/src/services/group_permissions.service.ts
+++ b/server/src/services/group_permissions.service.ts
@@ -27,17 +27,17 @@ export class GroupPermissionsService {
     private appRepository: Repository<App>,
 
     private usersService: UsersService
-  ) { }
+  ) {}
 
   async create(user: User, group: string): Promise<GroupPermission> {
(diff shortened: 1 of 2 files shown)


server/src/services/group_permissions.service.ts
```diff
@@ -27,13 +27,19 @@ export class GroupPermissionsService {
...
     private usersService: UsersService
-  ) {}
+  ) { }
 
   async create(user: User, group: string): Promise<GroupPermission> {
     if (!group || group === '') {
       throw new BadRequestException('Cannot create group without name');
     }
 
+    const reservedGroups = ['All Users', 'all_users', 'Admin', 'admin'];
```
gsmithun4, line 37: `['All Users', 'Admin']`
are enough

server/src/services/group_permissions.service.ts
```diff
@@ -27,13 +27,19 @@ export class GroupPermissionsService {
...
+
+    if (reservedGroups.includes(group)) {
+      throw new BadRequestException('Cannot create group with system reserved group name');
```
gsmithun4, line 40: shall we throw the same error
```
Group name already exist
```

gsmithun4 (changes requested):

gsmithun4:
@tanmoyopenroot Please fix the lint errors

server/src/services/group_permissions.service.ts
tanmoyopenroot: Sure, Will do

tanmoyopenroot: Done

server/src/services/group_permissions.service.ts
tanmoyopenroot: Done

gsmithun4:
deployed 
https://group-fixes-tj.herokuapp.com/

PR #3007, opened 2022-05-08
Feature :: Image widget enhancements 

### LINKED ISSUE #2994: [ Enhancement ] ::  image widget enhancements

1) Zoom in zoom out 
2) Fit to content , fit to boundary , stretch options addition
3) Loading state 

Add requested features.

Yes

stepinfwd:
resolves #2994


frontend/package.json
```diff
@@ -78,6 +78,7 @@
     "react-table": "^7.6.3",
     "react-table-plugins": "^1.3.1",
     "react-tooltip": "^4.2.18",
+    "react-zoom-pan-pinch": "^2.1.3",
```
Navaneeth-pk, line 81: Changes to `frontend/package-lock.json` file also needs to be committed.

sherfin94: Not sure if this is related to this, but `npm install` is failing for me:
image
@stepinfwd

sherfin94:
Seems to work well!

But I couldn't understand what `Object fit` is? What are the other options for it? Can we use a dropdown for it instead of code hinter?

image

Please see if we an apply a more unambiguous name for it, and please also document it.

sherfin94 (changes requested):
Requested some changes.

Plus one other major issue: Try creating an image widget and set some custom image from google images on develop branch. Then switch to this branch, image's `fit` will change and it will appear in a different manner. We need to display the existing images exactly as they are.

Also, the buttons are outside the widget right now, a better way would be to display them in the left bottom or right bottom, smaller buttons, transparent, and only make them fully visible when mouse is hovered over the image.

stepinfwd:
> Seems to work well!
> 
> But I couldn't understand what `Object fit` is? What are the other options for it? Can we use a dropdown for it instead of code hinter?
> 
> image
> 
> Please see if we an apply a more unambiguous name for it, and please also document it.

We will change it to drop down object fit has : cover , contain , auto etc as values same as css objectFit

stepinfwd:
Updated documentation @sherfin94

sherfin94 (approved):

Issue #3023, opened 2022-05-10
Add configuration to select required plugins


gsmithun4:
### Is your proposal related to a problem?
Currently we are installing all plugins while building, if the respective instance does not required some plugins, there no features to exclude them

### Describe the solution you'd like
Add configuration for it

### If the feature is approved, would you be willing to submit a PR?
Yes

PR #3024, opened 2022-05-10
Added configuration to select plugins

gsmithun4:
Resolves https://github.com/ToolJet/ToolJet/issues/3023


plugins/build-packages.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+if (!scope && excludes && excludes.length > 0) {
+  ignore = excludes.map((e) => `--ignore @tooljet-plugins/${e}`).join(' ');
+}
+
+const buildCommand = `lerna ${
+  scope
+    ? `--scope @tooljet-plugins/common --scope @tooljet-plugins/restapi ${scope} `
+    : ignore
```
akshaysasidrn, line 17: Does it make sense to only have includes by default since we can't use both? 🤔 
That'd mean we only maintain a whitelist by default and all the packages will be listed in _configs.json_.
If the user only wants specific packages, can simply comment others out?

gondar00 (approved):

docker/client.Dockerfile
```diff
@@ -7,18 +7,18 @@ RUN npm i -g npm@7.20.0
...
 # Fix for heap limit allocation issue
 ENV NODE_OPTIONS="--max-old-space-size=4096"
 
+ENV NODE_ENV=production
+
 # Build plugins
-COPY ./plugins/package.json ./plugins/package-lock.json ./plugins/
-RUN npm --prefix plugins install
-COPY ./plugins/ ./plugins/
 RUN npm run build:plugins
```
gondar00, line 19: Is this line required?

package.json
```diff
@@ -37,8 +37,8 @@
     "ts-node": "^10.1.0"
   },
   "scripts": {
-    "prebuild:plugins": "npm --prefix plugins install",
-    "build:plugins": "npm --prefix plugins run build",
+    "build:plugins": "NODE_ENV=production && node build-plugins.js",
+    "build:plugins:dev": "node build-plugins.js",
```
gondar00, line 41: Would it be possible to merge these two into one?

docker/client.Dockerfile
gsmithun4: https://github.com/ToolJet/ToolJet/pull/3024/files#diff-81041e84b391055b6814e5bca2e5c0883f825e4dd4011f987de3d2354ccbec21R11
We are copying entire plugins directory, no need to pick package and package-lock separately

package.json
gsmithun4: yes, can remove `NODE_ENV=production` and it will work based on the `NODE_ENV` set
```suggestion
    "build:plugins": "node build-plugins.js",
```

akshaysasidrn (approved):
The changes look good. 

So as of now we just skip oracledb on development. But let's say there are multiple packages that a user wants to skip on development, then one would have to go about and modify each plugin's package.json which would then show up for git changes. How should those changes be handled for? 🤔 

Maybe as an improvement for later should have some single config file wherein a dev can specify which plugins to install and which not to. That file should be autogenerated and should be git ignored as well. Wdyt?

gsmithun4:
> The changes look good.
> 
> So as of now we just skip oracledb on development. But let's say there are multiple packages that a user wants to skip on development, then one would have to go about and modify each plugin's package.json which would then show up for git changes. How should those changes be handled for? 🤔
> 
> Maybe as an improvement for later should have some single config file wherein a dev can specify which plugins to install and which not to. That file should be autogenerated and should be git ignored as well. Wdyt?

@gondar00 is working on it, what we are trying to configure is, oracle and sap hana needs extra configs to work on development, we are disabling them in development build.

PR #3046, opened 2022-05-11
[Improvement] Filepicker widget : The file picker should reject files with existing file names.

### LINKED ISSUE #3045: [Improvement] File picker widget
## Description

- Disable file selection on max count with `Pick multiple files` set to true
- check already added/selected file count before adding new files to the widget for multiple selections

arpitnath:
Resolves #3045


frontend/src/Editor/Components/FilePicker.jsx
```diff
@@ -105,6 +107,19 @@ export const FilePicker = ({
   const [accepted, setAccepted] = React.useState(false);
   const [showSelectdFiles, setShowSelectedFiles] = React.useState(false);
   const [selectedFiles, setSelectedFiles] = React.useState([]);
+  //* custom validator
+  function validateFileExists(file) {
+    const isValid = selectedFiles.some((selectedFile) => selectedFile.name === file.name);
```
sherfin94, line 112: @arpitnath Shouldn't this be based on the same full file path instead of same file name? Because different files can have same name, right?

sherfin94 (changes requested):

frontend/src/Editor/Components/FilePicker.jsx
arpitnath: Yeah, makes sense. I think we need to have the selected files' **paths** and compare them instead of their names.

sherfin94 (approved):

sherfin94 (approved):

PR #3059, opened 2022-05-16
added default value in the toggle widget

### LINKED ISSUE #3058: Enhancement : add default value to toggle widget
### Environment
Development

The default value should be provided in the properties of the Toggle Widget

The default value is not provided in the properties of the Toggle Widget

manishkushare:
Resolve #3058

### DIFF
--- a/frontend/src/Editor/Components/components.js
+++ b/frontend/src/Editor/Components/components.js
@@ -618,6 +618,7 @@ export const componentTypes = [
     },
     properties: {
       label: { type: 'code', displayName: 'Label' },
+      defaultValue: { type: 'toggle', displayName: 'Default value' },
     },
     events: {
       onChange: { displayName: 'On change' },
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/Components/components.js
```diff
@@ -618,6 +618,7 @@ export const componentTypes = [
     },
     properties: {
       label: { type: 'code', displayName: 'Label' },
+      defaultValue: { type: 'toggle', displayName: 'Default value' },
```
sherfin94, line 621: Lets call it default status instead of default value

```suggestion
      defaultValue: { type: 'toggle', displayName: 'Default status' },
```

sherfin94 (changes requested):
The latest changes with `resolveWidgetFieldValue` is not required. It seems to be already working for existing widgets as you have used the `?? false` operation on `properties.defaultValue`.

Unless you observed the app crashing without it, please revert that commit.

sherfin94 (approved):

PR #3063, opened 2022-05-16
added default prop in checkbox widget

### LINKED ISSUE #3062: Enhancement : Checkbox Widget --> add default value property
### Environment
Development
The checkbox widget should have a default value prop, to be false

There is no default value prop available

manishkushare:
Resolve #3062


sherfin94:
When we set default status to `true`, it is filling the check box instead of actually checking it: 
image

sherfin94 (changes requested):
Please fix the issue where checkbox is getting filled instead of checked

PR #3076, opened 2022-05-18
[Feature]: Inline style support for custom component

### LINKED ISSUE #3075: [Feature]: Support Inline Styles in Custom Component
Need support for inline styles in custom component

kavinvenkatachalam:
Resolves #3075


sherfin94 (changes requested):
We can access the unresolved value of `code` property from CustomComponent.jsx in this manner:

```const code = props.component.definition.properties.code.value;```

This way we wouldn't have to hack properties resolution.

kavinvenkatachalam:
> We can access the unresolved value of `code` property from CustomComponent.jsx in this manner:
> 
> `const code = props.component.definition.properties.code.value;`
> 
> This way we wouldn't have to hack properties resolution.

The `code` value can be accessed using the `component` prop. But while resolving the `code` value, I'm getting an error `Unexpected token ':'`. So, I skipped the resolvable when the property is `code` in `CustomComponent`. Any suggestions?

image

sherfin94:
> > We can access the unresolved value of `code` property from CustomComponent.jsx in this manner:
> > `const code = props.component.definition.properties.code.value;`
> > This way we wouldn't have to hack properties resolution.
> 
> The `code` value can be accessed using the `component` prop. But while resolving the `code` value, I'm getting an error `Unexpected token ':'`. So, I skipped the resolvable when the property is `code` in `CustomComponent`. Any suggestions?
> 
> image

Hey @kavinvenkatachalam, I'm not sure I understand. we don't need to resolve the value of `code`, that is the reason why we are accessing it from `component.definition` instead of `props`, is it not?. 

If you're facing `undefined` related errors, might be a good idea to use `component.definition?.properties?.code?.value ?? ''`

sherfin94 (approved):

Issue #2851, opened 2022-04-18
Limit "Update available" banner to admins


brentschroeter:
### Is your proposal related to a problem?

Currently when an instance detects an available Tooljet update, even if `process.env.CHECK_FOR_UPDATES === "0"` at runtime, it displays a banner to visitors reading, "Update available A new version of ToolJet has been released," with buttons to view release notes or skip the update. This banner appears to all authenticated users, including non-admins. (As an aside, though we have only observed this behavior with authenticated visitors, we have not confirmed whether it is different with unauthenticated visitors to publicly accessible applications.)

This leads to confusion with users, as well as potential breaches of dev-ops protocol if unprivileged users are able to initiate or skip an update.

<img width="407" alt="Screen Shot 2022-04-18 at 4 25 46 PM" src="https://user-images.githubusercontent.com/1895860/163872508-4c77ee23-904e-4263-8d4c-a7615e8e370c.png">

### Describe the solution you'd like

It would be preferable if the banner appeared only to authenticated administrators, and/or disabled entirely when `CHECK_FOR_UPDATES` is set to "0" at runtime.

### Describe alternatives you've considered

N/A, though suggestions are appreciated.

### If the feature is approved, would you be willing to submit a PR?

Yes

Navaneeth-pk:
@brentschroeter makes sense. Assigning the issue to you as mentioned in the PR. Thanks for contributing 🚀

brentschroeter:
@Navaneeth-pk Looking at the code, the second part of the solution (disable banner when `process.env.CHECK_FOR_UPDATES == '0'`) is trivial and actually a bug fix rather than a feature issue. The first part of the solution is a little more technically involved, as it involves messing with user Abilities, and entails some deeper philosophical questions about who should be allowed to skip versions: if there are multiple organizations then perhaps it isn't desirable to use organization admin membership as the guard for managing instance-wide update tasks.

My suggestion is to scope only the `process.env.CHECK_FOR_UPDATES` bug fix for this ticket. My naive guess is that sysadmins who would care deeply about the nuances of the rest are likely to run with `CHECK_FOR_UPDATES` disabled anyways for the sake of repeatable deployments. This is a one-liner PR I can submit anytime.

withshubh:
Hey @brentschroeter 
Any update on it?

PR #3084, opened 2022-05-18
[Bugfix] CHECK_FOR_UPDATES=0 is not respected (#2851)

brentschroeter:


server/src/controllers/metadata.controller.ts
```diff
@@ -52,7 +52,11 @@ export class MetadataController {
     const onboarded = data['onboarded'];
 
     if (process.env.NODE_ENV == 'production') {
-      if (process.env.CHECK_FOR_UPDATES) {
+      if (
```
akshaysasidrn, line 55: Could you update this as per the docs?
checking `process.env.CHECK_FOR_UPDATES === '1'` should do.

akshaysasidrn: I see, the default behaviour is to be enabled as per docs.

PR #3096, opened 2022-05-20
[Feature]: Added universal props for widgets with Tooltip

### LINKED ISSUE #3095: [Feature]: Universal Props for Widgets

If I wish to add a property to all the widgets, I need to go and add the property to all the widgets instead of maintaining it in a single place.

We can create a file with universal props which get added to all the widgets during run time.

Yes

kavinvenkatachalam:
Resolves #3095 

Added universal props for widgets & added tooltip property to all the widgets


frontend/src/Editor/component-properties-resolution.js
```diff
@@ -27,3 +27,19 @@ export const resolveStyles = (component, currentState, defaultValue, customResol
     return {};
   }
 };
+
+export const resolveGeneral = (component, currentState, defaultValue, customResolvables) => {
+  if (currentState) {
+    const general = component.definition.general;
```
sherfin94, line 33: Existing widgets seems to be crashing without this.
```suggestion
    const general = component.definition?.general ?? {};
```

sherfin94:
This is a bit of a nitpick, but I feel let's use the variable name `generalProperties` instead of `general` everywhere applicable, reason being that we'll also need a `generalStyles` section, making `general` ambiguous. I'm guessing this would also make  variables like `resolveGeneral` => `resolveGeneralProperties`, `resolvedGeneral` => `resolvedGeneralProperties` etc. 

Let's keep the label of the section `General` itself.

sherfin94 (changes requested):
Suggested few changes. Lastly:

There are a few lines in docs (`create-widget.md`) that mentions `components.js`, looks like we need to change that to `widgetConfig.js`.

sherfin94 (approved):

PR #3115, opened 2022-05-25
[Feature]: Trigger actions from RunJS

### LINKED ISSUE #3114: [Feature]: Trigger Actions from RunJS

As of now, it is not possible to run the query, setting and unsettling variable using RunJS in Tooljet. 

We can create a function that can run the query, set/unset the variable. For eg,

To run query - `actions.runQuery(queryName)` or `queries.queryName.run()`
To set variable - `actions.setVariable(variableName, variableValue)`
To unset variable - `actions.unSetVariable(variableName)`

Yes

kavinvenkatachalam:
Resolves #3114

To run query - `actions.runQuery(queryName)` or `queries.queryName.run()`
To set variable - `actions.setVariable(variableName, variableValue)`
To unset variable - `actions.unSetVariable(variableName)`


kavinvenkatachalam:
> Few improvement suggestions
> 
> * Can we make `runQuery` function return a promise so that it can be `await`ed from the RunJS query? This way, RunJS query would only complete after any queries that are `await`ed are also complete.
> * Also, we should be able to get the `result` and `status` of query's execution from that promise. `result` should contain the data returned by the query and `status` should be an object of the format: `{ succeeded : true, code: 201}` (Do suggest if you feel any different here).
> * Instead of `query` , can we call the object that contains actions as `actions`. `actions.runQuery(queryName)`, `actions.setVariable(variableName, value)` etc.
> * This is not necessary at the moment, but it will be nice if queries can also be run as `queries.queryName.run()` (returning a promise with the same functionality as mentioned above)

@sherfin94 As per suggestion returned the promise. Once the query is completed, then the object is produced in the below format (followed the format in our code base)

Success - `{status: ok, data: data}` 
Failed - `{status: failed, message: message}`

Let me know your suggestion if you've any different idea

sherfin94 (changes requested):
Please check these

1. Awaited data is not available on preview
2. If runQuery is not awaited, then the preview shows the result of the target query
    instead of the RunJS query in hand
3. If runQuery is awaited, then the target query's data variable is not populated
4. The resulting data, when accessed via inspector, the inspector crashes

@arpitnath Please help @kavinvenkatachalam with 4

kavinvenkatachalam:
> Please check these
> 
> 1. Awaited data is not available on preview
> 2. If runQuery is not awaited, then the preview shows the result of the target query
>    instead of the RunJS query in hand
> 3. If runQuery is awaited, then the target query's data variable is not populated
> 4. The resulting data, when accessed via inspector, the inspector crashes
> 
> @arpitnath Please help @kavinvenkatachalam with 4

@sherfin94 Fixed the first 3 issues

sherfin94 (commented):
Really sorry for the long delay, as we discussed on call, please consider if we can reuse the executeAction function in appUtils.js so that we can avoid redundancy.

sherfin94 (approved):

Issue #3127, opened 2022-05-26
[Bug] Authorization header with key visible for Rest API plugin


gsmithun4:
### What is the expected behaviour?
Authorization header with key should not be available in front-end

### Screenshots or Screencast
![tooljet-restapi-source-auth-header](https://user-images.githubusercontent.com/3417097/170417869-3b901c76-7b99-4af6-b6a6-b4b25100e2f4.jpg)

### If the bug is confirmed, would you be willing to submit a PR?
Yes

PR #3129, opened 2022-05-26
[BUG] Hide authorization header for Rest API

gsmithun4:
Resolves https://github.com/ToolJet/ToolJet/issues/3127


sherfin94:
@gsmithun4 For bearer authentication, the authorization is properly hidden. But if I select basic authentication, I am able to obtain username and password from devtools. Similarly for oauth, I am able to get client id and another property (redirect url I think). Is this in the scope of this PR?

PR #3131, opened 2022-05-26
[Feature]: Run query while Creating/Saving 

### LINKED ISSUE #3130: [Feature]: Save & Run button in Query Manager

While creating/updating the query, the user needs to save the query and run the query. This involves two steps and it is not intuitive.

We can replace the `save` button with `save & run` which runs the query after saving it

Yes

kavinvenkatachalam:
Resolves #3130


sherfin94 (changes requested):
Hey @kavinvenkatachalam , `Create and run` works correctly, but `Save and run` doesn't seem to run the query.

https://user-images.githubusercontent.com/20333972/170957549-498a717b-1b16-479c-a8cd-0d6063e1268c.mov

kavinvenkatachalam:
> Hey @kavinvenkatachalam , `Create and run` works correctly, but `Save and run` doesn't seem to run the query.
> 
>  Screen.Recording.2022-05-30.at.2.32.54.PM.mov

I'll look into it

kavinvenkatachalam:
@sherfin94 Query is running fine. There is some issue with the `Table.jsx` in updating the new value. From the below screenshot,`tableData` has the updated value. But due to memo and none of the dependencies got changed, the variable `data` holds the old value which is why the table is not updated. 

image

sherfin94:
> @sherfin94 Query is running fine. There is some issue with the `Table.jsx` in updating the new value. From the below screenshot,`tableData` has the updated value. But due to memo and none of the dependencies got changed, the variable `data` holds the old value which is why the table is not updated.
> 
> image

@kavinvenkatachalam When we run query by clicking on the `play` button or by means of some `Run query` action, the `data` of the query becomes `[]` while it is loading, and then it becomes the new data. This makes it easy for components to reflect the value. When when we run query while saving/creating, this does not seem to happen. Can we reuse the same way in which query is run when we click on the play button, here?. We might just have to reuse the corresponding function.

kavinvenkatachalam:
> > @sherfin94 Query is running fine. There is some issue with the `Table.jsx` in updating the new value. From the below screenshot,`tableData` has the updated value. But due to memo and none of the dependencies got changed, the variable `data` holds the old value which is why the table is not updated.
> > image
> 
> @kavinvenkatachalam When we run query by clicking on the `play` button or by means of some `Run query` action, the `data` of the query becomes `[]` while it is loading, and then it becomes the new data. This makes it easy for components to reflect the value. When we run query while saving/creating, this does not seem to happen. Can we reuse the same way in which query is run when we click on the play button, here?. We might just have to reuse the corresponding function.

Got it. I'll check and reuse that functionality

sherfin94 (changes requested):
When I click on `Save and run` without making any changes to the query, it seems to keep loading forever:
image

sherfin94 (changes requested):
The query `create and run` button seems to crash the app when clicked. (save and run is working)

https://user-images.githubusercontent.com/20333972/173767930-6ae9d34a-b477-4b13-bae9-2cd985b7bf07.mov

sherfin94 (approved):

PR #3152, opened 2022-05-30
Cypress test for the manage groups page

ajith-k-v:
Fixes: #3151 

1. Added cypress test for the manage groups page.
2. Added dropdown selector.
3. Added manage groups path.
4. updated common utils file.
5. Added data-cy attributes for the elements

Screenshot 2022-05-30 at 10 46 14 AM


ajith-k-v:
creating PR against develop

PR #3161, opened 2022-05-30
added default table type prop to table-bordererd in component.js

### LINKED ISSUE #3160: Bug : Table Widget --> tableType prop is not taking default value
### Environment
Developed

In Table Widget, the table type prop should display the default value that is 'bordered'

It is not displaying default value in table type prop

manishkushare:
resolve #3160

### DIFF
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -86,9 +86,7 @@ export function Table({
 
   if (typeof clientSidePagination !== 'boolean') clientSidePagination = true;
 
-  const tableTypeProperty = component.definition.styles.tableType;
-  let tableType = tableTypeProperty ? tableTypeProperty.value : 'table-bordered';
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -86,9 +86,7 @@ export function Table({
 
   if (typeof clientSidePagination !== 'boolean') clientSidePagination = true;
 
-  const tableTypeProperty = component.definition.styles.tableType;
-  let tableType = tableTypeProperty ? tableTypeProperty.value : 'table-bordered';
-  tableType = tableType === '' ? 'table-bordered' : tableType;
```
Navaneeth-pk, line 91: Removal of logic from here will have effects on backward compatibility.

manishkushare: Sure, shifting back to previous logic

Navaneeth-pk (approved):

PR #3175, opened 2022-05-31
 Display ToolJet's version in the dashboard

### LINKED ISSUE #3174: Feature Request : display ToolJet's current version in the user profile in the dashboard
We can display ToolJet's current version in the dashboard, in the user's profile.

manishkushare:
Resolve #3174


sherfin94 (changes requested):
@manishkushare , since we have to involve changes in docker file, it will affect every kind of deployments, this will be a cumbersome change.

Therefore, could you try to use this alternate approach instead:

1. ToolJet version is available in the backend, in the `globalThis` object that is available at several places
2. We are already sending config information such as google maps key from backend to the frontend using an API
3. Try to modify the API such that this version information is also send along with it, and simply use that on the front-end.

@arpitnath Could you guide Manish in doing this?

docker/client.Dockerfile.dev
```diff
@@ -7,7 +7,7 @@ RUN npm i -g npm@7.20.0
 
 # set working directory
 WORKDIR /app
-
+COPY ./.version ./.version
```
akshaysasidrn, line 10: You will have to update the other dockerfile setups as well. 
Will it be simpler to infer from `/metadata` call to the server which is being done [here](https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/App/App.jsx#L45)?

PR #3190, opened 2022-06-02
Workspace invite and user onboarding flow changes

gsmithun4:


frontend/src/ConfirmationPage/ConfirmationPage.jsx
```diff
@@ -183,7 +184,7 @@ class ConfirmationPage extends React.Component {
                   onClick={this.setPassword}
                   disabled={isLoading}
                 >
-                  Finish account setup
+                  Finish account setup{this.props.location.state.inviteToken ? ' and accept invite' : ''}
```
akshaysasidrn, line 187: Since both mails are sent as an invite, think it's not necessary to add - _**accept invite**_ explicitly. Can keep as - **_Finish account setup_.**

server/src/services/email.service.ts
```diff
@@ -100,7 +116,7 @@ export class EmailService {
           <p>Hi ${name || ''},</p>
           <br>
           <span>
-          ${sender} has invited you to use ToolJet workspace ${organisationName}. Use the link below to set up your account and get started.
+          ${sender} has invited you to use ToolJet workspace ${organizationName}. Use the link below to set up your account and get started.
```
akshaysasidrn, line 119: can we add a `:` after workspace?

frontend/src/App/App.jsx
```diff
@@ -129,13 +129,14 @@ class App extends React.Component {
             <Route path="/forgot-password" component={ForgotPassword} />
             <Route path="/reset-password" component={ResetPassword} />
             <Route
-              path="/invitations/:token"
+              path="/invitations/:token/:inviteToken?"
```
akshaysasidrn, line 132: Can we have two separate paths?
`invitations/:token` & `invitations/:token/workspaces/:organizationToken
It would be easy to reason about then.

frontend/src/ConfirmationPage/ConfirmationPage.jsx
```diff
@@ -23,7 +23,7 @@ class ConfirmationPage extends React.Component {
 
   setPassword = (e) => {
     e.preventDefault();
-    const token = this.props.location.state.token;
+    const { token, inviteToken } = this.props.location.state;
```
akshaysasidrn, line 26: let's use `organizationToken` as per backend?

akshaysasidrn (changes requested):

akshaysasidrn (approved):

PR #3200, opened 2022-06-03
[ Feature ]:: steps widget

### LINKED ISSUE #3199: [ Feature ] :: Steps widget

Currently tooljet doesn't have a steps widget implemented.

Implement steps widget.

Yes.

stepinfwd:
closes #3199


sherfin94 (changes requested):
Widget looks good, few improvements:

1. The space above the widget seems stubborn, I'm not able to get rid of it minimizing the widget. Can we avoid that unnecessary space at the top? (And also at the bottom if possible)
image

2. When 'Steps selectable' is turned off, the step names still show underlines, and the cursor turns into pointer, making them look like hyperlinks, lets just show them as plain text in this case

3. Can we allow colors using hex code? This will be important for some clients.

stepinfwd:
updated with required changes @sherfin94

sherfin94 (approved):

PR #3214, opened 2022-06-06
[Bugfix] Fixed Signup email case sensitive issue

### LINKED ISSUE #3213: Signup email should not be a case sensitive field, causing duplicate accounts with same email and different letter cases
### Version Information
Latest
### Environment

All

Users should not be able to create multiple accounts with the same email and different letter cases 

Email is a case sensitive field now

1. create an account with an email
2. use the previous email to create a new account, but use a different letter cases
3. Observe

Yes

shah21:
Fixes #3213


gsmithun4:
@shah21 Its happening sometime
https://www.loom.com/share/b4abcee851ed4e2285194294aeb2525c

server/src/services/organizations.service.ts
```diff
@@ -318,7 +318,7 @@ export class OrganizationsService {
     const userParams = <User>{
       firstName: inviteNewUserDto.first_name,
       lastName: inviteNewUserDto.last_name,
-      email: inviteNewUserDto.email,
+      email: inviteNewUserDto.email ? inviteNewUserDto.email.toLowerCase() : '',
```
gsmithun4, line 321:
```suggestion
      email: inviteNewUserDto.email?.toLowerCase(),
```

server/src/services/auth.service.ts
```diff
@@ -185,6 +185,7 @@ export class AuthService {
   }
 
   async signup(email: string) {
+    email = email ? email.toLowerCase() : '';
```
gsmithun4, line 188:
```suggestion
    email = email?.toLowerCase();
```

gsmithun4 (changes requested):

gsmithun4:
@shah21 we have to make changes here
https://github.com/ToolJet/ToolJet/blob/develop/server/src/controllers/organization_users.controller.ts#L24
`InviteNewUserDto` you can add transformations

server/src/services/organizations.service.ts
gsmithun4: please check if we can do it in dto transformations

server/src/services/auth.service.ts
gsmithun4: please check if we can do it in dto transformations

shah21: We can't do this through transformation for AppAuthenticationDto. cause that same dto is using with login function too. so that will cause `user not found or unauthorized` errors if an existed user with email that has Uppercase letters

shah21: I can create a new different dto for create controller function

server/src/services/organizations.service.ts
shah21: It is Possible

akshaysasidrn (changes requested):

server/migrations/1654596810662-ConvertAllUserEmailsToLowercaseAndDeleteDuplicateUsers.ts
```diff
@@ -0,0 +1,198 @@
...
+    );
+    await queryRunner.dropUniqueConstraint('users', columnUniques[0].name);
+
+    const usersQuery = entityManager
+      .getRepository(User)
+      .createQueryBuilder('users')
+      .leftJoinAndSelect('users.organizationUsers', 'organizationUsers')
+      .leftJoinAndSelect('users.apps', 'apps')
+      .leftJoinAndSelect('users.groupPermissions', 'groupPermissions')
+      .leftJoinAndSelect('users.userGroupPermissions', 'userGroupPermissions')
```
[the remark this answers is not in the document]
akshaysasidrn: Please ignore, migrating user and group permissions would do.

akshaysasidrn (approved):

gsmithun4 (dismissed):
LGTM

gsmithun4 (approved):
LGTM

PR #3218, opened 2022-06-07
Added url-encoded body support to restapi plugin

### LINKED ISSUE #3149: Query(RESTAPI): Post request cannot be successful, but in postman can
### Environment

    ToolJet Cloud, Chrome

Same REST API can be successfully requested as when debugging in postman.
HTTP methods: post
image

API has no request success, According to the error message prompt: the lack of the required field "usercode", but the actual phenomenon has been filled in, there seems to be no parameters in body passed to the API.

1. Add Queries, Select "REST API"
2. Fill in the content as the following screenshot
3. Click “PreView”

### Screenshots or Screencast
image

image

```
{
    "requestObject": {
        "requestUrl": "Address/user/login.do",
        "requestHeaders": {
            "user-agent": "got (https://github.com/sindresorhus/got)",
            "content-type": "application/x-www-form-urlencoded",
            "content-length": "103",
            "accept-encoding": "gzip, deflate, br"
        },
        "requestParams": {}
    },
    "responseObject": {
        "statusCode": 400,
        "responseBody": "{\"timestamp\":1653673802825,\"status\":400,\"error\":\"Bad Request\",\"message\":\"Required String parameter 'userCode' is not present\",\"path\":\"/user/login.do\"}"
    },
    "responseHeaders": {
        "access-control-allow-credentials": "true",
        "access-control-allow-methods": "POST, GET, OPTIONS, DELETE",
        "access-control-allow-headers": "Content-Type,Accept,token,headers",
        "content-type": "application/json;charset=UTF-8",
        "transfer-encoding": "chunked",
        "content-encoding": "gzip",
        "vary": "Accept-Encoding",
        "date": "Fri, 27 May 2022 17:50:02 GMT",
        "connection": "close"
    }
}
```

### Any possible solutions?
When I put a field from the body in params, I can request success, but in fact I hope these parameters are put in body.

Sorry, I just used it shortly, I couldn't distinguish it.

shah21:
Fixes #3149

### DIFF
--- a/plugins/packages/restapi/lib/index.ts
+++ b/plugins/packages/restapi/lib/index.ts
@@ -93,6 +93,7 @@ export default class RestapiQueryService implements QueryService {
 
     const headers = this.headers(sourceOptions, queryOptions, hasDataSource);
     const customQueryParams = sanitizeCustomParams(sourceOptions['custom_query_params']);
+    const isUrlEncoded = this.checkIfContentTypeIsURLenc(queryOptions['headers']);
 
     /* Chceck if OAuth tokens exists for the source if query requires OAuth */
     if (requiresOauth) {
@@ -142,7 +143,8 @@ export default class RestapiQueryService implements QueryService {
         ...paramsFromUrl,
         ...this.searchParams(sourceOptions, queryOptions, hasDataSource),
       },
-      json,
+      form: isUrlEncoded ? json : undefined,
+      json: !isUrlEncoded ? json : undefined,
     };
 
     if (authType === 'basic') {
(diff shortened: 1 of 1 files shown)


plugins/packages/restapi/lib/index.ts
```diff
@@ -142,7 +143,8 @@ export default class RestapiQueryService implements QueryService {
         ...paramsFromUrl,
         ...this.searchParams(sourceOptions, queryOptions, hasDataSource),
       },
-      json,
+      form: isUrlEncoded ? json : undefined,
```
akshaysasidrn, line 146: let's conditionally add the key as well, something like

```
...(isUrlEncoded ? {form: body} : {json: body})
```

akshaysasidrn (approved):

PR #3241, opened 2022-06-09
Cypress test for the manage groups page

ajith-k-v:
Fixes: #3151 

1. Added cypress test for the manage groups page.
2. Added dropdown selector.
3. Added manage groups path.
4. Updated common utils file.
5. Added data-cy attributes for the elements.

Screenshot 2022-06-09 at 4 12 38 PM


akshaysasidrn:
@ajith-k-v We need to add couple of test scenarios for managing groups.
1. Default groups (All users, Admin) cant be deleted.
2. Admin group cannot have permissions changed (ie. checkbox is disabled)
3. Permissions for each resource is honored. (ex: user with app create permission disabled, wouldn't be able to create apps)
5. New groups can be created and deleted. 

No need to add it as a part of this PR. We can create issues for each and then pick it up?

ajith-k-v:
> @ajith-k-v We need to add couple of test scenarios for managing groups.
> 
> 1. Default groups (All users, Admin) cant be deleted.
> 2. Admin group cannot have permissions changed (ie. checkbox is disabled)
> 3. Permissions for each resource is honored. (ex: user with app create permission disabled, wouldn't be able to create apps)
> 4. New groups can be created and deleted.
> 
> No need to add it as a part of this PR. We can create issues for each and then pick it up?

@akshaysasidrn 

1. Delete option is not available for default groups(All users, Admin).
2. Admin group cannot have permissions changed is verified on the spec.
3. New groups can be created and deleted operation is verified on the spec.
4. Permissions for each resource are honored. (ex: user with app create permission disabled, wouldn't be able to create apps)  For this scenario will add a new PR.

akshaysasidrn:
>     1. Delete option is not available for default groups(All users, Admin).
Just checking delete option is present for newly added ones but not present on default groups should be enough. 👍🏾

PR #3272, opened 2022-06-13
User account level security issue

gsmithun4:


server/src/services/organization_users.service.ts
```diff
@@ -49,57 +51,41 @@ export class OrganizationUsersService {
...
-    )?.organization;
+    if (this.configService.get<string>('DISABLE_MULTI_WORKSPACE') === 'true') {
+      // Resetting password if single organization
+      await this.usersService.updateUser(id, { password: uuid.v4() });
+    }
 
     await this.emailService.sendOrganizationUserWelcomeEmail(
-      updatedUser.email,
-      updatedUser.firstName,
+      organizationUser?.user?.email,
```
akshaysasidrn, line 81: We'd always want to expect email, first name etc here.
Wouldn't it make sense to return early incase `organizationUser` is not found rather than using optional chaining?

server/src/services/users.service.ts
```diff
@@ -147,24 +147,21 @@ export class UsersService {
...
       lastName,
       password: hashedPassword,
     };
 
     // removing keys with undefined values
-    cleanObject(updateableParams);
+    cleanObject(updatableParams);
 
-    let user: User;
+    await this.updateUser(userId, updatableParams);
```
akshaysasidrn, line 160: why remove it outside db transaction block? 🤔

gsmithun4: to reuse the `updateUser`  function

akshaysasidrn: If we are doing multiple updates, better to group it in a transaction to avoid partial updates being done. Can modify the function to recieve the entity manager inorder to do so.

akshaysasidrn (approved):

JamieSlome:
@akshaysasidrn - does this issue relate to the following report:
https://www.huntr.dev/bounties/86881f9e-ca48-49b5-9782-3c406316930c/

PR #3274, opened 2022-06-13
Feature : Color Picker Widget

### LINKED ISSUE #3273: Feature Request : Color Picker Widget
Feature Request --> Adding Color Picker Widget

Use case - The user can select the desired color and accordingly, the user will get access to color in various formats like hexadecimal, RGB, and RGBA in the inspector

manishkushare:
Resolve #3273


sherfin94 (commented):
The default color property should be a color property instead of a code property (user can use fx if needed)

image

sherfin94:
The default color property doesn't update the color until the page is reloaded, this means that the default color cannot be set programatically right now:

image

sherfin94:
The color picker's design should be similar to the one in the widget inspector: 
image

Please also make sure to style the widget accordingly in light mode and darkmode.

sherfin94 (changes requested):
Requested few changes.

sherfin94 (changes requested):
The color selector seems to have a bad habit of vanishing if we don't manoeuvre the cursor surgically. Please check.

https://user-images.githubusercontent.com/20333972/174071513-9c8aeaab-cac8-45b6-bbc0-0510f7d4eecd.mov

manishkushare:
@sherfin94 I have Fixed the issue, kindly check it.

sherfin94 (changes requested):
Could you make the default color property a color field itself instead of text field?

image

manishkushare:
@sherfin94  Changed the default color type to color in the properties and made sure, exposed variables are changing accordingly

sherfin94 (changes requested):
@manishkushare Component seems to crash when value supplied is not a valid color, like empty spaces or `undefined` etc.


https://user-images.githubusercontent.com/20333972/175268584-de60daf4-3292-4ef6-aa42-da015bb66070.mov



Could you make it so, that if the value supplied is not a string with length greater than zero, then the widget should not crash, instead just say invalid color on the label, and show empty color box if possible

sherfin94 (approved):

PR #3317, opened 2022-06-18
Cypress test for the manage SSO page in the single workspace. 

ajith-k-v:
Fixes: #3316 

1. Added cypress test for the manage SSO page in the single workspace.
2. Updated common selectors and text files.
3. Updated common utils file.
4. Added data-cy attributes for the elements.

Screenshot 2022-06-18 at 9 53 15 PM


gsmithun4 (approved):
LGTM

PR #3334, opened 2022-06-20
[enhancement] Removed version create modal pop up when opening a newly created app

### LINKED ISSUE #3294: Remove version create modal pop up when opening a newly created app
There is not much added value to a user to show version create modal to set up version name on initial app creation. 
This was added before the version name edit feature was not present.

create-app

Remove the version create modal pop-up and have the version v1 created on app creation by default. If the user wants to change the version name, can click on the versions dropdown and edit it.

None

shah21:
Resolves #3294


sherfin94 (changes requested):
Can we make it so, that when the app is created, instead of the toast saying `Version created`, say `Application created`  or say nothing at all. User might be a bit confused when they see `Version created` when they create an application
image

tonycurcuru7:
Looks good.

akshaysasidrn (approved):
LGTM

PR #3339, opened 2022-06-20
Cypress test for the manage sso page multi workspace

ajith-k-v:
Fixes: #3337 

1. Added cypress test for the manage SSO page in the multi workspace.
2. Updated manageSSO selector and text files.
3. Added data-cy attribute for login URL element

Screenshot 2022-06-20 at 7 55 40 PM


gsmithun4 (approved):
LGTM

PR #3342, opened 2022-06-21
[bugfix] Now users can't create multiple versions with the same name for an app

### LINKED ISSUE #3329: Users should not be able to create multiple versions with same name
### Version Information
Latest
### Environment

All

Users should not be able to create multiple versions with the same name

currently, we can create multiple versions with the same name, 
### Screenshots or Screencast

Screenshot 2022-06-20 at 2 13 36 PM

### Any possible solutions?
I think maybe adding a unique constraint is a good idea.

Yes

shah21:
Resolves #3329


server/src/services/apps.service.ts
```diff
@@ -294,6 +294,14 @@ export class AppsService {
       where: { id: versionFromId },
     });
 
+    const isVersionNameExisted = await this.appVersionsRepository.findOne({
```
akshaysasidrn, line 297: When concurrent requests are served, there is a chance that this check passes. The DB would still throw an error due to a unique constraint that is now kept in. 

Currently, the server wouldn't handle it thus returning an unintelligible error to the user with 500 status. This needs to be handled. That can be a separate PR. Probably we can configure typeorm to handle such DB errors. This will in turn make our code easier and we wouldn't have to do these checks manually to raise a custom error message.

akshaysasidrn (changes requested):

server/migrations/1655726247638-AddUniqueConstraintToVersionName.ts
```diff
@@ -0,0 +1,45 @@
...
+    await queryRunner.dropUniqueConstraint('app_versions', 'name_app_id_app_versions_unique');
+  }
+
+  private async migrateVersions(entityManager: EntityManager) {
+    const versions = await entityManager.find(AppVersion);
+    return (async () => {
+      const versionsChanged = [];
+      for (const version of versions) {
+        const { appId, name, id } = version;
+        if (versionsChanged.includes(id)) continue;
```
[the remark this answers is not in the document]
shah21: I didn't completely understand this way. even if a get the versions through organization -> apps -> versions, I should check when the next version is already changed or not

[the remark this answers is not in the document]
akshaysasidrn: Yeah the check is still needed, just that we can scope it via org -> apps -> versions to reduce the memory usage at once.

server/src/services/apps.service.ts
```diff
@@ -294,6 +294,14 @@ export class AppsService {
...
+    const versionNameExists = await this.appVersionsRepository.findOne({
+      where: { name: versionName, appId: app.id },
+    });
+
+    if (versionNameExists) {
+      throw new BadRequestException('Version name already exists.');
+    }
+
     let appVersion: AppVersion;
     await getManager().transaction(async (manager) => {
```
gsmithun4, line 306:
```suggestion
    try {
      await getManager().transaction(async (manager) => {
        appVersion = await manager.save(
          AppVersion,
          manager.create(AppVersion, {
            name: versionName,
            app,
            definition: versionFrom?.definition,
            createdAt: new Date(),
            updatedAt: new Date(),
          })
        );
        await this.setupDataSourcesAndQueriesForVersion(manager, appVersion, versionFrom);
      });
    } catch (error) {
      if (error.code === '23505' && error.constraint === 'name_app_id_app_versions_unique') {
        throw new BadRequestException('Version name already exists.');
      }
      throw new Error();
    }
```
Is it feasible? we could save an extra query
@akshaysasidrn

gsmithun4 (changes requested):

server/src/services/apps.service.ts
akshaysasidrn: This is feasible, but let's make changes to handle db exceptions separately on a different PR. We can handle this throughout the app to return appropriate responses gracefully. @shah21 is looking into this already.

server/migrations/1655726247638-AddUniqueConstraintToVersionName.ts
```diff
@@ -0,0 +1,51 @@
...
+  private async migrateVersions(entityManager: EntityManager) {
+    const apps = await entityManager.find(App, { relations: ['appVersions'] });
+
+    return (async () => {
+      for (const app of apps) {
+        const { appVersions } = app;
+        await (async () => {
+          for (const version of appVersions) {
+            const { name, id } = version;
+            const vesrionExists = await entityManager.findOne(AppVersion, { id, name });
```
akshaysasidrn, line 31: spell error here 🥸

server/src/services/apps.service.ts
```diff
@@ -511,6 +519,17 @@ export class AppsService {
     if (body.definition) editableParams['definition'] = body.definition;
     if (body.name) editableParams['name'] = body.name;
 
+    if ('name' in editableParams && Object.keys(editableParams).length == 1) {
```
gsmithun4, line 522: `if (body.name)`
is this enough?

gsmithun4 (changes requested):

server/src/services/apps.service.ts
shah21: Yes, that was enough.

gsmithun4 (approved):
LGTM

akshaysasidrn (approved):

PR #3347, opened 2022-06-21
[Bugfix] Exposed variables do not get the updates - Listview widget

### LINKED ISSUE #3346: Listview widget: Exposed variables do not get the updates
- If the `List data` property (the data which is consumed by the widget) gets any updates, should update the exposed variable `data` with the current consumed data.
- `selectedRow` exposed variable should be updated with the current selected row's data of the widget

- `selectedRow`: the exposed value for the first row's data is an empty object, and for the next is undefined when respective rows are selected or clicked.

1. Add a Listview widget to the canvas from the widget manager
2. Open the Leftside inspector and expand the **components** node and the widget node respectively.
3. The exposed value `data` has always one element (empty object)

### Screenshots or Screencast
Screenshot 2022-06-21 at 4 45 23 PM

arpitnath:
Resolves #3346


arpitnath:
Need to test this PR again as discussed @Nandinisaha13 
cc: @sherfin94

sherfin94:
Halting review for this one for the time being as I'm considering some urgent rearchitecting on the List View.

arpitnath:
Closing this PR, as the issue is fixed in #3454

Issue #333, opened 2021-06-26
Feature: previews for CodeHinter fields


Navaneeth-pk:
### Is your proposal related to a problem?

While typing something that includes dynamic variables on the CodehHinter field, we don't know what will be the actual resolved value of the entered text.

### Describe the solution you'd like

Show the preview with resolved dynamic variables at the bottom of the CodehHinter fields. This feature should be optional.

### If the feature is approved, would you be willing to submit a PR?
yes

PR #3353, opened 2022-06-21
Feature/add toast close button

DonRomaniello:
This turned out to be very straightforward, following the setup here:

https://react-hot-toast.com/docs/toast-bar

I've tested this with multiple toast types.
Screen Shot 2022-06-21 at 4 07 22 PM

Works for all:

Screen Shot 2022-06-21 at 4 05 51 PM


kavinvenkatachalam (changes requested):
Hey @DonRomaniello,

Suggested a few changes below. Additionally, when we change to `dark` mode the close button remains `dark` and it is invisible. So, I suggest you to pass a className to the existing `toastOption` and add CSS in `theme.scss` file to support dark mode. 

For example,

`toastOptions = {
        className: 'toast-dark-mode',
        style: {
          borderRadius: '10px',
          background: '#333',
          color: '#fff',
        }`
        
  **In theme.scss**  
   `.toast-dark-mode {
  .btn-close {
    filter: brightness(0) invert(1);
  }
}`

DonRomaniello:
@kavinvenkatachalam, thank you for your feedback. Thanks for bearing with me, this will be my first contribution outside of a classroom setting.

Initiating another pull request with the implemented changes.

PR #3364, opened 2022-06-22
Reverting woocommmerce operations file 

### LINKED ISSUE #3363: Woocommerce operations file is present although excluded through git ignore

plugins/packages/woocommerce/lib/operations.json should be ignored

plugins/packages/woocommerce/lib/operations.json file is present 

1.Pull latest develop and build plugins

stepinfwd:
close #3363


gsmithun4 (approved):
LGTM

PR #3369, opened 2022-06-23
Now the table will show the user groups ordered by created time

### LINKED ISSUE #3368: The user group table is not showing data by created time.
### Version Information
Latest
### Environment

All

The user group table should show the data as ordered by its created time
1. Add more groups to the table & observe

### Screenshots or Screencast

image
### Any possible solutions?
Order By the created time

Yes

shah21:
Resolves #3368

### DIFF
--- a/server/src/services/group_permissions.service.ts
+++ b/server/src/services/group_permissions.service.ts
@@ -211,7 +211,8 @@ export class GroupPermissionsService {
 
   async findAll(user: User): Promise<GroupPermission[]> {
     return this.groupPermissionsRepository.find({
-      organizationId: user.organizationId,
+      where: { organizationId: user.organizationId },
+      order: { createdAt: 'ASC' },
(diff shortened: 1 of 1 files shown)


server/src/services/group_permissions.service.ts
```diff
@@ -211,7 +211,8 @@ export class GroupPermissionsService {
 
   async findAll(user: User): Promise<GroupPermission[]> {
     return this.groupPermissionsRepository.find({
-      organizationId: user.organizationId,
+      where: { organizationId: user.organizationId },
+      order: { createdAt: 'ASC' },
```
gsmithun4, line 215: Should it be `DESC`? last created group at first?

gsmithun4 (changes requested):

akshaysasidrn (approved):

Issue #3108, opened 2022-05-23
Confirm password field for profile- change password section


gsmithun4:
### Is your proposal related to a problem?
There is no confirm password field in user profile page 

### Describe the solution you'd like
Add confirm password field and verify the new password and confirm password are same

### If the feature is approved, would you be willing to submit a PR?
No

deepakb:
Hi @gsmithun4,
Could you please assign this task to me? I want to work on this.

withshubh:
Hey @deepakb 
Are you still working on this issue?

deepakb:
> Hey @deepakb Are you still working on this issue?

I am facing some issues to set up the project locally.  I'll try once again and update.

withshubh:
Hey @deepakb 👋 
Let us know if you're still working on it?
If you have any questions please feel free to ask any on our [Slack Community](https://tooljet.slack.com/join/shared_invite/zt-r2neyfcw-KD1COL6t2kgVTlTtAV5rtg#/shared-invite/email).

Syed-Ansar:
Hey @withshubh I would like to work on this issue if deepakb isn't working.

withshubh:
Hey @deepakb 
Unassigning this issue as there was no update for more than 20 days.

withshubh:
Hey @Syed-Ansar - Assigned this issue to you.

Syed-Ansar:
Hey @withshubh I have some questions over the alignment of the input elements. Before moving forward I would like to discuss. Let me know If you are available for discussion.

withshubh:
@Syed-Ansar - I'd request you to please join our [Slack community](https://tooljet.slack.com/join/shared_invite/zt-r2neyfcw-KD1COL6t2kgVTlTtAV5rtg#/shared-invite/email) and post all your questions there.

Syed-Ansar:
Hey @withshubh can you review #3361

withshubh:
Hey @Syed-Ansar 
Someone from the team will review your PR in a day or two. In the meantime, you can pick the next issue you want to work on.

Syed-Ansar:
Hey @withshubh I have made another PR please review it. #3365

shubh22:
Issue has been fixed.

PR #3374, opened 2022-06-23
Added confirm in profile settings

Syed-Ansar:
Added Confirm password field for profile - change password section

fixes https://github.com/ToolJet/ToolJet/issues/3108


arpitnath (changes requested):
can you please revert the changes to `theme.scss`

Issue #3225, opened 2022-06-08
Implement draggable event handlers in widget's event property


shivangT:
### What is the expected behaviour?
While adding multiple events under the widget's properties, it should be draggable to change the order of events

### What is the current behaviour?
We're unable to drag and change the order of events, in order to do that we've to delete all the events and add them as per the new sequence.

### How to reproduce the issue?

1. Drag and drop a widget ( for eg Button)
2. Now inspect and check out its properties on the left pane. 
3. Expand the events panel and add different types of event handlers

### Screenshots or Screencast
<img width="299" alt="Screenshot 2022-06-08 at 1 46 38 PM" src="https://user-images.githubusercontent.com/32157641/172567435-59054644-1631-4efd-9cf7-88c360f6a390.png">

### Any possible solutions?

This feature already exists on our platform. You can refer to the table widget. Drag and drop a table and inspect it. Look for the "Columns" property on the right. We're able to change the position of the columns by drag and drop property.

<img width="300" alt="Screenshot 2022-06-08 at 1 48 19 PM" src="https://user-images.githubusercontent.com/32157641/172567746-3a18f5ce-93ff-4f3a-846d-82a031b52626.png">

shivangT:
hey @kavinvenkatachalam ,

Are you working on this? What's the status?

kavinvenkatachalam:
@shivangT Merged the PR a few days back #3381

withshubh:
https://github.com/ToolJet/ToolJet/pull/3381

PR #3381, opened 2022-06-24
[Feature]: Re-Order Events

### LINKED ISSUE #3380: [Feature]: Re-Order Events

Right now it is not possible to re-order the events

Feature to re-order events with the help of `react-beautiful-dnd`

No

Yes

kavinvenkatachalam:
Resolves #3380 #3225


shubh22:
Task linked: [CU-2h1aaux Re-order Events](https://app.clickup.com/t/2h1aaux)

frontend/src/Editor/Inspector/EventManager.jsx
```diff
@@ -603,62 +605,156 @@ export const EventManager = ({
...
-                      </svg>
-                    </span>
-                  </div>
-                </div>
-              </div>
+  const reorderEvents = (startIndex, endIndex) => {
+    const result = component.component.definition.events;
+    const [removed] = result.splice(startIndex, 1);
+    result.splice(endIndex, 0, removed);
+    eventsChanged(result);
```
sherfin94, line 612: @kavinvenkatachalam, Even if I removed this `eventsChanged` call, the reordering function will still work, due to the mutation caused by the preceding two `splice` calls. Though I don't see any bugs, I'm feeling a bit scared about introducing state mutation here, can we do it without mutation? At worst we can just use 
`const result = _.cloneDeep(component.component.definition.events);`  on line 609

Perhaps we can do something similar on the third splice call that exists on the `removeHandler` function as well.

sherfin94 (changes requested):
Suggested a code enhancement.

sherfin94 (approved):

PR #3405, opened 2022-06-27
Chore: Make typescript non dev dependency for plugins build

### LINKED ISSUE #3404: AMI build failing on github action
### Environment
Github Action

Should build ami image

fails due to missing tsc dependency when build on node env prod

Tagged version trigger for ami build fails

### Any possible solutions?
Make typescript as non dev dependency as plugins folder build in both prod or dev env requires typescript

akshaysasidrn:
- fixes #3404


gondar00:
@akshaysasidrn can we add it to `peerDependencies` instead, currently when we do a `tooljet plugin create <plugin_name>` the dependancies are overridden by lerna and typescript dependency gets removed

akshaysasidrn:
@gondar00 Have made the change

gondar00 (approved):

PR #3414, opened 2022-06-28
[enhancement] Handle internal database exceptions in a better way

### LINKED ISSUE #3413: Handle db exceptions in a better way

Currently, our backend only returns 500 code for all database Exceptions. 

Globally, Handle DB exceptions separately from HTTP errors. By doing this, a developer can easily understand what will be the error throwing by the Database and will not be needed to write extra catches from the controller or service.

Yes

shah21:
Resolves  #3413


shubh22:
Task linked: [CU-2h1aprj Handle DB Exceptions](https://app.clickup.com/t/2h1aprj)

server/src/all-exceptions-filter.ts
```diff
@@ -10,19 +23,56 @@ export class AllExceptionsFilter implements ExceptionFilter {
...
+  private handleQueryExceptions(exception: any): ErrorResponse {
+    const status = HttpStatus.UNPROCESSABLE_ENTITY;
+    const code = (exception as any).code;
+    let message: string = (exception as QueryFailedError).message;
+
+    switch (code) {
+      case PostgresErrorCode.UniqueViolation:
+        message = 'Already Existed!';
+        break;
+      case PostgresErrorCode.CheckViolation:
```
akshaysasidrn, line 60: Let's just raise Postgres error for CheckViolation, ForeignKeyViolation with 500. This can be checked for with validations at frontend/backend. Moreover, we would need to know of such exceptions to add in any such validation that was missed.

akshaysasidrn (approved):

PR #3427, opened 2022-06-29
[Feature]: Added box-shadow property

### LINKED ISSUE #3426: [Feature]: Box-Shadow Property

Support `box-shadow` property for all the widget

Add `box-shadow` property to universal props

NA

Yes

kavinvenkatachalam:
Resolves #3426


sherfin94 (commented):
Hey @kavinvenkatachalam , works great.

Is it possible to move this to a folder name `General` on the styles tab? The reason why I'm asking this, is that shadow property might also come for individual components like `Text`, for example, so it might become a bit ambiguous. If its too difficult lets merge this.

sherfin94 (approved):

PR #3449, opened 2022-06-30
Bugfix :: query data if filtered in one place affects everywhere displaying query list 

### LINKED ISSUE #3447: [ Bugfix ]:: When query is filtered affects different places where query list is accessed 🐛
Total query to be present.

https://user-images.githubusercontent.com/31006028/176656162-445d4f49-ff57-40e3-bcf6-622e65fc5918.mov

1.Filter query from query list 
2.Check query list with help of a button and event runQuery.

stepinfwd:
closes #3447


kavinvenkatachalam (approved):
LGTM

Issue #3470, opened 2022-07-04
[Feature] Add a notification center


gondar00:
### Is your proposal related to a problem?
In #3442 whenever a user is mentioned he will be notified using a email that someone has mentioned him, similarly all his notifications can be clubbed inside a notification center.

### Describe the solution you'd like
Add notification center.

UI: 
<img width="1366" alt="image" src="https://user-images.githubusercontent.com/12490590/177153196-0dcd8886-8d9a-48e7-9984-1a9562e933d5.png">
<img width="1390" alt="image" src="https://user-images.githubusercontent.com/12490590/177153592-d3c6d82b-92eb-44c5-9636-118a05151c97.png">

### If the feature is approved, would you be willing to submit a PR?
Yes

PR #3484, opened 2022-07-05
feat: add notification center

gondar00:
fixes: #3470


frontend/src/_components/NotificationCenter/Notification.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import moment from 'moment';
+import { commentsService } from '@/_services';
+
+export const Notification = ({ id, creator, comment, updatedAt, commentLink, isRead, fetchData }) => {
+  const getComment = (comment) => {
+    var regex = /(\()([^)]+)(\))/g;
```
TheLearneer, line 7: Isn't it better not to use `var`?

server/src/services/comment.service.ts
```diff
@@ -88,6 +91,50 @@ export class CommentService {
...
+          id: notification.id,
+          creator,
+          comment: notification.comment.comment,
+          createdAt: notification.comment.createdAt,
+          updatedAt: notification.comment.updatedAt,
+          commentLink,
+          isRead: notification.isRead,
+        };
+      });
+      return Promise.all(_notifications);
```
gsmithun4, line 132: should use await to resolve all `_notifications`

server/src/services/comment.service.ts
```diff
@@ -124,6 +171,16 @@ export class CommentService {
     return foundComment;
   }
 
+  public async updateCommentUser(commentUserId: string, body: any) {
+    const item = await this.commentUsersRepository.update(commentUserId, { ...body });
+    return item;
+  }
+
+  public async updateAllCommentUser(body: any) {
+    const item = await this.commentUsersRepository.update({}, { ...body });
```
gsmithun4, line 180: please pick and pass the updatable params, passing entire body will cause security issues

gsmithun4 (changes requested):

server/src/services/comment.service.ts
gondar00: async/await implicitly returns promise,

```
return Promise.all(_notifications);
```
is same as 
```
return await Promise.all(_notifications);
```

server/src/services/comment.service.ts
gondar00: got it, I've updated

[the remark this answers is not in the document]
gondar00: Added FindCondition

shubh22:
Task linked: [CU-2h1aq0e Create comment notifications](https://app.clickup.com/t/2h1aq0e)

PR #3499, opened 2022-07-06
s3 upload object not saving data.

### LINKED ISSUE #3498: S3 upload object not saving data.
s3 upload query to save uploaded  data.

when uploading data to  bucket data is not shown.

1.upload data to s3 data bucket.

Yes.

stepinfwd:
closes #3498

### DIFF
--- a/plugins/packages/s3/lib/operations.ts
+++ b/plugins/packages/s3/lib/operations.ts
@@ -52,7 +52,7 @@ export async function getObject(client: S3Client, options: object): Promise<obje
 
 export async function uploadObject(client: S3Client, options: object): Promise<object> {
   const encoding = options['encoding'] || 'utf8';
-  const body = new Buffer(options['data'], encoding);
+  const body = new Buffer(uploadData(options['data']), encoding);
   const command = new PutObjectCommand({
     Bucket: options['bucket'],
     Key: options['key'],
@@ -74,4 +74,10 @@ export async function signedUrlForPut(client: S3Client, options: object): Promis
     expiresIn: options['expiresIn'] || 3600,
   });
   return { url };
+}
+function uploadData(data: any) {
+  if (!data) {
+    return undefined;
+  }
+  return typeof data === 'string' ? data : JSON.stringify(data);
 }


plugins/packages/s3/lib/operations.ts
```diff
@@ -75,3 +75,9 @@ export async function signedUrlForPut(client: S3Client, options: object): Promis
   });
   return { url };
 }
+function uploadData(data: any) {
+  if (!data) {
+    return undefined;
```
gsmithun4, line 80:
```suggestion
    return;
```

plugins/packages/s3/lib/operations.ts
```diff
@@ -75,3 +75,9 @@ export async function signedUrlForPut(client: S3Client, options: object): Promis
   });
   return { url };
 }
+function uploadData(data: any, contentType: string) {
+  if (!data) {
+    return;
+  }
+  return typeof data === 'object' && contentType == 'application/json' ? JSON.stringify(data) : data;
```
gsmithun4, line 82: contentType can be `application/json; charset=utf-8`
shall we check if contentType includes `application/json`

gsmithun4 (changes requested):

Nandinisaha13:
<html>
<body>
<google-sheets-html-origin>Test Case

Test Cases | Steps | Result
-- | -- | --
Should be able to read object. | Create an application. | Application gets created.
  | Click on source. | Datasource pop up opens.
  | Click on add button. | Add new data source pop up opens up.
  | Click on AWS S3 data source. | AWS S3 pop opens up.
  | Enter the credentials & test the connection. | Credentials are entered and connection is verified.
  | Save it. | AWS S3 is saved.
  | Create a query for upload object with key, upload data, bucket, content type(application/json, csv/text, application/json; charset=utf-8) & click preview. | The query is executed.
  | Create another query for read object with bucket and key & click preview. | The query is executed & object can be read.


</body>
</html>

gsmithun4 (approved):
LGTM

Issue #3514, opened 2022-07-07
Bug : [widgetConfig] set 'show' to false in the exposedVariable of Modal Widget


manishkushare:
### Environment
Development

### What is the expected behavior?
In the inspector, the initial show value of Modal Widget inside components should be false

### What is the current behavior?
In the inspector, the initial value show value of the Modal Widget inside components is Null
### How to reproduce the issue?

1. Drag the Modal widget in the editor
2. Open the Inspector from the top of the left sidebar
3. In the Inspector ---> open components ---> open Modal ---> show (Initially it is showing Null)

### Screenshots or Screencast
<img width="1440" alt="Screenshot 2022-07-07 at 1 45 06 PM" src="https://user-images.githubusercontent.com/37823141/177727856-6275da9c-4748-48e8-98e0-740e081cd9ed.png">

### Any possible solutions?
1. Look for  widgetConfig.js  file 
Inside widgetConfig.js ---> look for Modal widget ---> Inside Modal Widget, Look for show value inside exposedVariable, set it's default value to false

### Can you identify the location in the driver source code where the problem exists?
path : (frontend --> src --> Editor --> WidgetManager --> widgetConfig.js)

PR #3521, opened 2022-07-07
Changed the exposedVariables show value to be false instead of null in widgetConfig.js

NoahMLoomis:
This fixes bug #3514


shubh22:
Task linked: [CU-2hr96xk Bug : [widgetConfig] set 'show' to false in the exposedVariable of Modal Widget](https://app.clickup.com/t/2hr96xk)

PR #3530, opened 2022-07-07
[enhancement]JSON tree component : handles "Property accessors, for properties starting with numeric digit"

### LINKED ISSUE #3520: JSON tree component needs to handle "Property accessors"
### Description
**Component** : LeftSidebar Inspector 
- nodes which are of **Object** type with property/keys starting with numeric digit 
For example: 
```js
{
 1stDay: {...},
 ...,
}
```
**Copy path to clipboard** option, return the absolute path of the node, for the above example the returned path will be `root.key1.key2.1stDay`.
Trying to resolve this path will throw a **Syntax Error** (In Code Hinters' preview box)

In the `object.property` syntax, the property must be a valid JavaScript identifier(variable, function, or property.).

Identifiers are case-sensitive and can contain **Unicode letters, $, _, and digits (0-9), but may not start with a digit**

### Any possible solutions?
Use **bracket notation** syntax `object[property]`

In the object[property] syntax, the property is just a string or Symbol. So, it can be any string, including '1foo', '!bar!', or even ' '.

Ref: [Read More about Property accessors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors)

arpitnath:
Resolves #3520 

### Screenshot(s)
**Before**
Screenshot 2022-07-08 at 1 31 34 AM

**After**
Screenshot 2022-07-08 at 1 32 14 AM

### DIFF
--- a/frontend/src/_ui/JSONTreeViewer/JSONTreeViewer.jsx
+++ b/frontend/src/_ui/JSONTreeViewer/JSONTreeViewer.jsx
@@ -176,7 +176,12 @@ export class JSONTreeViewer extends React.Component {
         currentPath = prevRelPath ? `${prevRelPath}.${node}` : node;
 
         if (prevType === 'Object') {
-          abs = `${prevPath}.${node}`;
+          //use bracket notation if the node starts with a numeric digit
+          if (node.match(/^\d/)) {
+            abs = `${prevPath}["${node}"]`;
+          } else {
+            abs = `${prevPath}.${node}`;
+          }
         } else if (prevType === 'Array') {
           abs = `${prevPath}[${node}]`;
         } else {


shubh22:
Task linked: [CU-2hrac31 JSON tree component needs to handle "Property accessors"](https://app.clickup.com/t/2hrac31)

Nandinisaha13:
<html>
<body>
<google-sheets-html-origin>Test Case

Test Cases | Steps | Result
-- | -- | --
Should be able to have correct path of identifier starting with numeric digits | Create an application. | Application is created.
  | Create a query for run js with const x = {   obj1st: 1,   Obj2nd: 2,   '3rdObj': 3, };  return x;as the code | The query is created & run.
  | Drag and drop a button widget. | Button widget is displayed on canvas.
  | Click on the inspector and pin it. | The inspector is pinned.
  | Copy to clipboard for the 3rd object in queries. | Path is copied.
  | Go to button's code hinter. | Code hinter opens up.
  | Paste the path in between {{}}(the path is [] without dot notation) | It displays the correct value of it.


</body>
</html>

sherfin94 (approved):

PR #3559, opened 2022-07-10
Remove the border from the bottom of the page where we have pagination

Ankush263:


Navaneeth-pk:
Please create an issue and get it assigned before creating a PR. Unable to understand why this change is needed. 
Also `frontend/src/_styles/tabler.scss` should not be modified as it is a minified CSS file. 
@Ankush263

Issue #3581, opened 2022-07-11
Hide properties in container widget


shivangT:
### What is the expected behaviour?
Hide the property option for this widget as we've no properties associated with this widget for now. Put a null check to hide the accordion if no property is available.

### What is the current behaviour?
When we inspect the container widget, we can see the properties tab to be empty.

### How to reproduce the issue?

1. Go to this [link](https://app.tooljet.com/)
2. Now click on "create new application"
3. Now drag and drop the container widget from the widgets inventory
4. Inspect the widget and explore its properties
5. You'll find no properties when you expand the "Property" accordion

### Screenshots or Screencast
<img width="1440" alt="Screenshot 2022-07-11 at 6 56 38 PM" src="https://user-images.githubusercontent.com/32157641/178274800-c92b6d54-96fd-44b4-8d67-fbdebe4a98a4.png">

PR #3589, opened 2022-07-11
Added hide logic to widget properties accordion menu item when it contains no items

Nathan-Bisson:
Fixes: #3581


Nathan-Bisson:
@arpitnath Thank you for the review! I noticed that the lint check is failing. I reviewed the logs and it appears to be failing on unrelated files to this PR. Wondering what the best course of action is for this here? I'm happy to get these resolved, but am wondering if it would be best to do this under a new branch, or this current one? Please let me know what you think!

sherfin94 (approved):

Issue #3579, opened 2022-07-11
Remove tooltip option from Modal widget


shivangT:
### What is the expected behaviour?
Need to remove this option from the widget's properties since it fulfills no use case for it.

### What is the current behaviour?
We've a tooltip option under general acccordion. 

### How to reproduce the issue?

1. Go to this [link](https://app.tooljet.com/)
2. Now click on "create new application"
3. Now drag and drop the Modal widget and inspect it
4. On the right-hand side inspector hide the General accordion since we have only tooltip in it.
 
### Screenshots or Screencast
<img width="1440" alt="Screenshot 2022-07-11 at 6 50 57 PM" src="https://user-images.githubusercontent.com/32157641/178273708-36a708e5-ffe6-450b-a751-f721899167be.png">

PR #3596, opened 2022-07-12
Added widget property filters, and filtered general from modal widget

Nathan-Bisson:
- Noticed a number of issues in the queue relate to filtering various widget setting's from the accordions list
- Change in this commit adds on the ability to filter various sections from a widget's settings by adding it to a filter object to the relevant section
- This will prevent if statements getting too long if more sections need to be filtered from a widget's settings in the future
- Added in filtering for the Modal widget's General section as specified in the linked issue

Fixes: #3579


sherfin94 (approved):
Sweet! @Nathan-Bisson

PR #3610, opened 2022-07-13
Hotfix: Split data and db migrations

### LINKED ISSUE #3605: Separate data and DB migrations

Problem when running data migrations as a part of DB migrations is that data migration can be run before all the DB migrations are done. Thus, when the app tries to load a particular entity, it may error as it can have updated columns for which DB migration is yet to be done. As a workaround for this, we currently have to explicitly select the columns on the entity such that the whole of the entity isn't tried to be loaded into the memory.

Provide a means to separate b/w data and DB migrations. Data migrations can be made to run after DB migrations so that entity load errors can be skipped and those can be wrapped in separate transaction with a different ormconfig. 

n/a

yes

akshaysasidrn:
fixes #3605 

- Separates out config for DB and data migrations
- Data migrations are to be kept in the folder data-migrations/*
- Db migrations and data migrations are run in separate transactions
- Adds a new npm script db:migrate:data which will be invoked with db:migrate


server/data-migrations/1654596810662-ConvertAllUserEmailsToLowercaseAndDeleteDuplicateUsers.ts
```diff
@@ -24,50 +26,34 @@ export class ConvertAllUserEmailsToLowercaseAndDeleteDuplicateUsers1654596810662
       .createQueryBuilder('users')
       .select(['users.id', 'users.email', 'users.invitationToken'])
       .leftJoinAndSelect('users.organizationUsers', 'organizationUsers')
-      .leftJoinAndSelect('users.apps', 'apps')
```
akshaysasidrn, line 27: `@Afterload` query for app versions was being triggered for each rows selected from the app entity and thus resulting in timeouts when running migrations

PR #3635, opened 2022-07-18
[hotfix] Public apps are redirecting to login page while using environment variables inside an app

### LINKED ISSUE #3634: Public apps are redirecting to login page while using environment variables inside an app
### Version Information
Latest
### Environment

All

Everyone should be able to view the app 

Public apps are redirecting to the login page while using environment variables inside an app

1. Create an app with a text widget that contains environment variable value
2. Make the app public
3. Run the public app like as a non-tooljet user

Yes

shah21:
fixes #3634


shubh22:
Task linked: [CU-2kbrgbq Public apps are redirecting to login page while using environment variables inside an app](https://app.clickup.com/t/2kbrgbq)

frontend/src/Editor/Viewer.jsx
```diff
@@ -127,16 +127,22 @@ class Viewer extends React.Component {
...
     };
-    await orgEnvironmentVariableService.getVariables().then((data) => {
-      data.variables.map((variable) => {
-        variables[variable.variable_type][variable.variable_name] =
-          variable.variable_type === 'server' ? 'HiddenEnvironmentVariable' : variable.value;
-      });
+
+    let variablesResult;
+    if (this.state.currentUser) {
+      variablesResult = await orgEnvironmentVariableService.getVariables();
```
gsmithun4, line 138: It will fail for this condition:
- User logged in to workspace A
- opening public app of workspace B
- `organization-variables` API will return variables of workspace A instead of B

Shall we check if app is public and load the data accordingly instead of checking the availability of `currentUser`?

server/src/controllers/org_environment_variables.controller.ts
```diff
@@ -29,7 +29,13 @@ export class OrgEnvironmentVariablesController {
...
   @Get()
   async get(@User() user) {
-    const result = await this.orgEnvironmentVariablesService.fetchVariables(user);
+    const result = await this.orgEnvironmentVariablesService.fetchVariables(user, null);
+    return decamelizeKeys({ variables: result });
+  }
+
+  @Get(':app_slug')
+  async getVariablesFromApp(@Param('app_slug') slug) {
+    const result = await this.orgEnvironmentVariablesService.fetchVariables(null, slug);
```
gsmithun4, line 38: Shall we add a guard for this route? It should be allowed only if the app is public

gsmithun4 (changes requested):

server/src/services/org_environment_variables.service.ts
```diff
@@ -6,27 +6,36 @@ import { Repository } from 'typeorm';
...
   constructor(
     @InjectRepository(OrgEnvironmentVariable)
     private orgEnvironmentVariablesRepository: Repository<OrgEnvironmentVariable>,
-    private encryptionService: EncryptionService
+    private encryptionService: EncryptionService,
+    private appsService: AppsService
   ) {}
 
-  async fetchVariables(currentUser: User): Promise<OrgEnvironmentVariable[]> {
+  async fetchVariables(currentUser: User, slug: string): Promise<OrgEnvironmentVariable[]> {
```
gsmithun4, line 20: Shall we pass `organizationId` to the function? 
```
fetchVariables(currentUser.organizationId)
```
and for new route set app to request from `IsPublic` guard and 
```
fetchVariables(app.organizationId)
```

server/src/modules/organizations/organizations.module.ts
```diff
@@ -25,6 +25,17 @@ import { AppGroupPermission } from 'src/entities/app_group_permission.entity';
 import { UserGroupPermission } from 'src/entities/user_group_permission.entity';
 import { EncryptionService } from '@services/encryption.service';
 import { OrgEnvironmentVariablesService } from '@services/org_environment_variables.service';
```
gsmithun4, line 27: try removing `OrgEnvironmentVariablesService` import, other imports won't be required

gsmithun4 (changes requested):

server/src/modules/organizations/organizations.module.ts
shah21: that is not working

server/src/modules/org_environment_variables/is-public.guard.ts
```diff
@@ -0,0 +1,13 @@
+import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common';
+import { AppsService } from '@services/apps.service';
+
+@Injectable()
+export class IsPublicGuard implements CanActivate {
+  constructor(private appsService: AppsService) {}
+
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const { params } = context.switchToHttp().getRequest();
+    const app = await this.appsService.findBySlug(params['app_slug']);
```
gsmithun4, line 10:
```
if (!request?.params?.app_slug) {
  return false;
}
const request = context.switchToHttp().getRequest();
const app = await this.appsService.findBySlug(request.params.app_slug);
request.app = app;
return !!app?.isPublic;
```
Shall we check if `app_slug` is present in request and set the app in request if app available, 
and get the `app` from controller using decorator?

gsmithun4 (changes requested):

gsmithun4 (approved):
LGTM

PR #3662, opened 2022-07-25
Bug Fix : converted ms into second by multiplying it with 1000 for success notification

### LINKED ISSUE #3661: Bug : Notification on success not running for desired duration
### Environment
Development

When show notification on success is enabled, the desired success message should be notified for a selected duration. Let's say, if the notification duration is set to 5, then the message should be reported for 5 seconds

Whenever we set some notification duration, notification duration is running for some milli sec only

1. Create one query or one can use any previously created query
2. Go into the advanced tab and toggled ON/enable, show notification on success
3. You will be able to see two input fields, one for the success message and the other for notification duration
4. Put any value in the notification duration field, and run the query
5. You can see success notification just for a few msec, but not for the set notification duration

[Untitled app - Tooljet - 25 July 2022 - Watch Video](https://www.loom.com/share/25d5e773434a4159b9a3c1023d7a8bac)

### Any possible solutions?
Multiply the notification duration by 1000 to convert it into seconds.

appUtils.js

Yes

manishkushare:
Resolve #3661

### DIFF
--- a/frontend/src/_helpers/appUtils.js
+++ b/frontend/src/_helpers/appUtils.js
@@ -752,7 +752,7 @@ export function runQuery(_ref, queryId, queryName, confirmed = undefined, mode) 
           }
 
           if (dataQuery.options.showSuccessNotification) {
-            const notificationDuration = dataQuery.options.notificationDuration || 5000;
+            const notificationDuration = dataQuery.options.notificationDuration * 1000 || 5000;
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-2kbncyx Bug in Show notification in success? (Advanced option)](https://app.clickup.com/t/2kbncyx)

gondar00 (approved):

PR #3710, opened 2022-08-01
Adding PG_DB to the query

MehdiElAouni:
Fixing the "Can't connect to the database. Kindly check the credenials provided in the .env file!" error by adding PG_DB to the psql query

### DIFF
--- a/deploy/ec2/setup_app
+++ b/deploy/ec2/setup_app
@@ -8,7 +8,7 @@ fi
 
 export $(grep -v '^#' .env | xargs)
 
-if psql -d postgresql://$PG_USER:$PG_PASS@$PG_HOST/$PG_DB -c 'select now()' > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


deploy/ec2/setup_app
```diff
@@ -8,7 +8,7 @@ fi
 
 export $(grep -v '^#' .env | xargs)
 
-if psql -d postgresql://$PG_USER:$PG_PASS@$PG_HOST -c 'select now()' > /dev/null 2>&1
+if psql -d postgresql://$PG_USER:$PG_PASS@$PG_HOST/$PG_DB -c 'select now()' > /dev/null 2>&1
```
akshaysasidrn, line 11: Hey @MehdiElAouni, could you change the `$PG_DB` value to `postgres` rather? It is default db which can be used to check for connection. In the current change, it can fail when the particular db isn't created yet.

PR #3746, opened 2022-08-05
[Feature]: Added Localisation

### LINKED ISSUE #3748: [Feature]: Localisation Support

As many people from different countries use Tooljet, we can introduce localisation and make the tool available in their local language.

- Packages `i18next` and `react-i18next` can be used to support multi-language. Package `i18next-http-backend` can be used to fetch the translations asynchronously to the tool and `i18next-browser-languagedetector` can be used to detect the browser language.
- `editor` and `dashboard` folders can be created under the `translation` folder in the assets, to maintain the translation of all the languages. The translation file should be named according to the language code. For example, if is it `English` then the file should be named as `en.json`
- To translate the text, for a functional component we can import `useTranslation` hook and we can call the function `t` with the respective translation key. 
- Similarly for a class component we can use `withTranslation` HOC and we can call the function `t` with the respective translation key. 
-  We can pass a fallback text while calling the `t` function as a second argument which will taken incase the translation key is not found.

No

Yes

kavinvenkatachalam:
Resolves #3748


shubh22:
Task linked: [CU-2h1bftm Localisation using i18n](https://app.clickup.com/t/2h1bftm)

sherfin94 (dismissed):

sherfin94 (approved):

PR #3768, opened 2022-08-08
[Bugfix] If the query is successfully run, success toast should be fired despite its triggered by any source.

### LINKED ISSUE #3767: Toast is only fired when queries are run from query manager with always success type
### Description
Queries can be fired from the Query manager, the LeftSidebar inspector, from events and actions from the RunJS query.

If the query is successfully run, success toast should be fired despite its triggered by any source.

Only queries triggered from Query Manager fires the toast. And if the query's execution fails for any reason, a success toast is fired.

1. Create and save a query in the Query Manager, e.g., RestAPI query
2. Fire the query from Query Manager.
If the query is completed successfully, we get a success toast; if it fails, the error is logged in the Debugger with a success toast.
3. Open the LeftSidebar Inspector, and expand the **queries** node. Hover and click on the run icon.
The query runs but no toast is fired.
4. Add a widget to the Canvas, and add a **Run Query** action on any of the events. (For button widget, add the action with **On click** event)
The query runs but no toast is fired.
5. Create a **Run JS** query and run the RestAPI query.
Syntax: `queries.queryName.run()`
The query runs but no toast is fired.

arpitnath:
Resolves #3767

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -922,11 +922,7 @@ class Editor extends React.Component {
               style={{ marginTop: '3px' }}
               className="btn badge bg-light-1"
               onClick={() => {
-                runQuery(this, dataQuery.id, dataQuery.name).then(() => {
-                  toast(`Query (${dataQuery.name}) completed.`, {
-                    icon: '🚀',
-                  });
-                });
+                runQuery(this, dataQuery.id, dataQuery.name);
               }}
             >
               <div className={`query-icon ${this.props.darkMode && 'dark'}`}>
--- a/frontend/src/Editor/Viewer.jsx
+++ b/frontend/src/Editor/Viewer.jsx
@@ -122,7 +122,7 @@ class Viewer extends React.Component {
   runQueries = (data_queries) => {
     data_queries.forEach((query) => {
       if (query.options.runOnPageLoad) {
-        runQuery(this, query.id, query.name);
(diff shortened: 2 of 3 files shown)


shubh22:
Task linked: [CU-2p77dh6 Toast is only fired when queries are run from query manager with always success type](https://app.clickup.com/t/2p77dh6)

Nandinisaha13:
<html>
<body>

Test Case | Steps | Result
-- | -- | --
Should be able to fire a toast while query is run from inspector. | Create an application. | Application gets created.
  | Create a custom js query. | Query is created.
  | Open the inspector. | Inspector is opened.
  | Click on queries & click on run button. | The query is run & toast is fired.
Should be able to fire a toast while query is run via event handler. | Create an application. | Application gets created.
  | Create a custom js query. | Query is created.
  | Drag & drop a button. | Button is displayed on canvas.
  | Create an event handler with action as run query & select query name. | Event is created.
  | Click on button. | The query is run & toast is fired.
Should be able to fire a toast while query is run via event handler in advanced section. | Create an application. | Application gets created.
  | Create a custom js query. | Query is created.
  | Create another custom query. | Query is created.
  | Go to advanced tabs. | Advanced panel is opened.
  | Create an event handler with action as run query & select query name. | Query is created.
  | Run the second js. | The second query is run & then first is run & toasts are fired.


</body>
</html>

kavinvenkatachalam (approved):
LGTM

Issue #3784, opened 2022-08-11
[Perf] Reduce webpack output bundle size


gondar00:
### Is your proposal related to a problem?
The current bundle generated from webpack weighs around **7.7mb** in production mode. This size can be further optimized by introducing few small changes in the webpack config

### Describe the solution you'd like

1. Use webpack in production mode
The current config was missing a `=` symbol before `--mode` and `production` changing it to [`--mode=production`](https://webpack.js.org/configuration/mode/) gives two tiny benefits are that it enables minification, removes white spaces etc and informs some packages like react to use their production builds by setting NODE_ENV to production

2. Split Code Chunks (Vendor)
Promote loading scripts in parallel by splitting the main.js bundle into two bundles (main bundle and vendor bundle) vendor bundle is a chunk generated by the `node_modules` folder and can be easily added in the webpack config as follows: 

```
  optimization: {
    splitChunks: {
      cacheGroups: {
        vendors: {
          test: /[\\/]node_modules[\\/]/,
          name: "vendor",
          chunks: "all"
        }
      }
    }
  },
```
This gives out two chunks vendor and main as shown in image below: 

<img width="1133" alt="image" src="https://user-images.githubusercontent.com/12490590/184114902-1d8aa660-4d23-40b2-9ff2-aaaf476c0e4e.png">

3. Compressing Bundles
Add gzip compression along with the bundle to enable the web server to serve compressed bundles to the client 
To achieve this have used [compression-webpack-plugin](https://www.npmjs.com/package/compression-webpack-plugin) in conjuction with nestjs [compression](https://docs.nestjs.com/techniques/compression). 

4. Use [esbuild](https://github.com/evanw/esbuild) as bundler and minifier.
Using esbuild to minify and bundle the code the build time reduces by 75%, before adding esbuild the build time was **151119 ms** and after adding it the build time came down to **49474 ms**

Before esbuild: 
<img width="1174" alt="image" src="https://user-images.githubusercontent.com/12490590/184816853-ab6d899c-a65f-4a03-a6c6-4b20876af661.png">
After esbuild: 
<img width="1128" alt="image" src="https://user-images.githubusercontent.com/12490590/184816901-47b05acc-dc83-4644-b909-68c78f88eff7.png">

Note: In case of nginx the [`gzip on`](https://docs.nginx.com/nginx/admin-guide/web-server/compression/) needs to be turned on. 

[...]

PR #3785, opened 2022-08-11
perf: optimize webpack bundle output

gondar00:
fixes: #3784


joy-amoga:
Can you also remove all console.logs in Production Build, so that when we are doing console.log from RunJS, we can see the consoles.

gondar00:
Hi @joy-amoga, I can see that removing console.log might introduce bugs in the code: 

[ref](https://esbuild.github.io/api/#drop): 
image

Navaneeth-pk (approved):

PR #3796, opened 2022-08-16
[Bug]: Fixed PDF Unstable Issue

### LINKED ISSUE #3795: [Bug]: PDF Widget Unstable
### Environment

Production

PDF widget should behave normally when it is resized

The page number keeps on changing and the page crashes while resizing the widget

1. Add a PDF widget to the canvas
2. Resize the widget so that more than one page is visible

### Any possible solutions?

The state is getting updated in the `IntersectionObserver` and when two pages are visible, the page number state keeps updating values which cause the page to be unresponsive. It can be avoided by updating the state only for the first time.

Yes

kavinvenkatachalam:
Resolves #3795


shubh22:
Task linked: [CU-2mthfy7 PDF Widget unstable](https://app.clickup.com/t/2mthfy7)

sherfin94 (approved):

Issue #3808, opened 2022-08-16
Wrap DB queries inside typeorm manager


gsmithun4:
### Is your proposal related to a problem?

Currently we don't have any rollback mechanism if any one of the query failed in a group of queries 

### Describe the solution you'd like
wrap DB calls inside manager, if any error is thrown all the queries executed in current transaction will be rollback

### If the feature is approved, would you be willing to submit a PR?
Yes

PR #3809, opened 2022-08-16
Wrap DB queries inside manager to enable rollback

gsmithun4:
Resolves https://github.com/ToolJet/ToolJet/issues/3808

### DIFF
--- a/server/src/helpers/utils.helper.ts
+++ b/server/src/helpers/utils.helper.ts
@@ -54,7 +54,7 @@ export function lowercaseString(value: string) {
   return value?.toLowerCase();
 }
 
-export async function dbManagerWrapper(operation: (...args) => any, manager: EntityManager): Promise<any> {
+export async function dbTransactionWrap(operation: (...args) => any, manager?: EntityManager): Promise<any> {
   if (manager) {
     return await operation(manager);
   } else {
--- a/server/src/helpers/utils.helper.ts
+++ b/server/src/helpers/utils.helper.ts
@@ -58,7 +58,7 @@ export async function dbTransactionWrap(operation: (...args) => any, manager?: E
   if (manager) {
     return await operation(manager);
   } else {
-    await getManager().transaction(async (manager) => {
+    return await getManager().transaction(async (manager) => {
       return await operation(manager);
     });
   }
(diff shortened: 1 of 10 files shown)


server/test/services/group_permissions.service.spec.ts
```diff
@@ -27,11 +27,7 @@ describe('GroupPermissionsService', () => {
     it('should validate uniqueness of group permission group name', async () => {
       const { adminUser } = await setupOrganization(nestApp);
 
-      const data = await service.create(adminUser, 'avengers');
-
-      expect(data.id).toBeDefined();
```
akshaysasidrn, line 32: Does this fail now?

server/src/helpers/utils.helper.ts
```diff
@@ -52,3 +53,13 @@ export function sanitizeInput(value: string) {
 export function lowercaseString(value: string) {
   return value?.toLowerCase();
 }
+
+export async function dbManagerWrapper(operation: (...args) => any, manager: EntityManager): Promise<any> {
```
akshaysasidrn, line 57: Can we reword this from `dbManagerWrapper` to `dbTransactionWrap` ?

Does refactoring this to the one below work?

```
export async function dbManagerWrapper(operation: (...args) => any, manager = getManager()): Promise<any> {
  await manager.transaction(async (manager) => {
    return await operation(manager);
  });
 }
```
The advantage being one wouldn't have to always pass the second param and have it wrapped in a transaction

akshaysasidrn:
Please check the failing e2e cases 🙇🏾

server/test/services/group_permissions.service.spec.ts
gsmithun4: for update request we are not expecting any response params in front end

gsmithun4: so returning empty response

server/src/helpers/utils.helper.ts
gsmithun4: Not working, if `manager` exists we have already started transactions for previous queries  `await manager.transaction` throws error

akshaysasidrn: Okay, thought it would result in nested transactions 🤔 
Ignore then 👍🏾

akshaysasidrn (approved):

PR #3812, opened 2022-08-17
[Bug]: Fixed Page Crash while Inspecting the widget with CSA

### LINKED ISSUE #3810: [Bug]: Page Crashes while Inspecting the widget with CSA
### Environment

Production

Should open the inspector when we select the widget

Page crashes when we open the widget inspector

1. Drop the Button & Text widget to the canvas
2. Click on the Button widget to open an inspector
3. Add events and set the action as a control component and select the text widget component -> Set text -> New Text
4. Delete the text widget
5. Select the Button widget and the page crashes

### Any possible solutions?

Check for an empty array in the function `getComponentActionOptions` & `getComponentActionOptions` while filtering the components will solve the issue

Yes

kavinvenkatachalam:
Resolves #3810

### DIFF
--- a/frontend/src/Editor/Inspector/EventManager.jsx
+++ b/frontend/src/Editor/Inspector/EventManager.jsx
@@ -108,7 +108,9 @@ export const EventManager = ({
 
   function getComponentActionOptions(componentId) {
     if (componentId == undefined) return [];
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-2qf8ppv Page Crashes while Inspecting the widget with CSA](https://app.clickup.com/t/2qf8ppv)

sherfin94 (approved):

PR #3853, opened 2022-08-24
[docs]: Provide border to the images

shivangT:
-  Replaced "class" with "classname" to provide border to an image
- Update this for Introduction, Setup, Security, Tracking, Tutorials, Datasource reference


withshubh (changes requested):
- We replaced the markdown with the `` tag but it is still wrapped under the `<div>` which is not required. That needs to be removed.
- The alt text is missing from all the `` tags. It's good SEO - please check this @Navaneeth-pk if it's required then we will have to add it.

Navaneeth-pk:
@withshubh yes, alt attribute is required.

shivangT:
Hey @withshubh,
-The <div> containers are required when the images are not of full width. I've removed the unnecessary div but you will find it in a few places because those images need to be central aligned (removing divs makes it left aligned)
-Reverted the alt text for all  tags

withshubh (approved):

PR #3881, opened 2022-08-26
Feature: Reverse proxy support

### LINKED ISSUE #983: Add base URL setting for better reverse proxy support

akshaysasidrn:
Fixes #983

### DIFF
--- a/frontend/src/_helpers/appUtils.js
+++ b/frontend/src/_helpers/appUtils.js
@@ -8,7 +8,7 @@ import {
   computeComponentName,
 } from '@/_helpers/utils';
 import { dataqueryService } from '@/_services';
-import _ from 'lodash';
+import _, { isEmpty } from 'lodash';
 import moment from 'moment';
 import Tooltip from 'react-bootstrap/Tooltip';
 import { componentTypes } from '@/Editor/WidgetManager/components';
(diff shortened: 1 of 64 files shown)


sherfin94:
Dashboard app icons seems doesn't seem to be working for me: 
image

sherfin94:
In the editor also there seems to be problems when I try running on local: 
image

image

frontend/src/_helpers/appUtils.js
```diff
@@ -901,6 +901,14 @@ export const getSvgIcon = (key, height = 50, width = 50) => {
   return <Icon style={{ height, width }} />;
 };
 
+export const assetPath = (path) => {
+  if (isEmpty(__webpack_public_path__)) return path;
+
+  if (__webpack_public_path__.at[-1] !== '/') return `${__webpack_public_path__}${path}`;
```
sherfin94, line 907:
```suggestion
  if (__webpack_public_path__.at(-1) !== '/') return `${__webpack_public_path__}${path}`;
```
@akshaysasidrn , this fixes the problems on my local that I reported with screenshots earlier

sherfin94 (changes requested):

PR #3888, opened 2022-08-26
[feature] Hide apps in dashboard

### LINKED ISSUE #3887: [feature] Add an option to hide app from dashboard

Add an option to hide the app from the dashboard 

Add an action in the apps permission tab to hide a particular app from the dashboard for specific users

Yes

shah21:
Resolves #3887

### DIFF
--- a/server/migrations/1661448846469-AddisHiddenFromDashboardToAppGroupPermissions.ts
+++ b/server/migrations/1661448846469-AddisHiddenFromDashboardToAppGroupPermissions.ts
@@ -1,0 +1,19 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+export class AddisHiddenFromDashboardToAppGroupPermissions1661448846469 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn(
+      'app_group_permissions',
+      new TableColumn({
+        name: 'is_hidden_from_dashboard',
+        type: 'boolean',
+        default: false,
+        isNullable: false,
+      })
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.dropColumn('app_group_permissions', 'is_hidden_from_dashboard');
+  }
+}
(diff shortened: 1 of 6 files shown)


server/migrations/1661448846469-AddisHiddenFromDashboardToAppGroupPermissions.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+export class AddisHiddenFromDashboardToAppGroupPermissions1661448846469 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.addColumn(
+      'app_group_permissions',
+      new TableColumn({
+        name: 'is_hidden_from_dashboard',
```
akshaysasidrn, line 8: Can we rename this in terms of an action?

Can user perform **x** action on the app in this group?
Where **x** currently are: _read, update, delete_
Can suggest it be _read_on_dashboard_

Alternatively, could we make do by just hiding apps on the dashboard for which user doesn't have update permission? 🤔

shah21: @gsmithun4 what do you think, instead of adding a checkbox with `hide app from the dashboard` label. just check if the user has permission to edit the app. if then show the app in the dashboard.
cc: @shubh22

shubh22: @akshaysasidrn @shah21 - There are three cases - 
1. Users can edit the app (Edit Access)
2. Users can just view the app (View Access)
3. User can view the app but should not see it in the dashboard. - This is the case where the app is linked to another app and doesn't have any usability alone. (View Access  + Hide from Dashboard)

So we need to create a new option "Hide From Dashboard". It will be visible only when the admin chooses the "View" Permission. It will be 'Unchecked' by default but if the admin 'checks' it, then users with view access won't see the app in the dashboard but will be able to access the app if given a URL. Users with edit access will still see the app in the dashboard.

akshaysasidrn: @shubh22 Thanks for clarifying. Understood now.

gsmithun4 (approved):
LGTM

emidhun:
@gsmithun4 Tested and verified on the deployed environment.
cc: @shah21

PR #3900, opened 2022-08-29
Bug Fixed : Overflow dropdown in the columns section inside table widget inspector  is not consistent with the dark theme

### LINKED ISSUE #3899: [Bug] overflow dropdown in columns section in Table widget inspector is not consistent with dark theme
### Version Information
v1.23.1

### Environment
Development

The **Overflow** dropdown in the columns section inside the table widget inspector should be consistent with the dark theme.

The **Overflow** dropdown in the columns section inside the table widget inspector is not consistent with the dark theme.

2. Switch to the dark theme by clicking on the half moon logo present at the bottom of the left sidebar (if the app is not in a dark 
   theme by default, if it is, you can skip step 2)
   
Screenshot 2022-08-29 at 6 43 19 PM

3. Drag the **Table** widget from the right sidebar on the canvas.
4. Open widget inspector for table widget by clicking on the label of dragged table widget on the canvas. (widget inspector will 
    be seen in the right sidebar,)
5. Inside the Properties tab in the widget inspector, look for the **Columns** section. You will find (x)
    number of buttons representing columns in the Table widget respectively.
Screenshot 2022-08-29 at 6 37 52 PM

6. Click any one of those buttons (for example - **id**) and a pop-over window will appear
7. In the pop-over window, look for the **Overflow** property (3rd property from the top), Click on its input select element, and 
   a dropdown will appear, and you can observe that dropdown is not consistent with the dark theme, it has white background 
   color in the dark theme, we need to change to the dark theme color.
  (To understand how it should be checked for the **Column type** dropdown, and see how it is consistent with the dark theme).

### Screenshots or Screencast
[Tooljet - Dashboard - 29 August 2022 - Watch Video](https://www.loom.com/share/40dcd7779b564b21a9ba6498ae34de38)

### Any possible solutions?
Add class `'select-search-dark'` if the dark mode is true to selectSearch component for the Overflow section

frontend>src>Editor>Components>Table>Table.jsx 

Yes

manishkushare:
Resolves #3899

### DIFF
--- a/frontend/src/Editor/Inspector/Components/Table.jsx
+++ b/frontend/src/Editor/Inspector/Components/Table.jsx
@@ -206,6 +206,7 @@ class Table extends React.Component {
             <div className="field mb-2">
               <label className="form-label">Overflow</label>
               <SelectSearch
+                className={`${this.props.darkMode ? 'select-search-dark' : 'select-search'}`}
                 options={[
                   { name: 'Wrap', value: 'wrap' },
                   { name: 'Scroll', value: 'scroll' },


shubh22:
Task linked: [CU-2rhjnn8 table widget dark theme](https://app.clickup.com/t/2rhjnn8)

kavinvenkatachalam (approved):

sherfin94 (approved):

Issue #3913, opened 2022-08-30
[BUG] REST API Oauth access token API url-encoded request not working


gsmithun4:
### Version Information
latest version

### Environment
all envs

### What is the expected behaviour?
REST API Oauth if token api is url encoded, it should be identified by setting `"Content-type:application/x-www-form-urlencoded"` to `Access Token URL custom headers` 

### What is the current behaviour?
Not working sending data in body

### If the bug is confirmed, would you be willing to submit a PR?
Yes

PR #3924, opened 2022-08-31
[BUG] REST API Oauth access token API url-encoded request not working

gsmithun4:
Resolves https://github.com/ToolJet/ToolJet/issues/3913


shah21 (commented):
LGTM
if you can please also test with the outlook oauth too

PR #3928, opened 2022-09-01
[ Bugfix ] :: calendar preview is broken when events are added as a result of any query

### LINKED ISSUE #3927: [BUG ]🐛  :: calendar preview is broken when events are added as a result of any query
calendar to be visible while previewing app

calendar is broken in app preview if event is added

1) Get event data as a result of any query in calendar
2) Run query on page load
2) Preview this app

Screen Recording 2022-08-31 at 12.56.16 PM.mov

stepinfwd:
closes #3927

### DIFF
--- a/frontend/src/Editor/Components/Calendar.jsx
+++ b/frontend/src/Editor/Components/Calendar.jsx
@@ -31,7 +31,10 @@ export const Calendar = function ({
   const style = { height };
   const resourcesParam = properties.resources?.length === 0 ? {} : { resources: properties.resources };
 
-  const events = properties.events ? properties.events.map((event) => prepareEvent(event, properties.dateFormat)) : [];
+  const events =
+    typeof properties?.events === Array
+      ? properties?.events?.map((event) => prepareEvent(event, properties.dateFormat))
+      : [];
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-2rx0yd1 calendar preview is broken when events are added as a result of any query](https://app.clickup.com/t/2rx0yd1)

frontend/src/Editor/Components/Calendar.jsx
```diff
@@ -30,8 +30,10 @@ export const Calendar = function ({
 }) {
   const style = { height };
   const resourcesParam = properties.resources?.length === 0 ? {} : { resources: properties.resources };
-
-  const events = properties.events ? properties.events.map((event) => prepareEvent(event, properties.dateFormat)) : [];
+  const events =
+    typeof properties?.events === Array
```
kavinvenkatachalam, line 34:
```suggestion
    Array.isArray(typeof properties?.events)
```

kavinvenkatachalam (changes requested):
@stepinfwd `typeof` in the array will always be `object`. To check if it is an array, you can use `Array.isArray()`

kavinvenkatachalam (approved):

PR #3936, opened 2022-09-02
[Feature]: Icon Widget

### LINKED ISSUE #3935: [Feature]: Icon Widget

Widget to add an icon will be a good addition to the list of existing widgets

Create a new Icon widget with the following properties and styles

Properties | Events | Styles | CSA
-- | -- | -- | --
Icon Picker | Click | Color | Click
Visibility | Hover | | Visibility

No

Yes

kavinvenkatachalam:
Resolves #3935


frontend/src/Editor/Components/Icon.jsx
```diff
@@ -0,0 +1,44 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [visibility]);
+
+  registerAction('click', async function () {
+    fireEvent('onClick');
+  });
+
+  registerAction('hide', async function (hide) {
+    setIconVisibility(typeof hide !== 'boolean' ? resolveWidgetFieldValue(hide, currentState) : hide);
+  });
```
gondar00, line 26: Wouldn't this get triggered on every render cycle?

frontend/src/Editor/Components/Icon.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <div className={`icon-widget`} style={{ display: hideIcon ? 'none' : '' }}>
```
gondar00, line 29: Would suggest using class-name instead of style property, 
```suggestion
    import cx from 'classnames';
    ...
    <div className={cx('icon-widget', { 'd-none': hideIcon })}>
```

frontend/src/Editor/Components/Icon.jsx
```diff
@@ -0,0 +1,44 @@
...
+      <IconElement
+        color={iconColor}
+        style={{ width: width, height: height }}
```
gondar00, line 32:
```suggestion
        style={{ width, height }}
```

frontend/src/Editor/Components/Icon.jsx
kavinvenkatachalam: Yes, it will be triggered on every render. We can move it `useEffect` with the dependencies. 
cc: @sherfin94

sherfin94: The second argument of `registerAction` takes in a dependency array that acts just like how useEffect works. We might need to revise the architecture of CSA in the near future though. But for the time being you can just pass any dependencies as the second argument of `registerAction` and it will re-register whenever the dependency gets updated.

sherfin94 (commented):
Made few code enhancement suggestions, component looks great!

sherfin94 (approved):

kavinvenkatachalam:
Task linked: [CU-2k5y167 Icon Widget](https://app.clickup.com/t/2k5y167)

shubh22:
Task linked: [CU-2k5y167 Icon Widget](https://app.clickup.com/t/2k5y167)

PR #3956, opened 2022-09-05
Bug fixed : canvas background colour is not consistent with the applied theme

### LINKED ISSUE #3955: Bug : Canvas background colour is not consistent with the dark theme
### Version Information
v1.24.0

### Environment
Development

The canvas background color in the editor should be consistent with the applied theme

The canvas background color in the editor is not consistent with the applied theme.

1. Suppose a new application is created in the light theme, even if we change the theme to a dark theme from the application, the background colour of the canvas will persist colour from the light theme only.
2. Suppose a new application is created with the dark theme, even if we change the theme to a light theme from the application, the background colour of the canvas will persist colour from the dark theme only.

Steps for the **light theme** by default
1. Let's assume, you are in light mode by default 
2. Now create a new application from the dashboard (In the dashboard you are in light theme by default)
3. Once the application is open, check the background colour of the canvas (You can also observe it, from the settings section in the left sidebar), it is "#edeff5" for the light theme by default.
4. Now, change to a dark theme, by clicking on the half moon icon at the bottom of the left sidebar. You can observe entire application is changed to a dark theme, except for the canvas background color, still persist the colour from the light theme as "#edeff5" instead of "#2f3c4c" (to confirm it, check the canvas background colour in the settings from the left sidebar again)

Steps for the **dark theme** by default
1. Let's assume, you are in dark mode by default 
2. Now create a new application from the dashboard (In the dashboard you are in dark theme by default)
3. Once the application is open, check the background colour of the canvas (You can also observe it, from the settings section in the left sidebar), it is "#2f3c4c" for the dark theme by default.
4. Now, change to a light theme, by clicking on the half moon icon at the bottom of the left sidebar. You can observe entire application is changed to a light theme, except for the canvas background color, still persist the colour from the dark theme as "#2f3c4c" instead of "#edeff5" (to confirm it, check the canvas background colour in the settings from the left sidebar again)

### Screenshots or Screencast
[Tooljet - Dashboard - 5 September 2022 - Watch Video](https://www.loom.com/share/7c59b9d060f643c6b67af33f18aea13e)

### Any possible solutions?
when toggling the theme, change the background colour  of the canvas accordingly 

frontend > src > Editor 

Yes

manishkushare:
Resolves #3955

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -1118,6 +1118,9 @@ class Editor extends React.Component {
       },
       showQuerySearchField: false,
     });
+    if (['#2f3c4c', '#edeff5'].includes(this.state.appDefinition.globalSettings.canvasBackgroundColor)) {
+      this.globalSettingsChanged('canvasBackgroundColor', newMode ? '#2f3c4c' : '#edeff5');
+    }
     this.props.switchDarkMode(newMode);
   };
 
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -1016,6 +1016,14 @@ class Editor extends React.Component {
     return canvasBoundingRect?.height;
   };
 
+  computeCanvasBackgroundColor = () => {
+    const { canvasBackgroundColor } = this.state.appDefinition?.globalSettings ?? '#edeff5';
+    if (['#2f3c4c', '#edeff5'].includes(canvasBackgroundColor)) {
+      return this.props.darkMode ? '#2f3c4c' : '#edeff5';
+    }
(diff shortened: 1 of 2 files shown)


shubh22:
Task linked: [CU-2t2zcnc canvas background is not consistent with the applied theme](https://app.clickup.com/t/2t2zcnc)

kavinvenkatachalam (changes requested):
@manishkushare Noticed a few issues

- In the viewer mode, the canvas is not consistent. It seems like it appears to be light even in the dark mode
- Canvas is not changing to dark in the case -> Open dashboard -> Change to light mode -> Edit app -> Switch to dark mode

kavinvenkatachalam (approved):

PR #3970, opened 2022-09-07
[Plugins] Adds support for pagination in listing objects from  AWS S3 buckets

### LINKED ISSUE #3971: Support to paginate Object list from AWS  S3 buckets
Currently while listing S3 bucket's objects from a particular directory or a sub-directory, default the action returns up to 1,000 key names. The response might contain fewer keys but will never contain more.
 
**Proposal**
Able to set the number of key names to be returned and allow to set offset for pagination.

arpitnath:
Resolves #3971


gsmithun4:
deployed https://tj-3921.herokuapp.com/

gsmithun4 (approved):
LGTM

akshaysasidrn (approved):

PR #4024, opened 2022-09-12
Enhancement : [ Button Widget ] 2.0

### LINKED ISSUE #4023: Enhancement : [ Button widget] refactor the widget
### Environment 
Development

### Proposed improvements 
1. There is only a click event on the button, we can try to add more events.
2. No way to handle disable, visibility, and loading of button widget by other widgets using component-specific action.
3. Move the location of states (loading, visible, disable) from the styles tab to the properties tab in the widget inspector and make FX first
4. Keep general, layout menus open by default in widget inspector

1. Add onHover event which will be triggered when we move our mouse cursor on the button widget and with that we can perform desired actions.

2. Add more component-specific actions such as disable, visible, and loading.
disable - the button widget should be disabled if set to true
visible - the button widget should be visible if it is set to true else hidden. 
loading - loader should be visible in  button widget if set true

3. We can move visibility and disabled props from the style tab to the properties tab, in the table widget inspector
Steps to reproduce the above issue
- Drag the Button widget on the canvas
- Open the Button widget inspector by clicking on the label of the button widget dragged on canvas
- Widget inspector for the button will appear on the right sidebar. 
- Go into the styles tab and look for visibility  and disabled menus, we have to move them under the properties tab

4. Keep the general and layout menus open in the button widget inspector, by setting isOpen to true for General and layout titles in DefaultComponent.jsx

Yes

manishkushare:
Resolves #4023


frontend/src/Editor/Components/Button.jsx
```diff
@@ -28,18 +42,34 @@ export const Button = function Button({ height, properties, styles, fireEvent, r
...
         className={cx('jet-button btn btn-primary p-1 overflow-hidden', {
-          'btn-loading': loadingState,
+          'btn-loading': loading,
         })}
         style={computedStyles}
         onClick={(event) => {
           event.stopPropagation();
           fireEvent('onClick');
         }}
+        onMouseMove={(event) => {
```
pateketrueke, line 69: Using `onMouseMove` will trigger on every mouse movement, right? I think `onHover` then is not the right event to trigger. Instead it would be `onMouseEnter` or `onMouseOver` may be?

shubh22:
Task linked: [CU-2ttrnnp Button widget - component 2.0](https://app.clickup.com/t/2ttrnnp)

frontend/src/Editor/Components/Button.jsx
```diff
@@ -1,45 +1,91 @@
...
+
+    registerAction('disable', async function (value) {
+      setDisable(value);
+    });
 
-  registerAction('setText', async function (text) {
-    setLabel(text);
-  });
+    registerAction('visibility', async function (value) {
+      console.log(typeof value, value, 'resolve', 'csa');
```
kavinvenkatachalam, line 64:
```suggestion
```

kavinvenkatachalam (changes requested):
@manishkushare Few suggestions,
1. Can you check the CSA? Seems like its not working in the editor
2. Remove unwanted console.log
3. Instead of using `resolveWidgetFieldValue` we need to replace it with `styles.visibility`. You can refer - https://github.com/ToolJet/ToolJet/pull/3936#discussion_r977502871

kavinvenkatachalam (approved):

PR #4036, opened 2022-09-14
Enhancement : [ text input widget ] refactor

### LINKED ISSUE #4035: Enhancement : [ Text input widget] refactor
### Environment 
Development

### Proposed improvements 

1. Move the location of states (loading, hidden, disable) and make FX first from the styles tab to the properties tab, in a widget inspector.
 - In the current version, we don't have a loading state for the text input widget, and hidden and disable states are in the styles tab.
 - So, we can move them to the properties tab, by adding all 3 props inside the properties object for the text input widget in widgetConfig file

2. Add new event - onFocus, unFocus (Blur)
 - In the current version, we only have bonded two events as onChange and onEnter
 - We can bind two more events, like onFocus and unFocus
 - onFocus will make the cursor visible in the widget, and perform the applied operation when the mouse enters the widget
 - unFocus (Blur) will make the cursor fade away and again perform a specific operation when the mouse leaves the widget

3. Add new Component specific actions
 - In the current deployment, the text input widget has support for only, setText and clearText functions.
 - we can add more such actions like setFocus, setBlur,  disable, loading, and hide, which will perform semantic operations when called upon.
 
4. add text color prop in the styles tab in the widget inspector
- We need to add support to change the color of the text inside the input field if users require it.

Yes

manishkushare:
Resolves #4035


shubh22:
Task linked: [CU-2u58p8u Text input widget enhancement](https://app.clickup.com/t/2u58p8u)

frontend/src/Editor/Components/TextInput.jsx
```diff
@@ -32,11 +41,23 @@ export const TextInput = function TextInput({
     setValue('');
     setExposedVariable('value', '').then(fireEvent('onChange'));
   });
+  registerAction('setFocus', async function () {
+    textInputRef.current.focus();
+  });
+  registerAction('setBlur', async function () {
+    textInputRef.current.blur();
+  });
```
kavinvenkatachalam, line 49:
```suggestion

  useEffect(() => {
    registerAction('setFocus', async function () {
      textInputRef.current.focus();
    });
    registerAction('setBlur', async function () {
      textInputRef.current.blur();
    });
  }, [textInputRef.current]);
```

frontend/src/Editor/Components/TextInput.jsx
```diff
@@ -32,11 +41,23 @@ export const TextInput = function TextInput({
...
+  registerAction('disable', async function (value) {
+    setDisable(value);
+  });
+  registerAction('visibility', async function (value) {
+    setVisibility(value);
+  });
```
kavinvenkatachalam, line 55:
```suggestion

  useEffect(() => {
    registerAction('setText', async function (text) {
      setValue(text);
      setExposedVariable('value', text).then(fireEvent('onChange'));
    });
    registerAction('clear', async function () {
      setValue('');
      setExposedVariable('value', '').then(fireEvent('onChange'));
    });
    registerAction('disable', async function (value) {
      setDisable(value);
    });
    registerAction('visibility', async function (value) {
      setVisibility(value);
    });
  }, []);
```

kavinvenkatachalam (changes requested):
@manishkushare Few suggestions,

1. Move all the `useState` above and `useEffect` below which will make better readability of the code
2. Move all the register actions inside `useEffect` with corresponding dependencies

kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

PR #4052, opened 2022-09-16
[Feature] Now multiple users can login through RestAPI-Oauth 🚀 

### LINKED ISSUE #4051: [feature] Multiple users should be able to log in using restapi-oauth authentication

Currently, the restapi plugin can only use single-user OAuth login information. So even if another user tries to run that query it will use the existing token and load prior user data 

Change the plugin functionality to store and use multiple tokens according to different users.

Yes

shah21:
Resolves #4051

### DIFF
--- a/server/src/services/data_queries.service.ts
+++ b/server/src/services/data_queries.service.ts
@@ -106,7 +106,7 @@ export class DataQueriesService {
     let result;
 
     try {
-      return await service.run(sourceOptions, parsedQueryOptions, dataSource.id, dataSource.updatedAt);
+      return await service.run(sourceOptions, parsedQueryOptions, dataSource.id, dataSource.updatedAt, { user });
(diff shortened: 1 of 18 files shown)


server/src/services/data_queries.service.ts
```diff
@@ -106,18 +106,23 @@ export class DataQueriesService {
     let result;
 
     try {
-      return await service.run(sourceOptions, parsedQueryOptions, dataSource.id, dataSource.updatedAt);
+      return await service.run(sourceOptions, parsedQueryOptions, dataSource.id, dataSource.updatedAt, { user });
```
gsmithun4, line 109: user will have sensitive info like password, shall we pick and pass details which are required, better if we can create a type for user

gsmithun4 (changes requested):

PR #4057, opened 2022-09-16
[Bug]: Fixed localisation for subpath

### LINKED ISSUE #4053: i18n.js is not using the SUB_PATH variable, if any defined (used in reverse-proxy setups)
### Version Information

`:latest` [tag](https://hub.docker.com/layers/tooljet/tooljet-ce/latest/images/sha256-3e70e97672cb0e9f702a3311e2fd9c3bdd643f4f48bd53cef00fd2dc8516e889?context=explore).

### Environment
Docker `tooljet/tooljet-ce:latest` behind Caddy.

i18n.js should use `SUB_PATH`  (process.env.SUB_PATH)

It looks like following files is impacted:
https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/i18n.js

Insted of just the hardcoded string:
https://github.com/ToolJet/ToolJet/blob/e8ae8a9fe64eb1c7783fbc5ae91655671a007b35/frontend/src/i18n.js#L16

It's just using `/assets/translations/{{lng}}.json`, so the lang-files can't be resolved if the user is trying to visit the system from a reverse-proxy solution.

1. Setup ToolJet
2. Use `SUB_PATH` as documented in the docs
3. Try and open up a page in the front-end

### Screenshots or Screencast

image

### Any possible solutions?

Use `process.env.SUB_PATH` like it's implemented in other logics. I can see that `maybeSetSubPath` would be useful.

Here: https://github.com/ToolJet/ToolJet/blob/e8ae8a9fe64eb1c7783fbc5ae91655671a007b35/frontend/src/i18n.js#L16

I guess so... ;)

kavinvenkatachalam:
Resolves #4053


gondar00 (approved):

exetico:
Looks good.

Also, this one needs some care-taking, too:
https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/_components/PrivateRoute.jsx#L16

(Currently `SUB_PATH` is not in use)

gondar00 (approved):

PR #4058, opened 2022-09-16
Enhancement : [ Date range picker widget] bind onSelect event

### LINKED ISSUE #4026: Add Event trigger to RangePicker

Currently, there is no way to trigger an event if the dates are changed on the RangePicker widget. That means that if we want to change the dates on the dashboard and reload the charts, we have to use two DatePickers.

RangePicker configuration pane should have "Events" section, just like DatePicker pane.

Currently using two separate DatePickers, but it's not ideal.

manishkushare:
Resolves #4026

### DIFF
--- a/frontend/src/Editor/Components/DaterangePicker.jsx
+++ b/frontend/src/Editor/Components/DaterangePicker.jsx
@@ -12,6 +12,7 @@ export const DaterangePicker = function DaterangePicker({
   setExposedVariable,
   width,
   darkMode,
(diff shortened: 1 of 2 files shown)


shubh22:
Task linked: [CU-2uauyr3 Date range picker - bind onSelect event listner](https://app.clickup.com/t/2uauyr3)

kavinvenkatachalam (approved):

PR #4073, opened 2022-09-20
[Bugfix] Calendar widget : Default date is not updated dynamically and widget crashes for invalid dates

### LINKED ISSUE #4061: [Calendar]: Default date is not updated dynamically

Changing the value from the editor, it must update the date in the widget.

The "Default date" property is used to set a date (at day level).  it doesn’t  change if we update it from the editor. Currently, it only changes when we reload the page.

1. Visit the [link](https://app.tooljet.com/) and create a new application.
2. Drag and drop the calendar widget from the widgets inventory on the left
3. inspect the widget to access it's "Default date" property
4. Now, try to update the default date property
5. Either the date in the widget doesn’t update or the UI breaks when we reload the page when default date is left empty.

arpitnath:
Resolves #4061


Nandinisaha13:
<html>
<body>

Test Case | Steps | Expected Result
-- | -- | --
Verify that calendar not crashing while the default date is empty. | Drag & drop calendar. | Calendar should be displayed on canvas.
  | Click on calendar. | Calendar should be selected & properties should be displayed.
  | Empty the default date & refresh the app. | The default date should be empty & widget does not crash.
Verify that calendar displays correct date while default date is changed. | Drag & drop calendar. | Calendar should be displayed on canvas.
  | Click on calendar. | Calendar should be selected & properties should be displayed.
  | Change date in default date. | The date should be displayed on calendar.
Verify that calendar not crashing while the default date is invalid. | Drag & drop calendar. | Calendar should be displayed on canvas.
  | Click on calendar. | Calendar should be selected & properties should be displayed.
  | Enter invalid default date & refresh the app. | The default date should be invalid & widget does not crash.


</body>
</html>

sherfin94 (changes requested):
Hey @arpitnath ,

The `Today`, `Next` and `Back` button on calendar does not work now.

To make it work, my suggestion is the following implementation:

1. Parse and store the 'Default date' property in an internal state variable of the calendar component, as a moment object.
2.  Use this state variable to generate the date object required for the `date` property of ReactCalender
3. Whenever the value of that property changes, update the state variable also.
4. Whenever the user clicks on `Next`/`Back`/`Today` button, recompute this state variable accordingly:
    - Add or subtract based on whether they clicked on `Next` or `Back`
    - Add/sub 1 day if mode is day, 7 days if mode is week, and 1 month if mode is month
    - If they clicked on today, just update it to today's date.

sherfin94 (changes requested):
Hey @arpitnath , I couldn't find any issue in your code in first glance, but the calendar navigation doesn't seem to function properly, please check:


https://user-images.githubusercontent.com/20333972/191683844-2273d726-0113-4b79-9f10-4a41f679de85.mov

PR #4080, opened 2022-09-21
[improvement] Table : adds number as a cell type

### LINKED ISSUE #4079: Feature request: Number as a cell type for Tables widgets
## Description
Currently, we do not support **Number** as a cell type for the Table widget.

### Proposed improvements
- option to choose **Number** as a cell type along with other cell types for the Tables widget
- On edit column, the input field should only expect **Numbers**
- Data can be sorted (in ascending and descending order)
- Validations: Minimum value and maximum value

arpitnath:
Resolves #4079


sherfin94 (changes requested):
Hey @arpitnath ,

I set the `id` field of default table to `number` type.
I set it to editable.
I changed every cell inside that column to new numbers.
But the exposed data inside `dataUpdates` and `changeSet` was still of `string` type instead of `number` type.

Please check.

image

image

image

PR #4087, opened 2022-09-21
[ Enhancement] : adding local storage support in query manager panel

### LINKED ISSUE #4086: [ Enhancement ] : Adding localStorage support to query panel to persist size and position of panel on hard reload
In the query panel component, whenever the user is  changing the size and position of the panel and opens the app in a new tab. All the changes get lost.

### How to reproduce the problem?
1 . Create a new app or, open an existing app.
2. In the editor, try to hide or open the query Manager, by clicking on either (^) up arrow or down arrow respectively.
3.  Also, try one more scenario, we can change the height of the query panel by moving the panel up or down using the mouse.
Once, we set it to a particular height, try reloading the page and you can observe, the height is not persisted, rather it used the default height.
4. And after making changes, open the same app, in a new tab, and see, if the app persists in the prev state. or just hard reload the existing tab itself.

### Screenshots or Screencast
[Tooljet - Dashboard - 22 September 2022 - Watch Video](https://www.loom.com/share/10f63136c3f94c32b623e77af3ab220e)

We can use localStorage to store the size and position of the query panel. So, that we can persist prev state on hard reloading.

Yes

manishkushare:
Resolves #4086


shubh22:
Task linked: [CU-2v1vhaq Adding local storage support to query panel](https://app.clickup.com/t/2v1vhaq)

frontend/src/Editor/QueryManager/QueryManager.jsx
```diff
@@ -46,7 +46,8 @@ class QueryManagerComponent extends React.Component {
     };
 
     this.previewPanelRef = React.createRef();
-    this.buttonConfig = JSON.parse(localStorage.getItem('queryManagerButtonConfig'));
+    this.queryManagerPreferences = JSON.parse(localStorage.getItem('queryManagerPreferences'));
+    this.buttonConfig = this.queryManagerPreferences?.buttonConfig ?? {};
```
kavinvenkatachalam, line 50:
```suggestion

    if (localStorage.getItem('queryManagerButtonConfig') === null) {
      this.buttonConfig = this.queryManagerPreferences?.buttonConfig ?? {};
    } else {
      this.buttonConfig = JSON.parse(localStorage.getItem('queryManagerButtonConfig'));
      localStorage.setItem(
        'queryManagerPreferences',
        JSON.stringify({ ...this.queryManagerPreferences, buttonConfig: this.buttonConfig })
      );
      localStorage.removeItem('queryManagerButtonConfig');
    }
```

kavinvenkatachalam (changes requested):
@manishkushare Added a small suggestion to support backward compatibility for the users who configured the button on the query panel previously.

kavinvenkatachalam (approved):

PR #4092, opened 2022-09-22
[ Improvement ] :: Table  widget

### LINKED ISSUE #1316: Stop pageination

Currently you can click on the next page even if there is no more data.

Currently there are Visibility and Disable. Something like Stop Pagination would be useful with a simple boolean flag. Going back should still work of course. Just not further forward.

stepinfwd:
* added option to disable server side pagination arrows
* server side total records number
resoves #4132 
resolve #1316


sherfin94 (changes requested):
Hey @stepinfwd ,

The Total records functionality looks fine

The other feature related to pagination arrows in server side pagination seems to have been miscommunicated. Please consider:

1. `<` and `>` buttons need separate boolean properties to enable and disable, and NOT a single property.
2. Enable and disable here means, active and inactive, NOT visible and invisible. So these two properties can be named 'Next page button active' and 'Previous page button active'
3. If these two properties are not supplied, fallback to `true`

To add a little context, in server side pagination, whether these two pages are available oftentimes depend on the response from server and therefore need to be configured dynamically by the developer.

shubh22:
Task linked: [CU-2vjy6zz enableNextButton (server side) , BackButton (server side) -Table](https://app.clickup.com/t/2vjy6zz)

shubh22:
Task linked: [CU-2vjy6z9 Option for total Records (server side) - Table widget](https://app.clickup.com/t/2vjy6z9)

sherfin94 (changes requested):
Please look into these:

* Next button and Prev button should be enabled by default for backward compatibility (I think this was miscommunicated by my last comment)
* Rename 'Enable prev button' and 'Enable next button' to 'Enable next page button' and 'Enable previous page button'
*

sherfin94 (approved):

PR #4098, opened 2022-09-22
[ hotfix ] : [ Table Widget ]  filter pop-up and dropdown inside it are  not scrollable

### LINKED ISSUE #4097: [ hotfix ] :  [ Table Widget ] Filter container and dropdown inside it are not scrollable
### Version Information
v1.25.5

### Environment
All

In the Table widget, the filter container should be scrollable and the dropdown inside the same container should be scrollable as well.

The filter container is not scrollable hence if we have lists of filters applied, it won't be possible to see all, as overflow is hidden.
The same goes with the dropdowns inside the filter container, they are not scrollable, so we cannot scroll down and select other options which are not visible to us since overflow is hidden.

1. Create a new application or open an existing one. You will be redirected to the editor
2. Now, look for the Table widget in the right sidebar and drag it on the canvas.
3. Look for the filter button on the top right side of the widget. Click on it, to open the filter pop-up
4. In the filter pop-up, click on the **+add filter** button present at the bottom of the pop-up.
5. A row will appear at the main body of the pop-up to configure the filter according to our requirements. ( Repeat step 4, multiple times, so that we can check, if all the added filters are visible/accessible to us or not. )
At one point, when there are multiple filters, and they go out of flow from the parent container height, that's our first matter of concern. We need to avoid this and make the container scrollable so that we can add  "n" number of filters and access them by scrolling the parent container.
6. Now let's focus on another matter of concern, let's try to configure any one of the filter rows that we added in the above step.
7. To do so, let's consider configuring a second input select field from left to right, click on it, and a dropdown will appear, now try scrolling the dropdown, it won't and thus we cannot access the options which are hidden due to overflow. That's our second matter of concern, we need to make it overflow so, that we can scroll down and select any desired option from the dropdown.

### Screenshots or Screencast
[Tooljet - Dashboard - 23 September 2022 - Watch Video](https://www.loom.com/share/0685f75e21c945a8b664ab26be040be8)

### Any possible solutions?
1. Make the filter container scrollable
2. Replace the react-search-select library with custom **selectComponent.jsx** component available in the code base

frontend > src > Editor >  Components > Table > Table.jsx 

Yes

manishkushare:
Resolves #4097


shubh22:
Task linked: [CU-2v7en4a Hotfix : filter section is not scrollable](https://app.clickup.com/t/2v7en4a)

sherfin94 (commented):
Hey @manishkushare , PR works well!, please resolve the conflicts and then we can merge it.

kavinvenkatachalam (approved):
LGTM

frontend/src/Editor/Components/Table/Filter.jsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import SelectSearch, { fuzzySearch } from 'react-select-search';
```
kavinvenkatachalam, line 2: @manishkushare Can you remove `SelectSearch` if it is not used anywhere?

kavinvenkatachalam (changes requested):
@manishkushare Suggested a minor change

PR #4100, opened 2022-09-23
fix: pass-down dataQueries to allow runQuery on modals, closes #4099

### LINKED ISSUE #4099: [bug] runQuery does not work if CustomComponent is within a modal
### Version Information

v1.18.0 - v1.25.6

### Environment

MacOS / Linux / Docker (I tested all of them, but is not an infra issue)

Custom component should be able to run queries.

I does not, if the component is in a modal, queries won't run.

1. Add a query with js-code "run" that just `alert('WAT')`
2. Add custom-component on the page, with:

```jsx
import React from 'https://cdn.skypack.dev/react';
import ReactDOM from 'https://cdn.skypack.dev/react-dom';

const RunQuery = ({runQuery}) => (
  <>
    <button onClick={() => {runQuery('run')}}>run query</button>
  </>
);
const ConnectedComponent = Tooljet.connectComponent(RunQuery);
ReactDOM.render(<ConnectedComponent />, document.body);
```
3. Click on the "run query" button, it would `alert('WAT')` as expected
4. Add a modal, open it and place a custom-component with the same code
5. Click on the "run query" and nothing will happen
6. Add a button and run this query with onClick, it will work

### Screenshots or Screencast

Not needed, the behavior remains the same, no matter what.

I got this on the console:

```
CustomComponent.jsx (52:8) TypeError: Cannot read properties of undefined (reading 'filter')
    at CustomComponent.jsx:44:56
```
### Any possible solutions?

It seemed to be just a missing prop, PR is already created!

Yeah, I did, the `SubContainer.jsx` component was not passing down the `dataQueries` and that's it.

It's on the way! :beers:

pateketrueke:
fixes #4099

### DIFF
--- a/frontend/src/Editor/SubContainer.jsx
+++ b/frontend/src/Editor/SubContainer.jsx
@@ -33,6 +33,7 @@ export const SubContainer = ({
   darkMode,
   containerCanvasWidth,
   readOnly,
+  dataQueries,
   customResolvables,
   parentComponent,
   onComponentHover,
@@ -422,6 +423,7 @@ export const SubContainer = ({
             onComponentOptionChanged={onComponentOptionChangedForSubcontainer}
             onComponentOptionsChanged={onComponentOptionsChanged}
             key={key}
+            dataQueries={dataQueries}
             currentState={currentState}
             onResizeStop={onResizeStop}
             onDragStop={onDragStop}


joshuamarquez:
+1

withshubh:
Thanks for raising the PR @pateketrueke 🎉 
We will get it reviewed soon.

sherfin94 (approved):

Issue #4093, opened 2022-09-22
Table enhacement


stepinfwd:
### Describe the solution you'd like

Implement following improvements 

-  disable sorting - option to enable/disable
- hover row:  add hoveredRow event ( expose - howeveredRowId ) 
- on changes discarded - add event 
- selectedRow - Add CSA

### If the feature is approved, would you be willing to submit a PR?
yes

PR #4104, opened 2022-09-23
Enhancement :: table events :: onHover

stepinfwd:
* onhover event added
part of #4093

### DIFF
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -49,6 +49,7 @@ export function Table({
   properties,
   variablesExposedForPreview,
   exposeToCodeHinter,
+  events,
 }) {
   const {
     color,
@@ -76,7 +77,7 @@ export function Table({
   const { t } = useTranslation();
 
   const [tableDetails, dispatch] = useReducer(reducer, initialState());
-
+  const [hoverAdded, setHoverAdded] = useState(false);
   const mergeToTableDetails = (payload) => dispatch(reducerActions.mergeToTableDetails(payload));
   const mergeToFilterDetails = (payload) => dispatch(reducerActions.mergeToFilterDetails(payload));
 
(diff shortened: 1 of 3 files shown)


sherfin94 (changes requested):
Hey @stepinfwd ,

It seems to be possible to go though the row without getting the on hover event triggered, if we slowly go through the row without touching any other row. In the following video I have added a show alert handler to the hover event.

Try using `onMouseOver`.

shubh22:
Task linked: [CU-2vjy6vd Hover Row - add event on table widget](https://app.clickup.com/t/2vjy6vd)

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -364,18 +364,39 @@ export function Table({
...
     if (pageCount <= pageIndex) gotoPage(pageCount - 1);
   }, [pageCount]);
 
+  function usePrevious(value) {
+    const ref = useRef();
+    useEffect(() => {
+      ref.current = value;
+    });
+    return ref.current;
+  }
```
kavinvenkatachalam, line 378: This function seems to be called every time and the old value is set to ref every time it is called. Instead of this, can we use `ref` directly on the `onMouseOver`?

`onMouseOver={(e) => {
                      const hoveredRowDetails = { hoveredRowId: row.id, hoveredRow: row.original };
                      ref.current = rowDetails?.hoveredRowId;
                      setRowDetails(hoveredRowDetails);
                    }}`

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -487,6 +508,13 @@ export function Table({
                         fireEvent('onRowClicked');
                       });
                     }}
+                    onMouseOver={(e) => {
+                      const hoveredRowDetails = { hoveredRowId: row.id, hoveredRow: row.original };
+                      setRowDetails(hoveredRowDetails);
+                    }}
+                    onMouseLeave={(e) => {
+                      setRowDetails({ hoveredRowId: '', hoveredRow: '' });
+                    }}
```
kavinvenkatachalam, line 517: We're setting the state every time when we hover and leave the row irrespective of the `onRowHovered` event on the table row. This operation seems expensive as it will re-render the Table component. Is it possible to restrict and trigger only when we set the `onRowHovered` event?
cc: @sherfin94

kavinvenkatachalam (approved):

PR #4105, opened 2022-09-23
Bugfix: update both comments collections when socket message is received

### LINKED ISSUE #3044: Resolved comments should be displayed on the comments card without page reloading or executing other actions.

The resolved comments should show on the comments card without page reloading or executing some other actions.

The resolved comments are shown on the comments card only after the page reloads or executes some other actions.

1.
2.
3.

### Screenshots or Screencast
[Video](https://www.loom.com/share/02f4f4c1cba2401e8f1a95df68282de1)

mslourens:
This PR fixes #3044 
I split both notification collections into resolved and unresolved, because setting the collection based on the state of the tab and using the same collection for both tabs feels weird. 
Also whenever a notifications message is received, it is not possible to determine whether an unresolved comment is marked as resolved or visa versa. Therefore I suggest we update both collections whenever a message is received.

### DIFF
--- a/frontend/src/Editor/CommentNotifications/index.jsx
+++ b/frontend/src/Editor/CommentNotifications/index.jsx
@@ -11,18 +11,28 @@ import TabContent from './Content';
 import useRouter from '@/_hooks/use-router';
 
 const CommentNotifications = ({ socket, toggleComments, appVersionsId }) => {
-  const [notifications, setNotifications] = React.useState([]);
+  const [unresolvedNotifications, setUnresolvedNotifications] = React.useState([]);
+  const [resolvedNotifications, setResolvedNotifications] = React.useState([]);
   const [loading, setLoading] = React.useState(false);
   const [key, setKey] = React.useState('active');
 
   const router = useRouter();
 
-  async function fetchData(k) {
-    const isResolved = k === 'resolved';
+  async function fetchData(k, both = false) {
     setLoading(true);
-    const { data } = await commentsService.getNotifications(router.query.id, isResolved, appVersionsId);
+    if (both) {
+      const [resolved, unresolved] = await Promise.all([
+        commentsService.getNotifications(router.query.id, true, appVersionsId),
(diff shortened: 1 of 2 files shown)


withshubh:
Thanks for your contribution @mslourens 🎉 
A team member will review the PR soon.

frontend/src/Editor/CommentNotifications/index.jsx
```diff
@@ -11,28 +11,39 @@ import TabContent from './Content';
...
-  const [notifications, setNotifications] = React.useState([]);
+  const [unresolvedNotifications, setUnresolvedNotifications] = React.useState([]);
+  const [resolvedNotifications, setResolvedNotifications] = React.useState([]);
   const [loading, setLoading] = React.useState(false);
   const [key, setKey] = React.useState('active');
 
   const router = useRouter();
 
-  async function fetchData(k) {
-    const isResolved = k === 'resolved';
```
gondar00, line 21: @mslourens I believe the issue exists when the user switches to the "resolved" tab, feel like the issue can be solved by adding another check for `isResolved` based on the tab `key` value and prevent making an extra api call i.e

```js
const isResolved = k === 'resolved' || key === 'resolved';
```

withshubh:
Hey @mslourens 
Please update the PR with the changes suggested above. 🙌

mslourens:
That's the whole issue here, for some reason the key is always 'acrive' when a notification is received. That's why I added the 
'both' argument to at least keep them both up to date. I am not very experienced with React, so I don't know why the key property is always active (even when the received tab is selected and the setKey hook is called with 'received'). This PR at least fixes the issue for now.

kavinvenkatachalam (changes requested):
@mslourens Thanks for the PR.

Can you change the base branch to `ToolJet:develop` and take a pull from latest develop?

tejakummarikuntla:
closing the PR due to inactivity

PR #4115, opened 2022-09-26
bug fixed : The result of the previous previewed query remains

### LINKED ISSUE #2827: The result of the previous previewed query remains
### Environment
MacOS 12.3.1

The preview pane shows the result of the previously previewed query

when the query gets deleted (because you clicked the plus button again, without saving the previous query), the preview panel should also be cleared

1. Create a REST api datasource with this endpoint: `https://reqres.in/`
2. Click the + button in the Queries panel to create a new query
3. Add this path to the datasource url: `api/users`
4. Click the preview button
5. Notice the results of the query in the preview panel
6. Click the + button again without saving the query
7. Notice the path input gets cleared, but the preview panel still shows the result of the previous query

### Screenshots or Screencast

https://user-images.githubusercontent.com/1907152/163372131-d7ca1ce5-e652-4560-ae3f-481a4f3558d4.mov

No

manishkushare:
Resolves #2827

### DIFF
--- a/frontend/src/Editor/QueryManager/QueryManager.jsx
+++ b/frontend/src/Editor/QueryManager/QueryManager.jsx
@@ -159,6 +159,9 @@ class QueryManagerComponent extends React.Component {
     //     }
     //   }
     // }
+    if (this.state.selectedQuery?.id !== nextProps.selectedQuery?.id) {
+      this.setState({ queryPreviewData: undefined });
+    }
     if (this.props.showQueryConfirmation && !nextProps.showQueryConfirmation) {
       if (this.state.isUpdating) {
         this.setState({
@@ -413,7 +416,15 @@ class QueryManagerComponent extends React.Component {
       queryPreviewData,
       dataSourceMeta,
     } = this.state;
-
+    console.log(
+      'selectedDataSource',
+      selectedDataSource,
+      'selectedQuery',
+      selectedQuery,
+      'queryPreviewData',
+      queryPreviewData,
+      'query'
+    );
     let ElementToRender = '';
 
     if (selectedDataSource) {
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-2vd8ffd The result of the previous previewed query remains](https://app.clickup.com/t/2vd8ffd)

frontend/src/Editor/QueryManager/QueryManager.jsx
```diff
@@ -413,7 +416,15 @@ class QueryManagerComponent extends React.Component {
...
-
+    console.log(
+      'selectedDataSource',
+      selectedDataSource,
+      'selectedQuery',
+      selectedQuery,
+      'queryPreviewData',
+      queryPreviewData,
+      'query'
+    );
```
kavinvenkatachalam, line 427:
```suggestion
```

frontend/src/Editor/QueryManager/QueryManager.jsx
```diff
@@ -159,6 +159,9 @@ class QueryManagerComponent extends React.Component {
     //     }
     //   }
     // }
+    if (this.state.selectedQuery?.id !== nextProps.selectedQuery?.id) {
+      this.setState({ queryPreviewData: undefined });
+    }
```
kavinvenkatachalam, line 164:
```suggestion
```

Instead of setting state in `componentWillReceiveProps`, set it in `setStateFromProps` where we set all the other states. It will avoid the one cycle of re-render. 

`queryPreviewData:
          this.state.selectedQuery?.id !== props.selectedQuery?.id ? undefined : this.state.queryPreviewData,`

kavinvenkatachalam (changes requested):
@manishkushare Added a few suggestions.

kavinvenkatachalam (approved):

PR #4118, opened 2022-09-26
fixes: query params are passed as slug to the app url for go-to-app action

### LINKED ISSUE #4121: Query params are not passed as slug  for go-to-app action from editor
Query params are passed as a slug for go-to-app action from the editor and viewer.
On triggering `go-to-app` action with query params, opens the Tooljet app without the params as a slug.

1. Drop any widget from the widget manager (a table for example)
2. Add an event, On row selected, with action `go-to-app` and the name of the app and query params
3. Select an row, which will trigger the `go-to-action`

arpitnath:
Resolves #4121


akshaysasidrn (approved):

shubh22:
Task linked: [CU-2vd90g5 Go to app not working with query params passed](https://app.clickup.com/t/2vd90g5)

PR #4120, opened 2022-09-26
Add setState functions of components as dependencies for registerAction

### LINKED ISSUE #4119: Component specific actions are not working properly inside modal
### Environment

Enterprise deployment

Component specific actions for components inside a modal should work when the modal is closed and opened again.

Component specific actions for components inside a modal are not working when the modal is closed and opened again.

1. Add a modal
2. Add a text input and button inside the modal
3. Handle the Button's onClick event to set the text of text input
4. Close the modal
5. open the modal
6. click on the button
7. Observe that text did not get set

Yes

sherfin94:
Resolves #4119

The problem occurs due to the state setting functions of components getting reset when the components are re-rendered during opening and closing of modal.

The solution is to re-register each action whenever the state setting functions get changed, by adding them to the dependency array of `registerAction`.

### DIFF
--- a/frontend/src/Editor/Components/Button.jsx
+++ b/frontend/src/Editor/Components/Button.jsx
@@ -47,21 +47,37 @@ export const Button = function Button({
     fireEvent('onClick');
   });
 
-  registerAction('setText', async function (text) {
-    setLabel(text);
-  });
+  registerAction(
+    'setText',
+    async function (text) {
+      setLabel(text);
+    },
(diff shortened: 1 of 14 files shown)


kavinvenkatachalam (approved):
LGTM

shubh22:
Task linked: [CU-2vjuuve Component Specific Actions not working inside modal](https://app.clickup.com/t/2vjuuve)

PR #4128, opened 2022-09-27
[ Enhancement ] :: selectedRow CSA for table widget

stepinfwd:
part of #4093

### DIFF
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -330,14 +330,10 @@ export function Table({
     'selectRow',
     async function (id) {
       const item = currentState.components[component.name]['currentData'].filter((item) => item.id == id);
-      let rowId = '';
-      let original = '';
-      page.map((item, index) => {
-        if (item.original.id == id) {
-          rowId = item.id;
-          original = item.original;
-        }
-      });
+      const row = page.find((item, index) => item.original.id == id);
+      const rowId = row.id;
+      const original = row.original;
+
       const selectedRowDetails = { selectedRow: item[0], selectedRowId: rowId };
       mergeToTableDetails(selectedRowDetails);
       setExposedVariables(selectedRowDetails).then(() => {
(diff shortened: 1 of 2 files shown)


shubh22:
Task linked: [CU-2vjw22z Table Enhancements :: add selectedrow CSA](https://app.clickup.com/t/2vjw22z)

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -326,6 +326,27 @@ export function Table({
...
+      const item = currentState.components[component.name]['currentData'].filter((item) => item.id == id);
+      let rowId = '';
+      let original = '';
+      page.map((item, index) => {
+        if (item.original.id == id) {
+          rowId = item.id;
+          original = item.original;
+        }
+      });
+      const selectedRowDetails = { selectedRow: item[0], selectedRowId: rowId };
```
kavinvenkatachalam, line 341:
```suggestion
const row = page.find((item, index) => item.original.id == id);
      const selectedRowDetails = { selectedRow: item[0], selectedRowId: row?.id ?? '' };
```

kavinvenkatachalam (changes requested):
@stepinfwd Made a suggestion below. Instead of `map` which loops all the element of the array we can use `find` - it stops when it finds the first element.

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -326,6 +326,23 @@ export function Table({
     [serverSidePagination, clientSidePagination]
   );
 
+  registerAction(
+    'selectRow',
+    async function (id) {
+      const item = currentState.components[component.name]['currentData'].filter((item) => item.id == id);
+      const row = page.find((item, index) => item.original.id == id);
+      const rowId = row.id;
+      const original = row.original;
```
kavinvenkatachalam, line 335: Can you remove the `original` variable if it is not needed? And we can use `row.id` directly in the object instead of creating a new variable

kavinvenkatachalam (changes requested):
@stepinfwd Changes look good. One suggestion,

If I pass the `id` which is not in the table, it throws an error. Can we add an `undefined` check to skip that?

kavinvenkatachalam (dismissed):
LGTM

stepinfwd:
closing this one  , opened new implementation here https://github.com/ToolJet/ToolJet/pull/4591

PR #4130, opened 2022-09-27
[Enhancement] : [Table widget] Adding filters

### LINKED ISSUE #4129: [ Enhancement ] : [ Table widget ] introducing more filter options
### Environment
Development

1. In the current deployment, In the Table widget, there are multiple options available to filter the data in the widget.
We can add more filters such as **is empty, is not empty, and does not contains.**
2. Also, contains seems not to be working for all the fields, so we need to add logic for the contains filter as well.

2. Drag the Table widget on the canvas from the right sidebar.
3. In the Table widget, there is an option to add filters to filter the data, by clicking on the filter icon at the top right side of the widget. On clicking, a pop-up will appear.
4. In the pop-up, if no filters are present, you can add a filter by clicking on the +add filter button at the bottom of the widget
5. Pop-up will now show the content to configure the filters. Click on the second dropdown, it will show all the filter methods. There we can add more filters as discussed in point 1 in **Is your proposal related to problem** section
6. Now, for the second point in the **Is your proposal related to problem** section, please follow the following steps
    - In the first dropdown (responsible to select the column which we want to target). Select the **id** option from the same dropdown
    - Now, from the second dropdown (responsible to apply filter method), select **contains** the option.
    -  Now, in the value enter numeric 1
    - This way we have configured our first filter, and according to the configuration, it should filter the data, whose id is 1, and only display that data in the Table. But, instead, it is showing all the data that is provided by default to Table. That's our second concern.
 
### Screenshots or Screencast
[Tooljet - Dashboard - 27 September 2022 - Watch Video](https://www.loom.com/share/6b79ee6d218a4c87aaf64f120ecc1277)

1. Adding filter options such as is empty, is not empty, does not contains.
8. Writing logic for contains filter

Yes

manishkushare:
Resolves #4129

### DIFF
--- a/frontend/src/Editor/Components/Table/Filter.jsx
+++ b/frontend/src/Editor/Components/Table/Filter.jsx
@@ -95,6 +95,7 @@ export function Filter(props) {
               <SelectSearch
                 options={[
                   { name: 'contains', value: 'contains' },
+                  { name: 'does not contains', value: 'doesNotContains' },
                   { name: 'matches', value: 'matches' },
                   { name: 'does not match', value: 'nl' },
                   { name: 'equals', value: 'equals' },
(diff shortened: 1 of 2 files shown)


kavinvenkatachalam (approved):
LGTM

PR #4134, opened 2022-09-27
[ Enhancement ] : [ Table widget ] added support to download data in excel

### LINKED ISSUE #4133: [ Enhancement ] : Adding  excel downloading option in Table widget
### Environment 
Development

In the current deployment, there is no way to download the data from the **Table widget** in  excel format.

We can add the option to  download data in excel using the xlsx package

2. Drag the table widget on the canvas from the right sidebar.
3. In the Table widget, look for the download option at the extreme top right side of the widget.
4. Click on it, and it will open a pop-up window, which will allow us to download data in CSV.
5. Similar to this, if we want to download data in excel, for that we don't have any support as of now. That's our target, we need to provide that support in the widget.

### Screenshots or Screencast
Screenshot 2022-09-28 at 2 32 41 AM

Yes

manishkushare:
Resolves #4133


shubh22:
Task linked: [CU-2vjyuam Table : Add support to download data in excel](https://app.clickup.com/t/2vjyuam)

kavinvenkatachalam (dismissed):
LGTM

kavinvenkatachalam (approved):

PR #4135, opened 2022-09-28
[improvement] Table : adds number as a cell type

### LINKED ISSUE #4079: Feature request: Number as a cell type for Tables widgets

arpitnath:
Resolves #4079

### DIFF
--- a/frontend/src/Editor/Components/Table/columns/index.jsx
+++ b/frontend/src/Editor/Components/Table/columns/index.jsx
@@ -151,6 +151,70 @@ export default function generateColumnsData({
             }
             return <span style={cellStyles}>{cellValue}</span>;
           }
+          case 'number': {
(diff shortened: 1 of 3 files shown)


shubh22:
Task linked: [CU-2uw3j7z Table improvements: add Number type as a cell type](https://app.clickup.com/t/2uw3j7z)

kavinvenkatachalam (changes requested):
@arpitnath I can see in the code that if I toggle on the editable, seems like we're doing some validation for `minValue` and `maxValue`. But, I don't see the fields in the UI. Could you please take a look?

image

kavinvenkatachalam (approved):

PR #4137, opened 2022-09-28
Use only state setters as dependent functions for registerAction (Task id: CU-2vk03t9)

### LINKED ISSUE #4136: Dropdown and Radio button widget are causing infinite loop
### Environment

MacOS

Dropdown and Radio button should work properly

They cause infinite loop

1. Add a dropdown or radio button

Yes

sherfin94:
Resolves #4136

The bug occurred due to non-state-setting functions being passed as dependencies for `registerAction`. This fix simply replaces those functions with their dependent state-setting-functions in the dependency array of `registerAction`.

### DIFF
--- a/frontend/src/Editor/Components/DropDown.jsx
+++ b/frontend/src/Editor/Components/DropDown.jsx
@@ -51,7 +51,7 @@ export const DropDown = function DropDown({
     async function (value) {
       selectOption(value);
     },
-    [JSON.stringify(values), selectOption]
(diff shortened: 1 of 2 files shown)


shubh22:
Task linked: [CU-2vk03t9 Dropdown and Radio button are causing infinite loop](https://app.clickup.com/t/2vk03t9)

kavinvenkatachalam (approved):

PR #4144, opened 2022-09-29
[ hotfixed ] : Datepicker widget height is not consistent with the parent container

### LINKED ISSUE #4143: [ hotfix ] : [ Datepicker widget ] height is not adjusting with the container height
### Version Information
v1.26.0

### Environment
Production

The height of the Datepicker widget should be consistent with the height of its parent container.

The height is not consistent with the parent container. When we increase the height of the container, the height of the widget still remains where it was earlier.

1. Create a new application or open an existing one.
2. Drag Datepicker widget on the canvas from the right sidebar.
3. Try to increase the height of the container of the widget. You can observe, the height of the date picker is not consistent with the container height. That's our issue.
4. To understand better, drag TextInput widget on the canvas from the right sidebar. and Increase container height. The same behaviour is supposed to be with the date picker widget.

### Screenshots or Screencast

https://user-images.githubusercontent.com/37823141/192963249-c7f49f88-73e7-445a-9e62-856ad10f1384.mov

### Any possible solutions?
Make the height of the Date picker widget consistent with the parent container.

frontend/src/Edior/Components/Datepicker.jsx

Yes

manishkushare:
Resolves #4143

### DIFF
--- a/frontend/src/Editor/Components/Datepicker.jsx
+++ b/frontend/src/Editor/Components/Datepicker.jsx
@@ -102,7 +102,7 @@ export const Datepicker = function Datepicker({
         showYearDropdown
         dropdownMode="select"
         excludeDates={excludedDates}
-        customInput={<input style={{ borderRadius: `${borderRadius}px` }} />}
+        customInput={<input style={{ borderRadius: `${borderRadius}px`, height }} />}
       />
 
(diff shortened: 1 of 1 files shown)


sherfin94 (approved):

shubh22:
Task linked: [CU-2vrkugb hotfix : datepicker height](https://app.clickup.com/t/2vrkugb)

PR #4142, opened 2022-09-29
Feature :: Table sort disable/enable

stepinfwd:
part of #4093
*disable sorting - option to enable/disable

PR #4151, opened 2022-09-29
[improvements] Table - exposed active sort options and add an event

### LINKED ISSUE #4150: [improvements] Table - exposed active sort options and add an event
## Description
We do not expose the sort options applied on the table to the app state. And there are no events associated with sorting the table data.

### Proposed request
Sorting options should be exposed to the app state and an event **On sorting columns** could be fired when columns are sorted.

arpitnath:
Resolves #4150


shubh22:
Task linked: [CU-2uw3jkz Table improvements: Event for sort change, and expose applied sorts](https://app.clickup.com/t/2uw3jkz)

kavinvenkatachalam (dismissed):
Changes LGTM. 

@sherfin94 It should work regardless of #4142 but we need to test with disabled sort

kavinvenkatachalam (approved):

PR #4155, opened 2022-09-29
An event should be fired when columns are editable and any cell value gets an update by the user 

### LINKED ISSUE #4154: Table widget: An event should be fired when columns are editable and any cell value gets an update by the user
### Proposed request
Proposing a new event for Table widget `On cell value changed`.
An event should be fired when columns are editable and any cell value gets an update by the user.

arpitnath:
Resolves #4154


shubh22:
Task linked: [CU-2vrnagc [Table] Add an event for Cell change](https://app.clickup.com/t/2vrnagc)

kavinvenkatachalam (approved):
LGTM

PR #4159, opened 2022-09-29
update shebang to bash

mikebarr24:


mikebarr24:
Reverting to original grep command in ./server/entrypoint.sh

Issue #3957, opened 2022-09-05
[Bug]:Not able to delete recently created applications


ajith-k-v:
### Environment

    ToolJet Cloud 

### What is the expected behaviour?

Not able to delete recently created applications.

### What is the current behaviour?

It should be able to delete all the applications. 

### How to reproduce the issue?

1. Create an application.
2. Mention a user in the comments
3. Try to delete the application.

### Screenshots or Screencast
[Video](https://www.loom.com/share/66861fa805d14f8895a6306b4f8cbcd3)

PR #4161, opened 2022-09-30
[Fix] Delete app not working if any user mentioned in comments

gsmithun4:
Resolves https://github.com/ToolJet/ToolJet/issues/3957


akshaysasidrn:
Changes looks good. 👍🏾 

Just two comments.
1. Please add/update spec for deleting app with comments.
2. We need to fix the column name read_on_dashboard separately as how we use the flag in the codebase is inverted.

akshaysasidrn (approved):

PR #4176, opened 2022-10-01
update the readme.md docx

JPatil22:
Adding the detail of **Heroku** for free tier and highlighting the details that are important


arpitnath:
Pull requests associated with issues created with `hactoverfest` tag would only be considered valid.

Navaneeth-pk (approved):

Issue #4165, opened 2022-10-01
Proposal to add border color option to Text Input widget


arpitnath:
## Proposed feature
Option to change and update the border color of the text input widget with the `Border color` property in the style tab.

### Description
Currently, the text input widget only supports **Border radius**, **Visibility**, and **Disable** as style definitions.

<img width="616" alt="Screenshot 2022-10-01 at 10 19 24 AM" src="https://user-images.githubusercontent.com/67645175/193393088-1548028e-d521-42f8-b4c7-db3d96d7b6b2.png">

reference: https://github.com/ToolJet/ToolJet/pull/1721

TheLearneer:
If it is up for grabs, I would love to work on it 🤚

arpitnath:
Hi @TheLearneer, assigning the issue to you.
Please make sure to mention the issue in your PR, it will be easier for us to review and merge faster.
Thanks for contributing and happy hacking 🚀

PR #4191, opened 2022-10-01
feat: Add customizable border color for text input component

TheLearneer:
fixes: #4165 

Example:
Screen Shot 2022-10-01 at 1 34 05 PM

**NOTE**: The right most input is the default text input without any border color defined.

### DIFF
--- a/frontend/src/Editor/Components/TextInput.jsx
+++ b/frontend/src/Editor/Components/TextInput.jsx
@@ -13,12 +13,17 @@ export const TextInput = function TextInput({
 }) {
   const textInputRef = useRef();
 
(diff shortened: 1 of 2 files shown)


TheLearneer:
NOTE: Having no default value for the component seems to mess up the design a bit, so if you guys could provide the desired default value as the border, I can add it 😄

arpitnath:
@TheLearneer Yeah, we can have the existing border color as default value, e.g., `#dadcde`

TheLearneer:
Hi @arpitnath Just in case you missed it, the changes have been made 🙏

gsmithun4:
deployed https://tj-4191.herokuapp.com/

arpitnath (approved):

Issue #4207, opened 2022-10-01
Text widget: Default value for background color


arpitnath:
## Proposed improvement
Add default value for background color
Default value: **#fff00000**

> This won't set the default value for text widgets that are already dropped in the editor for older apps.

laurisseau:
Hey, can I work on this?

arpitnath:
Hi @laurisseau , 

Thanks for showing interest in contributing to ToolJet. We've assigned the issue to you. 

If you have any questions related to setting up ToolJet, feel free to join our Slack community here: [Slack community](https://tooljet.com/slack)

laurisseau:
where is the CSS file

arpitnath:
@laurisseau 
Add `backgroundColor: { value: '#fff00000' }` in widgetConfig.js
Line: 1841

laurisseau:
ok

nilanshu24:
Hi, I am interested to work, can i work on this?

AbhishekTiwari23:
If the issue still exists, then please let me know. I want to contribute

arpitnath:
@laurisseau Any updates on this issue?

Abhay5855:
Can I work on this issue?

biswadeep25:
@arpitnath Can you assign this issue to me?

withshubh:
Unassigned @laurisseau as there was no update on it.

withshubh:
Hey @biswadeep25 

Thanks for showing interest in contributing to ToolJet. We've assigned the issue to you. 

If you have any questions related to setting up ToolJet, feel free to join our Slack community here: [Slack community](https://tooljet.com/slack)

SanjoyPator1:
Hi, I am interested in this issue. Can I work on this issue?

withshubh:
Hey @biswadeep25 - Any update on the PR?

Aviroop-001:
Hi @withshubh, I'd like to enhance the feature, please assign me the issue.

withshubh:
@Aviroop-001 Thanks for showing interest in contributing to ToolJet. We've assigned the issue to you.

rushikeshadhav:
Is this issue still open? If yes then I would like to work on it.

PR #4222, opened 2022-10-02
Added default value for background color.

dmdbilal:
Issue #4207

With the reference if the issue #4207. I've added default value for background color in text widget.
@arpitnath


Navaneeth-pk:
Please read the contributing guide - PRs should be created only for the issues that are assigned to you. 
@dmdbilal

dmdbilal:
Can you assign me this issue.

Issue #4209, opened 2022-10-01
Text widget: Default value for text color


arpitnath:
## Proposed improvement
Add default value for text color
Default value: **#000000**

> This won't set the default value for text widgets that are already dropped in the editor for older apps.

PrabeshPP:
Hey! I am interested in this issue .Can you assign me ?

arpitnath:
Hi @PrabeshPP, 

Thanks for showing interest in contributing to ToolJet. We've assigned the issue to you. 

If you have any questions related to setting up ToolJet, feel free to join our Slack community here: [Slack community](https://tooljet.com/slack)

dmdbilal:
I want to work on this issue.

vedant-z:
@dmdbilal I think if someone has already been assigned this issue, you can't make PR for this one.

This is upto my knowledge regarding ToolJet's guidelines.

dmdbilal:
Can you assign me this issue.

Navaneeth-pk:
@dmdbilal please do not spam by commenting on issues that are assigned to someone else and by creating PRs for issues on which others are working on.

dmdbilal:
Ok sorry

PR #4223, opened 2022-10-02
Added default color value for text widget

dmdbilal:
Issue #4209

With the reference of the issue #4209. i've added default color value for text widget.
@arpitnath


Navaneeth-pk:
Please do not work on issues assigned to another person.
@dmdbilal

PR #4271, opened 2022-10-03
[ Enhancement ]: Ability to export versioned apps ( frontend )

### LINKED ISSUE #4147: Ability to export versioned apps
At present, there is no means to export a particular version of an app.
As the app development is done there can multiple version present and it may not be necessary to export whole of the versions created for it. Moreover as the export size increases there can be inconsistencies on import due to db timeouts.

As a user I should be able to:
- to export the current version
- to export any other version (from list latest version to oldest version)
- option to export all

Figma: https://www.figma.com/file/iC5z99E1arfbfgrOf20j8W/Export-app

yes

manishkushare:
Resolves #4147

### DIFF
--- a/frontend/src/HomePage/ExportAppModal.jsx
+++ b/frontend/src/HomePage/ExportAppModal.jsx
@@ -3,13 +3,11 @@ import { default as BootstrapModal } from 'react-bootstrap/Modal';
 import moment from 'moment';
 import { appService } from '../_services/app.service';
 import { toast } from 'react-hot-toast';
-import { useMounted } from '@/_hooks/use-mount';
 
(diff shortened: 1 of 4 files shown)


shubh22:
Task linked: [CU-2vrm5d4 Ability to export versioned apps](https://app.clickup.com/t/2vrm5d4)

kavinvenkatachalam (changes requested):
@manishkushare As discussed here are a few suggestions,

1. Add padding to the radio button as it looks too close to the border and please reduce the spacing between the radio button and the text
2. Change the cursor to a pointer while hovering it
3. Remove `useMounted` and check for undefined & set the initial version ID while getting the response from the BE which can reduce some unwanted render.

kavinvenkatachalam (approved):
LGTM

PR #4342, opened 2022-10-07
[Bug]: Triggered Selection Only On Main Canvas

### LINKED ISSUE #4332: [Bug]: Selection is triggered on widgets
### Environment

 ToolJet Cloud

The selection should be triggered only when the user clicks on the canvas and tries to drag it. For eg, when the card is moved from one position to another position in the Kanban widget, the selection should not be triggered.

The selection is getting triggered for the following cases which should be avoided,

- In Kanban widget, move the card from one position to another
- In Table widget, resize the column width

1. Drop the Kanban widget into the canvas
2. Try to move the card from one place to another, we can notice the selection.

### Screenshots or Screencast

selection

### Any possible solutions?

Should stop the `Selecto` component from selecting when we dragged from inside in any of the widgets

Yes

kavinvenkatachalam:
Resolves #4332


shubh22:
Task linked: [CU-2w9zv4g Selection Issue on Widgets](https://app.clickup.com/t/2w9zv4g)

sherfin94 (approved):

PR #4349, opened 2022-10-07
[Improvements] Table: Adds Support for Server side filtering

### LINKED ISSUE #4348: Table widget: Add support for server side filtering
## Proposed feature request
Option to toggle between client and server-side filtering the data and populating the same to the table widget.

On **server-side filtering** toggled on, the current data of the table should get any updates.
Instead, an event **On filter changed** can be binded with the action **Run Query**, which fires the query on filter change and populate the table with the filtered data returned from the API.

arpitnath:
Resolves #4348


kavinvenkatachalam (approved):
LGTM

PR #4389, opened 2022-10-12
[enahancement] Added app loader page 

### LINKED ISSUE #4388: [enhancement] Should manage errors when a user tries to access an app it doesn't have access

Now, it Opens the app editor when a user tries to access an app it doesn't have access to.

Redirect to another page (Dashboard or show 403 error page)

Yes

shah21:
Resolves #4388


shubh22:
Task linked: [CU-2x0tkdb Manage Forbidden/Unauthorised app errors](https://app.clickup.com/t/2x0tkdb)

frontend/src/AppLoader/AppLoader.jsx
```diff
@@ -0,0 +1,180 @@
...
+      })
+      .catch((error) => {
+        setErrorDetails(error);
+        setLoading(false);
+      });
+  };
+
+  const handleError = () => {
+    if (errorDetails?.data) {
+      const statusCode = errorDetails.data?.statusCode;
```
gsmithun4, line 34: Shall we change the get apps API and get the application's organization id in error case?
and we can call the switch organization API

gsmithun4 (changes requested):

gsmithun4:
deployed https://tj-app-loader.herokuapp.com/

frontend/src/Editor/Viewer.jsx
```diff
@@ -152,19 +157,69 @@ class ViewerComponent extends React.Component {
...
+        return (window.location.href = `login/${orgId}?redirectTo=${path}`);
+      }
+    );
+  };
+
+  handleError = (errorDetails, appId, versionId) => {
+    if (errorDetails?.data) {
+      const statusCode = errorDetails.data?.statusCode;
+      if (statusCode === 403) {
+        const errorObj = JSON.parse(errorDetails.data?.message);
```
gsmithun4, line 211: we should consider the scenario when `errorDetails.data?.message` empty or not valid json. try catch can be added and on error redirected to home page, same for editor loader

gsmithun4 (changes requested):

frontend/src/_helpers/utils.js
```diff
@@ -530,3 +530,15 @@ export const hightlightMentionedUserInComment = (comment) => {
...
+  let parsed;
+
+  try {
+    parsed = JSON.parse(json);
+  } catch (e) {
+    console.log('JSON parse error');
+  }
+
+  return parsed;
+}
```
gsmithun4, line 544:
```suggestion
export function safelyParseJSON(json) {
  try {
    return JSON.parse(json);
  } catch (e) {
    console.log('JSON parse error');
  }
  return;
}
```

gsmithun4 (approved):

PR #4416, opened 2022-10-13
[ bug fixed ]: [ Table widget ] selecting option from  badge column  breaks the widget

### LINKED ISSUE #4414: [ Bug ] : [ Table widget ] selecting option from badge column  breaks the widget
### Version Information
v1.26.2

### Environment
Development

When we select any option from the badge column in the table widget, that option should be visible in the respective column.

When we select any option from the badge column in the table widget, it breaks it.

1. Create a new application or, open an existing one
2. Drag the table widget on the canvas from the right sidebar.
3. Click anywhere on the table widget to open the widget inspector on the right sidebar.
4. In the widget inspector, in the properties tab, look for the **Columns** section. 
5. In the columns section, choose one of any existing columns or add a new column by clicking on +add column
6. for the chosen column, click on that particular column, it will open a pop-up window adjacent to the left side of the right sidebar.
7. In the pop-up window, look for the **column type** section, and select column type as **badge** from the dropdown
8. In the same window, for column type as a badge, please also add/update the label and value fields. It will contain labels and values for the options in the dropdown. Both values should be array data types. and wrap the array inside `{{}}`
example - `{{ [ 1 , 2 , 3 ] }}` add this in both label and value field.
9. Now once we configured, a respective column for badge type, click anywhere outside the pop-up window to close it.
10. Now in the table widget on the canvas, look for the respective column of the type badge.
11. You will see a circular shaped object in each row of that particular column.
12. Click on any one of the circles, and it will open a dropdown, which we configured in step 8. Select any one value. Ideally, that selected value should appear in place of that particular circle in that particular column. But, it's not working as expected, rather it breaks the widget. That's our problem statement.

### Screenshots or Screencast
[Tooljet - Dashboard - 14 October 2022 - Watch Video](https://www.loom.com/share/1e6926b46af54f20b996c7f975aa8350)

Yes

manishkushare:
Resolves #4414

### DIFF
--- a/frontend/src/Editor/Components/Table/CustomSelect.jsx
+++ b/frontend/src/Editor/Components/Table/CustomSelect.jsx
@@ -7,7 +7,9 @@ export const CustomSelect = ({ options, value, multiple, onChange }) => {
 
   function renderValue(valueProps) {
     if (valueProps) {
-      return valueProps.value.split(', ').map((value, index) => (
+      const stringifyValue = String(valueProps.value);
+      const arrayOfValueProps = stringifyValue.includes(',') ? stringifyValue.split(', ') : stringifyValue.split(' ');
+      return arrayOfValueProps.map((value, index) => (
         <span key={index} {...valueProps} className="badge bg-blue-lt p-2 mx-1">
           {value}
         </span>


shubh22:
Task linked: [CU-2x0qk11 Bug : badge column in table widget](https://app.clickup.com/t/2x0qk11)

kavinvenkatachalam (approved):
LGTM

PR #4463, opened 2022-10-16
Fix typos in translation keys, remove one unnecessary and broken line

### LINKED ISSUE #4461: [refactor]: refactor of translation keys
### What area of the codebase is to be refactored?
 
During my work on #4261 I've noticed one thing in all translation JSON's:
 
 

 
One of the keys has a typo in the name, and also, few lines down we have the same one, but, with the correct name. I'm pretty sure, that should be removed.

Also, there are a few typos in other keys, which I also want to change, for example, in keys like `connectionVerifeid`

### What are the potential advantages of this proposed refactoring?
We gonna get rid of unnecessary and incorrect duplicated translation key, which is not really needed, because, we have the correct one already

daniloff200:
closes #4461

### DIFF
--- a/frontend/assets/translations/en.json
+++ b/frontend/assets/translations/en.json
@@ -25,7 +25,7 @@
         "components":"components",
         "send":"Send",
         "noConnection":"could not connect",
-        "connectionVerifeid":"connection verified",
+        "connectionVerified":"connection verified",
         "left":"Left",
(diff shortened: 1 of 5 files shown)


withshubh:
Closing this issue. ref: https://github.com/ToolJet/ToolJet/issues/4461#issuecomment-1279986877

PR #4591, opened 2022-10-24
Feature :: SelectRow CSA :: [ Table ]

### LINKED ISSUE #4093: Table enhacement

stepinfwd:
close #4093 
*selectedRow - Add CSA


shubh22:
Task linked: [CU-2vjw22z Table Enhancements :: add selectedrow CSA](https://app.clickup.com/t/2vjw22z)

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -455,6 +455,21 @@ export function Table({
...
+      const row = rows.find((item, index) => item.original[key] == value);
+      if (row != undefined) {
+        const selectedRowDetails = { selectedRow: item[0], selectedRowId: row.id };
+        mergeToTableDetails(selectedRowDetails);
+        setExposedVariables(selectedRowDetails).then(() => {
+          fireEvent('onRowClicked');
+        });
+      }
+    },
+    []
```
sherfin94, line 471: Try adding `mergeToTableDetails` as dependency here.

Otherwise, if you add a table inside a modal, and close the modal, and open it again, and then run the csa, the highlight won't work.

sherfin94 (changes requested):
I've posted a minor concern.

frontend/src/Editor/Components/Table/Table.jsx
sherfin94: Please also ensure that the above case doesn't fail.

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -455,6 +455,21 @@ export function Table({
     },
     [serverSidePagination, clientSidePagination, setPaginationInternalPageIndex]
   );
+  registerAction(
+    'selectRow',
+    async function (key, value) {
```
sherfin94, line 460: Please see the video. Try these steps to avoid the issue.

- Use table's data from `tableData` instead of currentState to filter to the desired row. (Remember that we're trying to avoid `currentState` as much as possible)
- Pass the stringified version of that dependency also

https://user-images.githubusercontent.com/20333972/201629853-03a933c9-bcf0-45ff-8e8c-dddcd1623d85.mov

sherfin94 (changes requested):
One issue commented.

sherfin94 (changes requested):
Hey @stepinfwd , the CSA is not working when modal is closed and opened again, and then CSA is run. Please check

PR #4595, opened 2022-10-25
[Feature]: Form Widget

### LINKED ISSUE #4594: [Feature]: Form Widget

To create a form, we need to add events to the inputs, textarea, button, etc., and bind in JS to collect those values. It will be better to have a form widget that collects all the value with onSubmit event. 

Form widget can be created with sub container layer so that any widget can be dropped into it. If any of the input widget is added to the form widget, then those value needs to be validated and added to the `data` exposed variable. It contains two events, `onInvalid`- will be triggered when any of the validation error is found and `onSubmit` - will be triggered when there are no validation errors.  

NA

Yes

kavinvenkatachalam:
Resolves #4594

#### changes
- form component
- form component doc


shubh22:
Task linked: [CU-2wv4yv9 Form Widget](https://app.clickup.com/t/2wv4yv9)

Issue #4601, opened 2022-10-25
Add Regression for postgreSQL datasource,


emidhun:
### Specify the kind of test

Cypress integration-test

### Describe the test

Cypress Regression Test

PR #4602, opened 2022-10-25
Added cypress test for data source PostgreSQL

emidhun:
Screenshot 2022-11-09 at 3 23 58 PM

Fixes: #4601


cypress-tests/cypress/constants/texts/postgreSql.js
```diff
@@ -0,0 +1 @@
+export const postgreSqlText = {};
```
ajith-k-v, line 1: @emidhun I think the text file is missing here. Please add the file.

emidhun: Added the missing labels to the file.

Navaneeth-pk (approved):

PR #4607, opened 2022-10-25
Fix: Login redirect on subpath

### LINKED ISSUE #4606: ToolJet redirects public apps to login page when using subpath
### Version Information
v1.27.2

### Environment
All

Should be able to view the public app

Redirects to login page

1. Start server with preconfigured subpath
2. create an app and make it public
3. try viewing the app without login using the shareable link

yes

akshaysasidrn:
Fixes #4606 

Without subpath: https://tooljet-ce.u1tgj2vfnqgrm.ap-south-1.cs.amazonlightsail.com/
With subpath: https://with-subpath.u1tgj2vfnqgrm.ap-south-1.cs.amazonlightsail.com/installations/low-code/oss/tooljet/

### DIFF
--- a/frontend/src/_services/authentication.service.js
+++ b/frontend/src/_services/authentication.service.js
@@ -117,9 +117,7 @@ function resetPassword(params) {
 function logout() {
   clearUser();
   const loginPath = (window.public_config?.SUB_PATH || '/') + 'login';
-  history.push(
-    loginPath + `?redirectTo=${window.location.pathname?.startsWith('/sso/') ? '/' : window.location.pathname}`
-  );
+  history.push(loginPath + `?redirectTo=${window.location.pathname}`);
 }
 
 function clearUser() {
--- a/frontend/src/_services/authentication.service.js
+++ b/frontend/src/_services/authentication.service.js
@@ -116,7 +116,10 @@ function resetPassword(params) {
 
 function logout() {
   clearUser();
-  history.push(`/login?redirectTo=${window.location.pathname?.startsWith('/sso/') ? '/' : window.location.pathname}`);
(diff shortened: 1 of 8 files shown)


frontend/src/_services/authentication.service.js
```diff
@@ -116,7 +116,10 @@ function resetPassword(params) {
 
 function logout() {
   clearUser();
-  history.push(`/login?redirectTo=${window.location.pathname?.startsWith('/sso/') ? '/' : window.location.pathname}`);
+  const loginPath = (window.public_config?.SUB_PATH || '/') + 'login';
+  history.push(
+    loginPath + `?redirectTo=${window.location.pathname?.startsWith('/sso/') ? '/' : window.location.pathname}`
```
gsmithun4, line 121:
```suggestion
  history.push(
    loginPath + `?redirectTo=${window.location.pathname}`
```
the start with /sso/ check is no longer required since we have removed log out call from Authorize page

gsmithun4 (changes requested):

gsmithun4 (approved):

frontend/src/_services/authentication.service.js
```diff
@@ -116,7 +116,8 @@ function resetPassword(params) {
 
 function logout() {
   clearUser();
-  history.push(`/login?redirectTo=${window.location.pathname?.startsWith('/sso/') ? '/' : window.location.pathname}`);
+  const loginPath = (window.public_config?.SUB_PATH || '/') + 'login';
+  history.push(loginPath + `?redirectTo=${window.location.pathname}`);
```
gondar00, line 120: Thinking whether we should add `window.location.reload()` after this line? cos page is blank the first time

PR #4622, opened 2022-10-26
[Bugfix]: Widget inspector Toggle element (Fx): component stops receiving updates after the every first

### LINKED ISSUE #4621: Widget inspector Toggle element (Fx): component stops receiving updates after the every first
On toggling the value for n number of times, the toggle should update the component level state (fx toggle element) and also updates the widget definitions.

After the very first update, the fx toggle element, stops updating its own state along with widget definitions

1. Add any widget, for example, a button widget
2. Open the widget inspector
3. try toggling the **Loading state** with the toggle element

arpitnath:
Resolves #4621


kavinvenkatachalam (approved):

emidhun:
@Kavin Tested and verified the functionality,
Working perfectly for me.
Found one issue, that will fix by #4620.

PR #4663, opened 2022-10-31
bug fixed : filter is not getting applied

### LINKED ISSUE #4662: [ Bug ] : [ Table Widget ] filters are not getting applied
### Version Information
v1.28.2

### Environment
All

All the filters should work correctly.

The filters are not working correctly. 

2. Drag the Table widget on the canvas from the right sidebar.
3. In the Table widget, look for the filter button at the top right side of the widget ( 1st button in the top right side from left to right ) and click on it, to apply filters
4. Once clicked, a pop-up window will appear to add the filters. 
5. By default 'contains' condition is selected, and in that scenario, if you change the other two fields, the filters seem to work fine.
6. As soon as, you change the condition (second input field ), filters are not getting applied correctly.

### Screenshots or Screencast
[Tooljet - Dashboard - 31 October 2022 - Watch Video](https://www.loom.com/share/3d331b5aacdc454d96bbc1331b78d733)

### Any possible solutions?
In the latest changes, the key named operation in filterValue object which is used to get the info of the applied filters is changed to a new key named as condition. Replace all occurrences of `operation` with `condition` in custom-filter.js

frontend/src/Editor/Components/Table/custom-filter.js

Yes

manishkushare:
Resolves #4662

### DIFF
--- a/frontend/src/Editor/Components/Table/custom-filter.js
+++ b/frontend/src/Editor/Components/Table/custom-filter.js
@@ -1,65 +1,65 @@
 export default function customFilter(rows, columnIds, filterValue) {
   try {
-    if (filterValue.operation === 'equals') {
-      return rows.filter((row) => row.values[columnIds[0]].toString() === filterValue.value);
-    }
+    if (filterValue.value.length !== 0 && filterValue.condition.length !== 0 && filterValue.column.length !== 0) {
+      if (filterValue.condition === 'equals') {
+        return rows.filter((row) => row.values[columnIds[0]].toString() === filterValue.value);
+      }
+      if (filterValue.condition === 'ne') {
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-2yemvw2 Table filter not working](https://app.clickup.com/t/2yemvw2)

sherfin94 (changes requested):
Hey @manishkushare , PR seems to work well!.

Please see if you can fix this also:

- 'Matches' and 'Should not match' should be based on exact matching. Presently, 'Sarah' will match against 'S' , make it so, that 'Sarah' will only be matched against 'Sarah'

sherfin94 (approved):

PR #4667, opened 2022-10-31
Fixed failing dashboard specs

ajith-k-v:
Fixes: #4666


ajith-k-v:
Everything is fixed and the spec is failing due to UI inconsistency  mentioned in this [issue](https://tooljet.qatouch.com/v2/defects/view/p/xVP5/did/QggWp)

emidhun (changes requested):
I think We can remove hardcoded waits in all possible sections,
and We can have some additional test cases on regression.

Issue #2334, opened 2022-02-23
Popout code editor closes when clicking in the middle of the editor


mslourens:
### Version Information
develop branch

### Environment
Build: Local build
OS: MacOS 12.1
Browser: Vivaldi

### What is the expected behaviour?
The popout code editor closes when you click below the first line (with the placeholder)

### What is the current behaviour?
Only clicking outside the popout should the code editor close.

### How to reproduce the issue?

1. Add a new table to the page
2. Select the table and click `add column`
3. Toggle the `make editable` to true
4. Click on the popout button on one of the validation fields
5. The code editor popout is visible
6. Click below the first line (notice the cursor changing from `text` to `default`)
7. The popout closes, although you didn't click outside the popout.

### Screenshots or Screencast

https://user-images.githubusercontent.com/1907152/155361809-6f6df7ec-63c0-4615-b29f-384d0ddc3e5c.mov

### Any possible solutions?
Not yet

### Can you identify the location in the driver source code where the problem exists?
Portal.jsx

### If the bug is confirmed, would you be willing to submit a PR?
Yes

sherfin94:
Hey @mslourens , since you mentioned that you're willing to submit PR, assigning you. Thank you for contributing!

withshubh:
Hey @mslourens 
Are you working on this issue?

mslourens:
No, I haven't found the time yet, so you can pick it up, if you like.

geisterfurz007:
Mind if I have a shot at this?

withshubh:
Sure @geisterfurz007 ! Assigned the issue to you.

PR #4674, opened 2022-10-31
fix: popout editor closing for tables

geisterfurz007:
(Supposedly) a fix for #2334.

## Background

The columns properties being a popout themselves led to a click "outside" the popout when clicking into the codemirror editor pane.

## Proposed solution

Adopting the workaround already present for the EventManagers own popout that likely had a similar issue in the past, CodeHinter now allows listening for the popped out editor's presence. This is used to collect a list of all present popouts and if any of them is present, the overlay is requested to not close when clicked outside of it.

## Notes

I am unsure about the changes made in the CodeHinter in the `handleToggle` function. I originally used a `useEffect` hook but that rerendered constantly due to the popOverCallback being a new function reference on each render. Since the Table is a class component, I wasn't really sure how to avoid this always becoming a new function reference (even if `useCallback` would have worked, that would have been a lot more boilerplate than this adds already), so I figured, I would just call the callback whenever `setIsOpen` is called with the same value (despite not being 100% sure what's exactly happening there with the Promise). Didn't completely blow up on me in manual testing so I am reasonably confident in this!


kavinvenkatachalam (dismissed):

withshubh:
Hey @geisterfurz007   
Thanks for contributing to ToolJet. :raised_hands: 
While this PR is under review, you can submit the form to claim your Hacktoberfest swag: https://airtable.com/shr1qR6xhqsWO7LD7

kavinvenkatachalam (approved):

PR #4676, opened 2022-11-01
[docs] cypress setup and running

withshubh:
#### changes
- Added a guide on setting up cypress and running test specs

### DIFF
--- a/docs/docs/contributing-guide/testing.md
+++ b/docs/docs/contributing-guide/testing.md
@@ -3,5 +3,33 @@ id: testing
 title: Testing
 ---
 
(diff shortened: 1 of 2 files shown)


emidhun (changes requested):
I think we can have commands for open `npm run cy:open` and the user can have UI to select spec 
Screenshot 2022-11-01 at 10 48 54 AM
Can we modify the doc accordingly?

emidhun (changes requested):
I think we can reduce the size of the headers "headed mode" and "Headless mode" and we can remove the PG_creds.

withshubh:
> I think we can reduce the size of the headers "headed mode" and "Headless mode" and we can remove the PG_creds.

Made the changes suggested.

emidhun (approved):
LGTM

Navaneeth-pk (approved):

PR #4692, opened 2022-11-02
adds minio removeObject feature

### LINKED ISSUE #4691: minio removeObject method to be added.
Minio works nice @mslourens & @gondar00, thank you for the great work. 

Can you please add the `removeObject` method which can be very similar to getObject method. 

Then we can also delete the objects after processing them.

tolgaulas:
Closes #4691

### DIFF
--- a/plugins/packages/minio/lib/operations.ts
+++ b/plugins/packages/minio/lib/operations.ts
@@ -44,6 +44,20 @@ export async function getObject(minioClient: MinioClient, queryOptions: object):
   return { Body: bodyContents };
 }
 
+export async function removeObject(minioClient: MinioClient, queryOptions: object): Promise<object> {
+  const stream = await minioClient.removeObject(queryOptions['bucket'], queryOptions['objectName']);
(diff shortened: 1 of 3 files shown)


plugins/packages/minio/lib/operations.ts
```diff
@@ -44,6 +44,20 @@ export async function getObject(minioClient: MinioClient, queryOptions: object):
...
+  const streamToString = (stream: Stream) =>
+    new Promise((resolve, reject) => {
+      const chunks = [];
+      stream.on('data', (chunk) => chunks.push(chunk));
+      stream.on('error', reject);
+      stream.on('end', () => resolve(Buffer.concat(chunks).toString('utf-8')));
+    });
+  const bodyContents = await streamToString(stream);
+
+  return { Body: bodyContents };
```
gondar00, line 58: Hi @tolgaulas, this is not required, can checkout minio [example](https://github.com/minio/minio-js/blob/master/examples/remove-object.js) for reference

PR #4693, opened 2022-11-02
add remove object method for Minio

### LINKED ISSUE #4691: minio removeObject method to be added.

mslourens:
fixes #4691


plugins/packages/minio/lib/operations.ts
```diff
@@ -59,3 +59,7 @@ export async function signedUrlForPut(minioClient: MinioClient, queryOptions: ob
 
   return { url };
 }
+
+export async function removeObject(minioClient: MinioClient, queryOptions: object): Promise<object> {
+  return await minioClient.deleteObject(queryOptions['bucket'], queryOptions['objectName']);
```
gondar00, line 64: Hey @mslourens the method is [`removeObject`](https://github.com/minio/minio-js/blob/master/examples/remove-object.js)
```suggestion
  return await minioClient.removeObject(queryOptions['bucket'], queryOptions['objectName']);
```

gondar00 (changes requested):
Need to add another case to the switch statement for `remove_object` [here](https://github.com/ToolJet/ToolJet/blob/develop/plugins/packages/minio/lib/index.ts#L14)

plugins/packages/minio/lib/operations.ts
mslourens: Thanks for the review! I implemented your comments.

gondar00:
Thank you for contributing to Tooljet @mslourens 💯

gondar00 (approved):

PR #4697, opened 2022-11-03
Queries fail when triggered on page load and self triggering queries crashes the app

### LINKED ISSUE #4696: Queries fail when triggered on page load and self triggering queries crashes the app
1. Queries should be run successfully on page load.
2. Queries should raise an error toast if it tries to trigger the same query

1. On page load, the queries fail, as queries are yet to be fetched and added to the app state
2. On triggering the same query, the app crashes

**Queries fail when triggered on page load** 
1. Add a REST API query
3. Add a RunJS query and trigger the rest API query and toggle on the **Run this query on page load**
4. Reload the app
self triggering queries crashes the app
1. Add a RunJS query (runjs1) and trigger the rest API query and trigger the same query with **actions** `actions.runQuery('runjs1')` or `queries.runjs1.run()`
2. Click on privew or run query

arpitnath:
Resolves #4696

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -323,6 +323,7 @@ class EditorComponent extends React.Component {
                 },
                 showQuerySearchField: false,
               });
+              this.runQueries(data.data_queries);
             }
           );
         });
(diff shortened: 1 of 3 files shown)


shubh22:
Task linked: [CU-2yu1zyc Queries fail when triggered on page load and self triggering queries crashes the app](https://app.clickup.com/t/2yu1zyc)

kavinvenkatachalam (approved):
LGTM

PR #4698, opened 2022-11-03
[Bugfix] : validating query name on throws error for non-existing queries

### LINKED ISSUE #4702: Validating query name, throws error for non-existing queries
On creating new queries, the query name should be validated with an existing query, and throw an error and fire toast only if a query exists with the same name

Deleting a query and creating a query. with the same name fires an error toast for query names, when the query with the same existed previously.

1. Create a query and name it. for example: **fetchUsers**, click on **Create**
2. Delete the query **(fetchUsers)**
3. Add a new query and rename it to **fetchUsers** and click on **Create** button

arpitnath:
Resolves #4702


kavinvenkatachalam (approved):
LGTM

kavinvenkatachalam (approved):

PR #4730, opened 2022-11-08
[docs] table improvements

withshubh:
#### changes
- minor change in doc's readme
- updated table doc with all the feature updates


withshubh:
- @kavinvenkatachalam please review the form doc
- @stepinfwd Please review the table doc, and let me know if anything is missing. Also, need your help with the description for `Server-side sort` and `Server-side filter`

kavinvenkatachalam:
@withshubh There are two property section in the doc for the `FORM` widget . Can we make it one?

image

withshubh:
@kavinvenkatachalam moved the form doc(with changes suggested) in the feature PR: https://github.com/ToolJet/ToolJet/pull/4595/files

docs/docs/widgets/table.md
```diff
@@ -152,26 +198,53 @@ If you don't wish to use the search feature altogether, you can disable it from
...
+
+This event is triggered when the column name header is clicked to apply sorting in `asc` or `desc`. The `sortApplied` variable is updated with an object having `column` and `direction` values.
+
+### Cell value changed
+
+If any cell of the table is edited, the `cell value changed` event is triggered.
+
+### Filter changed
+
+This event is triggered when filter is added from the header of the table. `filters` property of the table is updated with an object for each filter added. The objects will have `condition`, `value`, and `column` keys.
```
sherfin94, line 247:
```suggestion
This event is triggered when filter is added, removed, or updated from the filter section of the table. `filters` property of the table is updated to reflect the status of filters applied. The objects will have properties: `condition`, `value`, and `column`. 
```

sherfin94 (changes requested):
I've made a small suggestion, but good work anyway!

docs/docs/widgets/table.md
```diff
@@ -125,22 +143,50 @@ If server-side search is enabled, `on search` event is fired after the content o
...
-Activate this option on the inspector to have the last selected(clicked on) row to be highlighted.
+Enable this option to have the last selected(clicked on) row to be highlighted.
+
+### Disable sorting
+
+Enable this option to lock the sorting of columns when clicked on column name.
+
+### Server-side sort
+
+
```
withshubh, line 185:
```suggestion
### Server-side sort
When Server-side sort is enabled, clicking on the column headers will not automatically sort the table, instead, the `Sort applied` event will be fired and the applied sorting will be exposed as `sortApplied`. You can use this data to run any query that feeds data to the table in a manner that reflects the sorting applied.
```

docs/docs/widgets/table.md
```diff
@@ -125,22 +143,49 @@ If server-side search is enabled, `on search` event is fired after the content o
...
+### Disable sorting
+
+Enable this option to lock the sorting of columns when clicked on column name.
+
+### Server-side sort
+When Server-side sort is enabled, clicking on the column headers will not automatically sort the table, instead, the `Sort applied` event will be fired and the applied sorting will be exposed as `sortApplied`. You can use this data to run any query that feeds data to the table in a manner that reflects the sorting applied.
+
+### Server-side filter
+
+
```
withshubh, line 188:
```suggestion
### Server-side filter
When Server-side filter is enabled, applying filters will not automatically filter the table, instead, the `Filter changed` event will be fired and the applied filters will be exposed as `filters`. You can use this data to run any query that feeds data to the table in a manner that reflects the filters applied.
```

sherfin94 (approved):

PR #4736, opened 2022-11-08
[Bugfix] Query Manager: Creating a query, the options are set to their previous state used to create the last query from the same datasource/plugin

### LINKED ISSUE #4711: Creating a query, the options are set to their previous state used to create the last query from the same datasource/plugin
### Environment
Latest develop

Creating a query, the options should be set to default.

Creating a query, the options are set to their previous state used to create the last query from the same datasource/plugin

1. Add a datasource from the datasource manager
2. Create a query, and toggle on transformations or update the query options
3. Save the query
4. Create a new query with the same datasource.

arpitnath:
Resolves #4711


kavinvenkatachalam (approved):
LGTM

PR #4786, opened 2022-11-17
[Bug]: Exposed Default View Properly on Calendar Widget

### LINKED ISSUE #4785: [Bug]: Default View on Calendar is not Exposed Properly
### Version Information

Latest Version

### Environment

ToolJet Cloud

When the default view of the calendar property is changed, the changed value needs to be exposed in the inspector

The changed value of the default view in the calendar inspector is not exposed properly

1. Drop the calendar widget into the canvas
2. Change the `defaultView` to week
3. Open the inspector and the value of the `currentView` is still set to the old value (ie, month) unless we change the view in the calendar

### Screenshots or Screencast

image

### Any possible solutions?

While changing the property in the widget inspector, we're not updating the `currentView` exposedVariable.

File - `Calendar.jsx`

Yes

kavinvenkatachalam:
Fixes #4785

### DIFF
--- a/frontend/src/Editor/Components/Calendar.jsx
+++ b/frontend/src/Editor/Components/Calendar.jsx
@@ -51,6 +51,7 @@ export const Calendar = function ({
 
   const [currentDate, setCurrentDate] = useState(defaultDate);
   const [eventPopoverOptions, setEventPopoverOptions] = useState({ show: false });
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-30ygrec Default View on Calendar is not Exposed Properly](https://app.clickup.com/t/30ygrec)

PR #4797, opened 2022-11-18
[Bug]: Fixed Undefined Variable Bug

### LINKED ISSUE #4347: Data is not immediately available after running "set variable" in runjs
### Environment

- ToolJet | v1.26.1 (docker)

Use `actions.setVariable` to set the global variable and then read `variables` to get the data.

I got `undefined`.

1. Reload page.
1. Execute the following runjs code.
```javascript
actions.setVariable('t', 'yes');
alert(variables.t);
```

### Screenshots or Screencast

image

kavinvenkatachalam:
Fixes #4347

### DIFF
--- a/frontend/src/_helpers/appUtils.js
+++ b/frontend/src/_helpers/appUtils.js
@@ -398,26 +398,26 @@ export const executeAction = (_ref, event, mode, customVariables) => {
       case 'set-custom-variable': {
         const key = resolveReferences(event.key, _ref.state.currentState, undefined, customVariables);
         const value = resolveReferences(event.value, _ref.state.currentState, undefined, customVariables);
-        const customAppVariables = { ..._ref.state.currentState.variables };
-        customAppVariables[key] = value;
+        const customAppVariables = { ..._ref.state.currentState };
+        customAppVariables.variables[key] = value; // Updated object directly to get the variable value immediately
 
         return _ref.setState({
           currentState: {
             ..._ref.state.currentState,
-            variables: customAppVariables,
+            variables: customAppVariables.variables,
           },
         });
       }
 
       case 'unset-custom-variable': {
         const key = resolveReferences(event.key, _ref.state.currentState, undefined, customVariables);
-        const customAppVariables = { ..._ref.state.currentState.variables };
-        delete customAppVariables[key];
+        const customAppVariables = { ..._ref.state.currentState };
+        delete customAppVariables.variables[key]; // Updated object directly to get the variable value immediately
 
         return _ref.setState({
           currentState: {
             ..._ref.state.currentState,
-            variables: customAppVariables,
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-314ak6r Undefined Variable Bug](https://app.clickup.com/t/314ak6r)

frontend/src/_helpers/appUtils.js
```diff
@@ -398,26 +398,26 @@ export const executeAction = (_ref, event, mode, customVariables) => {
       case 'set-custom-variable': {
         const key = resolveReferences(event.key, _ref.state.currentState, undefined, customVariables);
         const value = resolveReferences(event.value, _ref.state.currentState, undefined, customVariables);
-        const customAppVariables = { ..._ref.state.currentState.variables };
-        customAppVariables[key] = value;
+        const customAppVariables = { ..._ref.state.currentState };
```
sherfin94, line 401: Let's fix this variable naming to `currentState` instead of `customAppVariables` as it is pointing to the current state.

frontend/src/_helpers/appUtils.js
```diff
@@ -398,26 +398,26 @@ export const executeAction = (_ref, event, mode, customVariables) => {
...
+        customAppVariables.variables[key] = value; // Updated object directly to get the variable value immediately
 
         return _ref.setState({
           currentState: {
             ..._ref.state.currentState,
-            variables: customAppVariables,
+            variables: customAppVariables.variables,
```
sherfin94, line 407: Instead of mutating a key directly, can we do something like:

```
currentState: {
            ..._ref.state.currentState,
            variables: { ...currentState.variables, [key]: value },
}
```

frontend/src/_helpers/appUtils.js
```diff
@@ -398,26 +398,26 @@ export const executeAction = (_ref, event, mode, customVariables) => {
...
           },
         });
       }
 
       case 'unset-custom-variable': {
         const key = resolveReferences(event.key, _ref.state.currentState, undefined, customVariables);
-        const customAppVariables = { ..._ref.state.currentState.variables };
-        delete customAppVariables[key];
+        const customAppVariables = { ..._ref.state.currentState };
```
sherfin94, line 414: Lets fix naming here also

frontend/src/_helpers/appUtils.js
```diff
@@ -398,26 +398,26 @@ export const executeAction = (_ref, event, mode, customVariables) => {
...
+        delete customAppVariables.variables[key]; // Updated object directly to get the variable value immediately
```
sherfin94, line 415: instead of `delete`, how about we use lodash's `omit` method to omit the key that we don't want?

sherfin94 (changes requested):
Suggested a few changes, please check.

PR #4803, opened 2022-11-18
Multipage : hide page and unhide page feature

arpitnath:
- option to hide the page
- adds hide icon on the page handler
- option to unhide the page
- remove the hide icon
- hidden pages are not rendered in the viewer

### DIFF
--- a/frontend/src/Editor/LeftSidebar/SidebarPageSelector/PageHandler.jsx
+++ b/frontend/src/Editor/LeftSidebar/SidebarPageSelector/PageHandler.jsx
@@ -117,7 +117,7 @@ export const PageHandler = ({
                 data-toggle="tooltip"
                 title="hidden"
                 className="mx-2"
-                src="assets/images/icons/eye.svg"
+                src="assets/images/icons/eye-off.svg"
                 height={14}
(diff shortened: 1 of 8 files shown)


frontend/src/Editor/LeftSidebar/SidebarPageSelector/PageHandler.jsx
```diff
@@ -101,9 +111,26 @@ export const PageHandler = ({
...
             {page.name}
           </div>
-          <div className="col-auto">
+          <div className="col-auto page-icons">
+            {isHidden && (
+              <img
+                data-toggle="tooltip"
+                title="hidden"
+                className="mx-2"
+                src="assets/images/icons/eye.svg"
```
sherfin94, line 120: Please use the eye-off icon that you added.

sherfin94 (changes requested):
We still need to render the hidden page if it is accessed via switch page action or via URL.

Hiding is only applicable to the menu.

Right now it seems like hidden pages are banned from being rendered.

Please fix.

sherfin94 (approved):

PR #4814, opened 2022-11-21
Multipage/mobile UI

arpitnath:
Navigation for Mobile-ui


sherfin94 (changes requested):
The hamburger seems to be overlapping components, I will start a conversation with product regarding this.

The home page is always loaded in mobile mode, it does not seem to be navigating to the page of corresponding handle. Please check.

sherfin94 (approved):

Issue #4820, opened 2022-11-21
Cannot build tooljet-client:development Docker image


frafra:
### Version Information
devel, 86917971a6fb12575c642fdff4e9d7725472eb3a

### Environment
Docker

### What is the expected behaviour?
The compose stack should build successfully when running `docker compose build`.

### What is the current behaviour?

It is not possible to build the `tooljet-client:development` image.

### How to reproduce the issue?

1. Checkout develop
2. `cp .env.example .env` and update `.env`
3. `docker compose build`

### Please provide any traces or logs that could help here.
```
 => CANCELED [tooljet-client:development 6/7] RUN npm --prefix frontend install                                                                                                                       0.5s
------                                                                                                                                                                                                     
 > [tooljet-plugins:development 6/7] RUN npm --prefix plugins install:                                                                                                                                     
#0 0.537 TypeError: Class extends value undefined is not a constructor or null                                                                                                                             
#0 0.537     at Object.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/socks-proxy-agent/dist/agent.js:114:44)                                                                                   
#0 0.537     at Module._compile (internal/modules/cjs/loader.js:1085:14)                                                                                                                                   
#0 0.537     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)                                                                                                                     
#0 0.537     at Module.load (internal/modules/cjs/loader.js:950:32)
#0 0.537     at Function.Module._load (internal/modules/cjs/loader.js:790:14)
#0 0.537     at Module.require (internal/modules/cjs/loader.js:974:19)
#0 0.537     at require (internal/modules/cjs/helpers.js:92:18)
#0 0.537     at Object.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/socks-proxy-agent/dist/index.js:5:33)
#0 0.537     at Module._compile (internal/modules/cjs/loader.js:1085:14)
#0 0.537     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
#0 0.537     at Module.load (internal/modules/cjs/loader.js:950:32)
#0 0.537     at Function.Module._load (internal/modules/cjs/loader.js:790:14)
#0 0.537     at Module.require (internal/modules/cjs/loader.js:974:19)
#0 0.537     at require (internal/modules/cjs/helpers.js:92:18)
[...]

PR #4821, opened 2022-11-21
fix: rich text editor does not accept html (#4819)

### LINKED ISSUE #4819: Rich text editor "Default Value" does not accept HTML
### Version Information
1.29.0

### Environment

ToolJet Docker compose, using tooljet/tooljet-ce:latest

Setting `Default Value` to an HTML valid string should show the rendered HTML in the form.

Rich text editor escapes the input.

1. Create a rich text editor
2. Set the `Default Value` to `<b>test</b>`

Use `convertFromHTML` and `createFromBlockArray` instead of `createFromText` here:
- https://github.com/ToolJet/ToolJet/blob/12d10360101b66c8ee6fb9636e3d80ece786f902/frontend/src/Editor/Components/DraftEditor.jsx#L152
- https://github.com/ToolJet/ToolJet/blob/12d10360101b66c8ee6fb9636e3d80ece786f902/frontend/src/Editor/Components/DraftEditor.jsx#L170

https://draftjs.org/docs/api-reference-data-conversion#convertfromhtml

Yes, I can try.

frafra:
Should fix #4819.
I still need to test this PR, as I cannot test locally, due to a problem with the Docker development environment: #4820.

### DIFF
--- a/frontend/src/Editor/Components/DraftEditor.jsx
+++ b/frontend/src/Editor/Components/DraftEditor.jsx
@@ -148,8 +148,10 @@ const InlineStyleControls = (props) => {
 class DraftEditor extends React.Component {
   constructor(props) {
     super(props);
+    const blocksFromHTML = ContentState.convertFromHTML(this.props.defaultValue);
+    const contentState = ContentState.createFromBlockArray(blocksFromHTML.contentBlocks, blocksFromHTML.entityMap);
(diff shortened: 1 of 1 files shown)


frafra:
@tejakummarikuntla hi! Could you please write something on why it has been closed/rejected?

PR #4824, opened 2022-11-21
fix: change client.Dockerfile.dev node_modules path

### LINKED ISSUE #4823: client.Dockerfile.dev: webpack not found
### Version Information
devel, 86917971a6fb12575c642fdff4e9d7725472eb3a

### Environment
Docker.

The container should start without errors

webpack cannot start

1. Run the client container

### Any possible solutions?
Set the `PATH` correctly.

https://github.com/ToolJet/ToolJet/blob/86917971a6fb12575c642fdff4e9d7725472eb3a/docker/client.Dockerfile.dev#L13-L14
https://github.com/ToolJet/ToolJet/blob/86917971a6fb12575c642fdff4e9d7725472eb3a/docker/client.Dockerfile.dev#L20

Yes, I would.

frafra:
Fix #4823.

According to git blame, this bug should have been there for 10 months at least. I suppose that the dev images are not tested or used. I am trying to make them work.

### DIFF
--- a/docker/client.Dockerfile.dev
+++ b/docker/client.Dockerfile.dev
@@ -10,8 +10,8 @@ WORKDIR /app
 
 COPY ./package.json ./package.json
 
-# add `/app/node_modules/.bin` to $PATH
-ENV PATH /app/node_modules/.bin:$PATH
+# add `/app/frontend/node_modules/.bin` to $PATH
+ENV PATH /app/frontend/node_modules/.bin:$PATH
 
 # Fix for heap limit allocation issue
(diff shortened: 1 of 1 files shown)


akshaysasidrn:
Hey @frafra, thanks for the fix. This looks good.
Most of internal team use locally without docker for development since the plugins changes are synced slowly with the server and frontend, hence we missed it.

akshaysasidrn (approved):

frafra:
Thank you :)
Having a working Docker environment would make life easier for me for cooperating on the project, but I could use a bit of guidance to fix it before I can test my draft PRs :)

PR #4833, opened 2022-11-22
Feature - Multi Environment support

gsmithun4:


server/src/entities/data_source.entity.ts
```diff
@@ -41,15 +41,40 @@ export class DataSource extends BaseEntity {
...
   @JoinColumn({ name: 'plugin_id' })
   plugin: Plugin;
+
+  @OneToMany(() => DataSourceOptions, (dso) => dso.dataSource)
+  dataSourceOptions: DataSourceOptions[];
+
+  @OneToMany(() => DataQuery, (dq) => dq.dataSource)
+  dataQueries: DataQuery[];
+
+  options;
```
gondar00, line 74: Missing type cast

server/src/services/data_sources.service.ts
```diff
@@ -1,115 +1,245 @@
...
+              }
+            });
+            ds.options = options;
+          } else {
+            ds.options = { ...(ds.dataSourceOptions?.[0]?.options || {}) };
+          }
+          delete ds['dataSourceOptions'];
+          return ds;
+        })
+        ?.filter((ds) => ds.kind !== 'restapidefault' && ds.kind !== 'runjsdefault');
```
gondar00, line 59: is it `ds.kind === 'restapi'` or `ds.kind === 'restapidefault?'`

server/src/services/data_sources.service.ts
```diff
@@ -1,115 +1,245 @@
...
-    options: Array<object>,
-    appId: string,
-    appVersionId?: string, // TODO: Make this non optional when autosave is implemented
-    pluginId?: string
+    appVersionId: string,
+    pluginId: string,
+    manager: EntityManager
+  ): Promise<DataSource> {
+    const defaultDataSource = await manager.findOne(DataSource, {
+      where: { kind: `${kind}default`, appVersionId },
```
gondar00, line 114: Is it feasible to avoid adding `default` key to the kind? that way can avoid some checks at some places

gsmithun4: No, since existing restapi data source conflicts. we have to add some privileges or filters for runjs and restapi which is created without a datasource in previous versions. So to differentiate we should add default

server/data-migrations/1667076251897-BackfillDataSources.ts
```diff
@@ -0,0 +1,46 @@
...
+      const dataQueries = await entityManager.query(
+        'select kind, id from data_queries where data_source_id IS NULL and app_version_id = $1',
+        [version.id]
+      );
+
+      for await (const dataQuery of dataQueries) {
+        await entityManager
+          .createQueryBuilder()
+          .update(DataQuery)
+          .set({ dataSourceId: dataQuery.kind === 'runjs' ? runjsDS : restapiDS })
```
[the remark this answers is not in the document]
gsmithun4: Yes but we need the different values in other loop 
https://github.com/ToolJet/ToolJet/blob/feature/multi-env/server/data-migrations/1667076251897-BackfillDataSources.ts#L38

server/data-migrations/1630003364070-TurnOnSSLForEveryPostgresDataSource.ts
```diff
@@ -1,22 +1,18 @@
...
   public async up(queryRunner: QueryRunner): Promise<void> {
     const entityManager = queryRunner.manager;
-    const queryBuilder = queryRunner.connection.createQueryBuilder();
-    const dataSourceRepository = entityManager.getRepository(DataSource);
 
-    const dataSources = await dataSourceRepository.find({
-      where: { kind: 'postgresql' },
-      select: ['id', 'options'],
-    });
+    const dataSources = await entityManager.query('select id, options from data_sources where kind = $1', [
```
akshaysasidrn, line 7: Curious, what broke here to explicitly specify the SQL query?

gsmithun4: `options` is removed from DataSources entity

server/data-migrations/1669055405494-removePluginFromDataQuery.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+export class removePluginFromDataQuery1669055405494 implements MigrationInterface {
```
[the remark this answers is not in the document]
gsmithun4: Yes, but this should be executed after the `1667076251897-BackfillDataSources.ts`. Since migration is executing first the data-migration throws error

server/data-migrations/1667076251897-BackfillDataSources.ts
```diff
@@ -0,0 +1,46 @@
...
+      .select()
+      .from('app_versions', 'app_versions')
+      .getRawMany();
+
+    for (const version of versions) {
+      let runjsDS, restapiDS;
+      for await (const kind of ['runjs', 'restapi']) {
+        const dataSourceResult = await entityManager.query(
+          'insert into data_sources (name, kind, app_version_id, app_id) values ($1, $2, $3, $4) returning "id"',
+          [`${kind}default`, `${kind}default`, version.id, version.app_id]
```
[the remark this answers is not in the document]
gsmithun4: We can give any name, won't be visible anywhere in front end, but kind should not be `restapi`, since its already used and we can't differentiate between rest api queries having datasource and not. Named it in such a way that to identify they are default one, should not be visible or editable for users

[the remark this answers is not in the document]
gsmithun4: Yes, we are managing the same in app imports and migrations

akshaysasidrn (approved):

PR #4879, opened 2022-12-02
Bugfix/4673 duplicate workspace name

snoseeds:
- Fix the timeout issue of tests
- Make unit tests run serially to prevent DB conflict
- Ensure `entrypoint.sh` has and use appropriate config files
- Update docs for docker tests to run correctly
- Make generated org name unique if not specified
- Prevent duplicate org names when creating and updating org

### DIFF
--- a/docs/docs/contributing-guide/setup/docker.md
+++ b/docs/docs/contributing-guide/setup/docker.md
@@ -162,27 +162,26 @@ Once you've updated the Dockerfile, rebuild the image by running `docker-compose
 
 Test config picks up config from `.env.test` file at the root of the project.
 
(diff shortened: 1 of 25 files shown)


gsmithun4:
@shubh22 we should find some resolution to make existing workspace names non identical, will keep it on hold

snoseeds:
@gsmithun4, thanks for reviewing this PR. Is there a way I can help with regularizing existing workspace names to be non-identical? By the way, while the existing workflow allows duplicate names, we cannot even be too sure if that has already resulted in duplicate names until some queries have been ran. Maybe some database scripts can really be handy to fix the possibly existing duplicates.

PR #4880, opened 2022-12-02
fix: onSave for existing credentials and placeholder

### LINKED ISSUE #4923: [Feature] :: Persisting old credentials while editing datasources

When editing datasources even if the encrypted values have not been changed they get altered in the backend and this forces the user to enter secrets/credentials every time while saving or else the credentials will be replaced.

Have a placeholder for encrypted fields so that the user gets a sense of value being persisted and fix the mechanism on how/when the credentials are being updated

Yes

vjaris42:
Resolves #4923 
1. Added placeholder for saved credentials (encrypted) in data sources.
2. Fixed credentials reset behaviour on save action in data sources.
3. Fix test connection when nothing changed while in edit mode.

### DIFF
--- a/frontend/src/_components/DynamicForm.jsx
+++ b/frontend/src/_components/DynamicForm.jsx
@@ -88,7 +88,7 @@ const DynamicForm = ({
       case 'textarea':
         return {
           type,
-          placeholder: description,
+          placeholder: options?.[key]?.encrypted ? '**************' : description,
           className: 'form-control',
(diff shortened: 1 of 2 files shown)


gsmithun4 (changes requested):
Hi @vjaris42 @shubh22 We have to find some ways to make encrypted fields empty

gsmithun4:
Duplicate https://github.com/ToolJet/ToolJet/pull/5008

Issue #4854, opened 2022-11-28
Table widget "Download as CSV" misses some values


kanunnikau:
### Environment
ToolJet Docker

### What is the expected behaviour?
Selecting download as CSV/Excel in table widget should download all data, as visible in the table.

### What is the current behaviour?
If the table has custom column mapping (i.e. the visible "Column name" is different from the "key" that's used to parse incoming data) then downloaded file doesn't have values for those fields. I suspect that the table widget isn't exporting the data that it's displaying, but is trying to export and filter the incoming data directly. Column names in the exported files are correct (same as in the table widget), but data is missing.

### How to reproduce the issue?
- fetch data with SQL SELECT,
- display in the table widget,
- (might not be related to the issue) add an "Action" column with a button to show some row details,
- set "Column name" different from "key" for some columns,
- export as CSV.

CSV file will have same column names as the displayed table widget. Data from columns with custom "Column names" will be missing. First "Actions" column will be empty, as expected.

PR #4906, opened 2022-12-07
[ Bug fixed ] : [ Table widget ]  download as excel exports extra columns

### LINKED ISSUE #4858: Table widget "Download as Excel" exports extra columns
### Environment
ToolJet Docker v1.29.0

Table widget exports data to Excel same as it's displaying.

Table exports:
- all column names as visible on the screen,
- all data as provided to the table.
There can be a mismatch between column names and data structure. For example, if there is a first "Actions" column then it will also be present in the export, but will contain values of the next column (all values are shifted left).
Also, if the SQL result set contains extra fields (used in the SQL SELECT ... WHERE filter, but not added to the table widget and not visible there) then those fields will be added as the last, unnamed column.

- add 'Actions' column,
- have some extra columns in the SQL results that are not added to the table,
- export to Excel,
- see mismatch between column names and values. First column is called "Actions", but contains values of the second column, second column contains values of third, etc until the end — the last column has valid name, but contains values of a column that wasn't added to the table.

manishkushare:
Resolves #4858 & #4854


shubh22:
Task linked: [CU-32qxp85 Bug:  [ Table widget ] download as excel exports extra columns](https://app.clickup.com/t/32qxp85)

kavinvenkatachalam (approved):

kavinvenkatachalam (dismissed):

sherfin94 (approved):

PR #4908, opened 2022-12-07
[Feature] :: Allow users to be filtered via status

vjaris42:
Provisioned a user status dropdown through which the users can be filtered in Manage users page.

### DIFF
--- a/frontend/ee/components/UsersPage/UsersFilter.jsx
+++ b/frontend/ee/components/UsersPage/UsersFilter.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import Select from '@/_ui/Select';
 
 const userStatusOptions = [
+  { name: 'All', value: '' },
   { name: 'Active', value: 'active' },
   { name: 'Invited', value: 'invited' },
   { name: 'Archived', value: 'archived' },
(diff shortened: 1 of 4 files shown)


shubh22:
Task linked: [CU-32qyxnm [Feature] :: Allow users to be filtered via status](https://app.clickup.com/t/32qyxnm)

server/src/services/organizations.service.ts
```diff
@@ -166,6 +170,10 @@ export class OrganizationsService {
           qb.andWhere('lower(user.lastName) like :lastName', {
             lastName: `%${options?.lastName.toLowerCase()}%`,
           });
+        if (options?.status)
+          qb.andWhere('lower(organization_user.status) like :status', {
```
gsmithun4, line 174: Should not be wild card search, should be 
```
qb.andWhere('organization_user.status = :status'
```
no need to convert into lower cases

frontend/ee/components/UsersPage/UsersFilter.jsx
```diff
@@ -1,15 +1,27 @@
 import React from 'react';
+import Select from '@/_ui/Select';
+
+const userStatusOptions = [
```
gsmithun4, line 4: there should be `all` option

gsmithun4 (changes requested):

gsmithun4 (approved):

PR #4912, opened 2022-12-07
[Enhancement] :: Allow users to add custom query parameters for postgresql connections

vjaris42:


akshaysasidrn (approved):

akshaysasidrn:
How do we communicate with the users as to what connection options are available? Should we link the driver options in the docs?

shubh22:
@akshaysasidrn we can copy the events from our package and share them in our own doc. @withshubh can do this while documenting this PR. works?
cc - @vjaris42

sherfin94 (approved):

PR #4930, opened 2022-12-09
fixes: app import/export: page events should be associated with correct queries

arpitnath:
Fixes: import/export: page events should be associated with correct queries


shubh22:
Task linked: [CU-32xjfq4 on importing an app the queries associated with 'on page load' are removed](https://app.clickup.com/t/32xjfq4)

sherfin94 (approved):

Issue #4787, opened 2022-11-17
DTO validation for version name validation


withshubh:
### Is your proposal related to a problem?

The Version Name Validation should have a data transfer object.

### Describe the solution you'd like

Create a DTO ([example](https://github.com/ToolJet/ToolJet/blob/develop/server/src/dto/version-create.dto.ts)) for title length validation of the workspace

reference: https://github.com/ToolJet/ToolJet/pull/4612

vedant-z:
Hi @withshubh May I try this?
Also if you could explain a little bit on what needs to be done it would be great.

mikebarr24:
@withshubh If my current PR gets accepted, could you assign me to this. I think it's similar to what I've just submitted.
Thanks!

withshubh:
Sure! @mikebarr24 - assigned the issue to you.

PR #4938, opened 2022-12-10
chore: add backend version name length validation

mikebarr24:
Addresses #4787


gsmithun4:
please remove this validations since we have added the same in DTO level
https://github.com/ToolJet/ToolJet/blob/develop/server/src/services/apps.service.ts#L231

server/src/dto/version-create.dto.ts
```diff
@@ -1,11 +1,12 @@
-import { IsNotEmpty, IsOptional, IsString, IsUUID } from 'class-validator';
+import { IsNotEmpty, IsOptional, IsString, IsUUID, Length } from 'class-validator';
 import { Transform } from 'class-transformer';
 import { sanitizeInput } from '../helpers/utils.helper';
 
 export class VersionCreateDto {
   @IsString()
   @Transform(({ value }) => sanitizeInput(value))
   @IsNotEmpty()
+  @Length(1, 25)
```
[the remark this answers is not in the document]
mikebarr24: Would it be worth adding MinLength as well and then removing
```
    if (!versionName) {
      throw new BadRequestException('Version name cannot be empty.');
    }
```
from apps.service.ts?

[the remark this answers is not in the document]
mikebarr24: I hope this is OK. I've added the MinLength as well and removed the relevant logic from app.service.ts

server/entrypoint.sh
```diff
@@ -1,18 +0,0 @@
-#!/bin/bash
```
gsmithun4, line 1: Looks like its not related to this PR, please revert

gsmithun4 (changes requested):

server/entrypoint.sh
mikebarr24: Sorry! Still a bit clumsy with git. I think I've fixed it here.

mikebarr24: I wonder would it be better to create a new PR?

gsmithun4: looks good now

mikebarr24 (commented):
I think I've fixed this. Not entirely sure of the review process here so I might have done something wrong.

mikebarr24 (commented):
I think I've fixed this. (Sorry! Playing around with github and realise I can't delete this)

gsmithun4:
@mikebarr24 Please resolve conflicts

gsmithun4 (approved):

PR #4946, opened 2022-12-12
Resolve issue that caused components to not get exposed on certain apps (Task ID - #32xjeex)

sherfin94:
This PR moves the state setting functions inside `fetchApp` of Editor to the `then` block of `computeComponentState`, there by avoiding any async issues with state setting.


shubh22:
Task linked: [CU-32xjeex Multi pages - on reload or first app load the page components in the currentState is empty](https://app.clickup.com/t/32xjeex)

arpitnath (approved):

PR #4948, opened 2022-12-12
[v2-beta/Bugfix] : Tab widget background color should be consistent with the current theme

arpitnath:
Resolves: Tab widget background color should be consistent with the current theme


shubh22:
Task linked: [CU-33381xj Tab widget background color is white in dark-mode](https://app.clickup.com/t/33381xj)

sherfin94 (approved):

PR #4951, opened 2022-12-12
Feature: Tooljet Database

akshaysasidrn:


server/test/controllers/tooljet_db.e2e-spec.ts
```diff
@@ -0,0 +1,270 @@
+import * as request from 'supertest';
```
akshaysasidrn, line 1: This spec is invalid and needs revision

server/src/services/postgrest_proxy.service.ts
```diff
@@ -0,0 +1,97 @@
...
+      where: {
+        organizationId: user.defaultOrganizationId,
+        tableName: In(requestedTableNames),
+      },
+    });
+
+    const obtainedTableNames = internalTables.map((t) => t.tableName);
+    const tableNamesNotInOrg = requestedTableNames.filter((tableName) => !obtainedTableNames.includes(tableName));
+
+    if (isEmpty(tableNamesNotInOrg)) return internalTables;
```
gondar00, line 93: This can be removed since we are scoping to orgId

arpitnath (approved):

PR #4953, opened 2022-12-12
[Bugfix] fix issue when tooltip contains number

### LINKED ISSUE #4929: components with Tooltip option crashes when non-string value is entered

When a number `{{324}}` or `{{components.textInput1}}` values are entered then the component crashes and throws `Something went wrong`

The component should not crash when a non-string value is entered in tooltip option field.

1. Add a button and text-input component
2. Go to button's properties
3. Go to the Tooltip option and enter `{{342}}` or `{{components.textInput1}}`

### Screenshots or Screencast

Screenshot 2022-12-09 at 2 34 04 PM

mslourens:
Fixes #4929 - convert possible number to string when showing tooltip


sherfin94 (approved):
Thank you!

PR #4962, opened 2022-12-13
fixes: app crash on selecting page switch event for query manager

arpitnath:
Fixes:
App crash for add page switch event for
- [x] queries (Query manager)
- [x] form widget
- [x] page event handler (page handler menu)


shubh22:
Task linked: [CU-3338ay2 App is crashing when selecting switch page action from event manager](https://app.clickup.com/t/3338ay2)

frontend/src/Editor/QueryManager/QueryManager.jsx
```diff
@@ -801,6 +807,7 @@ class QueryManagerComponent extends React.Component {
                     components={this.props.allComponents}
                     apps={this.props.apps}
                     popoverPlacement="top"
+                    pages={this.state.pages}
```
kavinvenkatachalam, line 810: @arpitnath Can we access it through `props` directly instead of setting `state`?

kavinvenkatachalam (changes requested):
@arpitnath Added a suggestion on `QueryManager`

emidhun:
Verified the fixes now, and will have complete regression after all the fixes are done(v2).

kavinvenkatachalam (approved):

PR #4995, opened 2022-12-15
[Feature] :: Allow users to invite users in bulk

### LINKED ISSUE #4922: [Feature] :: Bulk upload users for invitations

Only one user can be invited at a time which is a little cumbersome when there are many users to be invited

Allow users to upload csv file with user details and have them processed and validated on the backend.

Yes

vjaris42:
Resolves #4922 
1. Only 500 users and a file size of 1MB can be processed at a time.
2. Added checks for valid groups and existing users.
3. Users are only added once all the validations are passed

### DIFF
--- a/server/src/services/organizations.service.ts
+++ b/server/src/services/organizations.service.ts
@@ -1,4 +1,5 @@
 import { BadRequestException, Injectable } from '@nestjs/common';
+import * as csv from 'fast-csv';
 import { InjectRepository } from '@nestjs/typeorm';
 import { GroupPermission } from 'src/entities/group_permission.entity';
 import { Organization } from 'src/entities/organization.entity';
@@ -14,6 +15,10 @@ import { OrganizationUsersService } from './organization_users.service';
 import { UsersService } from './users.service';
 import { InviteNewUserDto } from '@dto/invite-new-user.dto';
 import { ConfigService } from '@nestjs/config';
+import { decamelize } from 'humps';
+import { Response } from 'express';
+
+const MAX_ROW_COUNT = 500;
 
 type FetchUserResponse = {
   email: string;
@@ -28,6 +33,12 @@ type FetchUserResponse = {
 
 type UserFilterOptions = { email?: string; firstName?: string; lastName?: string };
 
+interface UserCsvRow {
+  first_name: string;
+  last_name: string;
+  email: string;
+  groups?: string;
+}
 @Injectable()
 export class OrganizationsService {
   constructor(
@@ -465,6 +476,7 @@ export class OrganizationsService {
       lastName: inviteNewUserDto.last_name,
       email: inviteNewUserDto.email,
     };
+    const groups = inviteNewUserDto.groups ?? [];
 
     let user = await this.usersService.findByEmail(userParams.email);
     let defaultOrganization: Organization,
@@ -495,7 +507,7 @@ export class OrganizationsService {
       user = await this.usersService.create(
         userParams,
         currentUser.organizationId,
-        ['all_users'],
+        ['all_users', ...groups],
         user,
         true,
         defaultOrganization?.id,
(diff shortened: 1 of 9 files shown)


server/src/services/organizations.service.ts
```diff
@@ -539,4 +551,101 @@ export class OrganizationsService {
...
+          return next(null, data.first_name !== '' && data.last_name !== '' && emailPattern.test(data.email));
+        });
+      })
+      .on('data', function () {})
+      .on('data-invalid', (row, rowNumber) => {
+        invalidRows.push(rowNumber);
+      })
+      .on('end', async (rowCount: number) => {
+        try {
+          if (rowCount > MAX_ROW_COUNT) {
```
gsmithun4, line 615: Best If we can do this check before validations, but I don't think its possible with fast-csv. @vjaris42 wdyt?

server/src/services/organizations.service.ts
```diff
@@ -539,4 +551,101 @@ export class OrganizationsService {
...
+      .transform((row: UserCsvRow, next) => {
+        return next(null, {
+          ...row,
+          groups: this.decamelizeDefaultGroupNames(row?.groups),
+        });
+      })
+      .validate(async (data: UserCsvRow, next) => {
+        await dbTransactionWrap(async (manager: EntityManager) => {
+          //Check for existing users
+          const user = await this.usersService.findByEmail(data?.email);
```
gsmithun4, line 585: Shall we use same `manager` for all queries?

server/src/services/organizations.service.ts
```diff
@@ -539,4 +551,101 @@ export class OrganizationsService {
...
+          }
+
+          if (existingUsers.length) {
+            throw new BadRequestException(
+              `User with email ${existingUsers.join(', ')} already exists. No users were uploaded`
+            );
+          }
+
+          for (let i = 0; i < users.length; i++) {
+            await this.inviteNewUser(currentUser, users[i]);
```
gsmithun4, line 638: This should be synchronous operation? and we should pass the entity manager.

gsmithun4 (changes requested):

server/src/services/organizations.service.ts
vjaris42: Actually no we can't do that and that's just not the limitation with fast-csv but any other library as well because the data is being processed in chunks so once it goes through the data only then we can check the row count.

server/src/services/organizations.service.ts
vjaris42: Since we decided to limit the number of rows to 500 we concluded that we can keep this synchronous. The problem with keeping this as an asynchronous insertion is that the pool size will be exhausted and it will fail. `await` makes sure that the previous insertion is done and sitting idle it will reuse that connection only. Let me know if any other options are suitable here. Thanks

server/src/services/organizations.service.ts
vjaris42: Yup I will add the manager from  getManager() here

server/src/services/organizations.service.ts
vjaris42: I will add the entitymanager here and check if its using the same connection from pool or not while keeping it asynchronous.

server/src/services/organizations.service.ts
gsmithun4: You can use the manager 
```
```

server/src/services/organizations.service.ts
```diff
@@ -508,35 +522,145 @@ export class OrganizationsService {
...
+          if (user?.organizationUsers?.some((ou) => ou.organizationId === currentUser.organizationId)) {
+            existingUsers.push(data?.email);
+          } else {
+            users.push(data);
+          }
+
+          //Check for invalid groups
+          const receivedGroups = data?.groups;
+          for (const group of receivedGroups) {
+            const orgGroupPermission = await manager.findOne(GroupPermission, {
```
gsmithun4, line 609: Shall we pre-load the groups list and check?

gsmithun4 (changes requested):

kavinvenkatachalam (approved):
LGTM

gsmithun4 (dismissed):

gsmithun4 (dismissed):

shubh22:
Task linked: [CU-85zrgm4ty Allow users to invite bulk users](https://app.clickup.com/t/85zrgm4ty)

gsmithun4 (approved):

PR #4997, opened 2022-12-15
[fix] Redirect URL didn't work for preview apps

### LINKED ISSUE #4996: Redirect URL didn't work for preview apps
### Version Information
v2-beta
### Environment

All

Should redirect to preview app if the 401 error came from there

1. Create an app and copy the preview app URL
2. Logout and run the preview app URL & Observe

Yes

shah21:
Resolves #4996

### DIFF
--- a/server/src/interceptors/valid.app.interceptor.ts
+++ b/server/src/interceptors/valid.app.interceptor.ts
@@ -1,0 +1,34 @@
+import {
+  ExecutionContext,
+  Injectable,
+  NotFoundException,
+  NestInterceptor,
+  CallHandler,
+  BadRequestException,
+} from '@nestjs/common';
+import { AppsService } from 'src/services/apps.service';
+import { Observable } from 'rxjs';
+import { App } from 'src/entities/app.entity';
+
+@Injectable()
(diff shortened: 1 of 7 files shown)


server/src/interceptors/valid.app.interceptor.ts
```diff
@@ -0,0 +1,34 @@
...
+    const { id, slug } = request.params;
+    let app: App;
+    if (id) {
+      app = await this.appsService.find(id);
+      if (!app) throw new NotFoundException('App not found. Invalid app id');
+    } else if (slug) {
+      app = await this.appsService.findBySlug(slug);
+      if (!app) throw new NotFoundException('App not found. Invalid app id');
+    } else {
+      throw new BadRequestException();
```
gsmithun4, line 28:
```suggestion
    if (!(id || slug)) {throw new BadRequestException();}
    const app = await this.appsService.find(id || slug);
    
      if (!app) throw new NotFoundException('App not found. Invalid app id');

```

gsmithun4 (changes requested):

gsmithun4 (approved):

PR #4999, opened 2022-12-16
[Bug fixed ] :[ Table widget ] Badge column type is always editable in table row regardless of the editable property

### LINKED ISSUE #2836: Badge column type is always editable in table row regardless of the editable property
### Environment
Docker

In a table, when a column of badge type is set non-editable, it should not be editable.

Badge column type is always editable in table row regardless of the editable property

frontend/src/Editor/Components/Table/Table.jsx
no `column.isEditable` used in below code.

    case 'badge':
    case 'badges': {
            return (
              <div>
                <CustomSelect
                  options={columnOptions.selectOptions}
                  value={cellValue}
                  multiple={columnType === 'badges'}
                  onChange={(value) => {
                    handleCellValueChange(cell.row.index, column.key || column.name, value, cell.row.original);
                  }}
                />
              </div>
            );
     }

manishkushare:
Resolves #2836


sherfin94 (changes requested):
Hey @manishkushare ,

It seems like instead of making the column non-editable, it is now disabled and dimmed away. Would it be possible to make it just non-editable?

Looking that the codebase, we are considering `!isEditable` and `disabled` as the same thing, this should not be the case.

One solution would be to move rendering badges from inside of the custom select to outside of it, if the column is not editable.

sherfin94 (approved):

kavinvenkatachalam (approved):

PR #5008, opened 2022-12-19
Make encrypted fields editable

### LINKED ISSUE #5007: [[Feature]] :: Make encrypted fields editable to empty string
There is no option to reset a encrypted field which is already saved in the db

Give an editable toggle to all encrypted fields which when toggled on existing saved datasources will reset the value and for new datasources the fields will be by default enabled for edit.

Yes

vjaris42:
Resolves #5007 

- Allow users to reset encrypted fields in existing datasource by clicking on the edit button
- By default when adding a new datasource edit button will not be shown.

### DIFF
--- a/docs/docs/data-sources/rethinkdb.md
+++ b/docs/docs/data-sources/rethinkdb.md
@@ -4,7 +4,7 @@ title: RethinkDB
 ---
 # RethinkDB
 
-ToolJet can connect to RethinkDB databases to read and write data. For more info visit::https://rethinkdb.com/api/javascript/.com/
+ToolJet can connect to RethinkDB databases to read and write data. For more info visit this [doc](https://rethinkdb.com/api/javascript).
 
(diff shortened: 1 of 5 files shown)


gsmithun4 (changes requested):
On create hide edit toggle switch
Click on edit and keep the field empty is not resetting the value

kavinvenkatachalam (changes requested):
@vjaris42 Few suggestions,

- Placement of the toggle seems different. Instead of setting it below, we can place to the right of the label. Check with Jaseem, he might help with the right placement
image

- When I try to toggle the switch for the first time, it is not selecting/unselecting the value. It works only after the second selection. Attaching the GIF,

Dec-20-2022 08-53-55

kavinvenkatachalam (dismissed):

shubh22:
Task linked: [CU-85zrgm4y6 Make encrypted fields editable](https://app.clickup.com/t/85zrgm4y6)

PR #5011, opened 2022-12-19
[ Bug fixed ] : [Multi-Select widget ] app freezes when we use control component action to control the same multi-select

### LINKED ISSUE #4975: app freezes when we use control component action to control the same multi-select

When we use the multi-select's event handler for control component action to control the same multi-select , the app builder freezes

The app should not freeze and should ideally throw an error.

1. Add a multi-select
2. Add event handler to multi-select and select control component action
3. Choose any action
4. Try to trigger the event from the component

manishkushare:
Resolves #4975

### DIFF
--- a/frontend/src/Editor/Components/Multiselect.jsx
+++ b/frontend/src/Editor/Components/Multiselect.jsx
@@ -75,7 +75,10 @@ export const Multiselect = function Multiselect({
   registerAction(
     'selectOption',
     async function (value) {
-      if (!selected.some((option) => option.value === value)) {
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-85zrggm59 app freezes when we use control component action to control the same multi-select](https://app.clickup.com/t/85zrggm59)

kavinvenkatachalam (approved):
LGTM

kavinvenkatachalam (approved):

Issue #5013, opened 2022-12-19
Update node version to 14.19.3


gondar00:
### Is your proposal related to a problem?
Update the node version to v14.19.3

### Describe the solution you'd like
The versions above the current node version i.e. 14.17.3 are majorly related to security fixes as seen in [changelog](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V14.md#14.17.3)

### If the feature is approved, would you be willing to submit a PR?
Yes

shubh22:
Closing this as we are moving to node 18 and decided to not upgrade to this

PR #5014, opened 2022-12-19
chore: update node version to 14.19.3

gondar00:
fixes: #5013


shubh22:
Upgrading to node 18 now. This will not be required.

PR #5021, opened 2022-12-20
fix: show confirmation box if filter options are empty in query

arpitnath:
Resolves:
**Delete row operation**
if filter options are not passed, i.e., with empty options all rows will be deleted


frontend/src/Editor/QueryManager/QueryManager.jsx
```diff
@@ -427,6 +427,12 @@ class QueryManagerComponent extends React.Component {
     this.setState({ buttonText: text, shouldRunQuery: shouldRunQuery });
   };
 
+  showConfirmationOnDeleteOperationFordbQuery = (queryKind, options) => {
```
akshaysasidrn, line 430: Can we move this loigic inside toojetdb component?

akshaysasidrn (approved):

PR #5025, opened 2022-12-20
File picker issues on CSV parsing

### LINKED ISSUE #4011: Unexpected behaviour when accessing parsed data from file picker with csv files
### Version Information
1.14.0
### Environment
Docker - self host

when csv files are uploaded using file picker component and accessing the parsed data
1. Even if a column value is of multiple lines then it should be considered as single value.
2. Even if the column value is with multiple comma's then the whole value should be considered as single value

when csv files are uploaded using file picker component and accessing the parsed data
1. If a column value is of multiple lines then each line in that column is being read as new line.
2. If the column value is with multiple comma's then the value is being considered only upto first comma and remaining content is missing.

1. Add two rows in sheet
2. In first row add multiple lines in a single column value
3. In second row add a value with multiple comma's in it
4. Download it as csv file
5. Select a file picker component. enable the parsed content option for it
6. Upload the csv file using file picker component.
7. Check the parsed data value for both rows

### Screenshots
downloaded_sheet_csv
sheet_sample

parsed_data_screenshot

This issue is reproduced in latest version also.

vjaris42:
Resolves #4011

- Parse multiline cell values
- Parse cell values including comma in the string


arpitnath (dismissed):
LGTM!

shubh22:
Task linked: [CU-85zrgm53u CSV parsing issue on file picker component](https://app.clickup.com/t/85zrgm53u)

PR #5055, opened 2022-12-26
[Bugfix]internal storage - column validations

arpitnath:
Fixes column: failed varchar and float validations


server/src/dto/create-postgrest-table.dto.ts
```diff
@@ -45,7 +45,11 @@ export class MatchTypeConstraint implements ValidatorConstraintInterface {
   }
 
   matchType(value, realtedType) {
```
akshaysasidrn, line 47: related*

akshaysasidrn (approved):

PR #5057, opened 2022-12-26
[ Bug fixed ]  : input field width is not fixed

manishkushare:
Resolves :

1. The width of the Input fields changes according to the size of the content in restApi , stripe and openAPI

image (2)

2. Spacing between rest API method input field and URL input field

Screenshot 2022-12-28 at 2 12 14 PM

3. On the focus, in the input fields in restAPI, openAPI and stripe, content is shifting to right. Also, on focus, removed applied border-radius to make it consistent with the default state
[Untitled app - Tooljet - 28 December 2022 - Watch Video](https://www.loom.com/share/582b3fee5a144361b701895ed2a92205)


sherfin94 (changes requested):
Hey @manishkushare , please check RestAPI data source, there is overlap between query method and url box.

Also, if it wouldn't take too much of effort, please see if you can make the right side border radius of the query method input box `0` instead of rounded.  So that it looks like this:

image

sherfin94 (approved):

PR #5069, opened 2022-12-28
[Fix] :: Header icon/buttons not visible on preview for newly created apps

vjaris42:
Resolves: 
when no components are added or an empty app is launched the icon is hidden even if we don't change the header setting


shubh22:
Task linked: [CU-85zrgw2n4 Logo is missing in app preview](https://app.clickup.com/t/85zrgw2n4)

arpitnath (approved):

PR #5096, opened 2022-12-29
[ Bug fixed ] : It should be Query Parametres instead of Query parameters

manishkushare:
Resolves :

Replace parameters with Parameters

Screenshot - 2022-12-27T184153 869


shubh22:
Task linked: [CU-85zrgyp9m It should be "Query Parametres" instead of Query parameters.](https://app.clickup.com/t/85zrgyp9m)

kavinvenkatachalam (approved):

PR #5098, opened 2022-12-29
[ Bug fixed ] : UI is not consistent when query name is large in query panel header

manishkushare:
Resolves :
In the query panel header, when the query name is too long, it overlaps with buttons present in the query panel header

Screenshot - 2022-12-28T112709 641

### DIFF
--- a/frontend/src/Editor/QueryManager/QueryManager.jsx
+++ b/frontend/src/Editor/QueryManager/QueryManager.jsx
@@ -578,7 +578,11 @@ class QueryManagerComponent extends React.Component {
                   </svg>
                 </span>
                 <div className="query-name-breadcrum d-flex align-items-center">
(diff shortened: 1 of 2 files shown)


shubh22:
Task linked: [CU-85zrgzcdm UI is not consistent when query name is large.](https://app.clickup.com/t/85zrgzcdm)

kavinvenkatachalam (changes requested):
@manishkushare Instead of substring we can do it in a simple CSS as follows: 

`width: 100px;
    white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;`

Adjust the width as per your need

kavinvenkatachalam (changes requested):
`Input` field is not looking good. Please take a look

image

image

kavinvenkatachalam (approved):

PR #5102, opened 2022-12-29
[Improvement] tooljet query - delete operation

arpitnath:
Resolves : Invalidate filters with invalid values on query manager delete rows option


shubh22:
Task linked: [CU-85zrgywrz Invalidate filters with invalid values on query manager delete rows option](https://app.clickup.com/t/85zrgywrz)

gondar00 (approved):

PR #5116, opened 2022-12-29
[ Bug fixed ]: GraphQL headers  url paramaters key  value UI is not as expected

manishkushare:
Resolves: GraphQL headers & URL parameters key & value UI is not as expected.

Screenshot - 2022-12-27T123545 804


shubh22:
Task linked: [CU-85zrgxvum GraphQL headers & url paramateres key & value UI is not as expected.](https://app.clickup.com/t/85zrgxvum)

kavinvenkatachalam (changes requested):
@manishkushare The button size is small when compared to the inputs. It looks bit odd. Can we make it look better?

image

kavinvenkatachalam (approved):

PR #5123, opened 2022-12-30
fix: able to delete the only version

### LINKED ISSUE #4978: able to delete the only version in the app

Currently, when there is only single version on the app, we are able to delete it.

Ideally, the user should not be able to delete the last/only version in the app

1. Create an app with a version
2. delete the version from the version manager

### Screenshots or Screencast

https://user-images.githubusercontent.com/25361949/207345333-5ef2bb3f-9720-4e4b-827e-c7c1b0ea77bc.mov

mikebarr24:
fixes #4978


server/src/controllers/apps.controller.ts
```diff
@@ -284,8 +284,12 @@ export class AppsController {
   @UseGuards(JwtAuthGuard)
   @Delete(':id/versions/:versionId')
   async deleteVersion(@User() user, @Param('id') id, @Param('versionId') versionId) {
+    const number = await this.appsService.fetchVersions(user, id);
     const version = await this.appsService.findVersion(versionId);
     const ability = await this.appsAbilityFactory.appsActions(user, id);
+    if (number.length <= 1) {
```
gsmithun4, line 290: Shall we add this after the ability checks?
https://github.com/ToolJet/ToolJet/pull/5123/files#diff-4cc88c767a6288a89c3d84175bac75d447c6fda36b23275ffcd88f323d2e979bR294

gsmithun4 (changes requested):

server/src/controllers/apps.controller.ts
mikebarr24: I think that should be it now.

gsmithun4 (dismissed):

gsmithun4:
@mikebarr24 Please resolve conflicts

mikebarr24:
Sorry. Don't think this is right and don't have time just now to fix.

shubh22:
Task linked: [CU-85zrht8vb Able to delete all the versions. and the app is crashing when exporting the app without any version.](https://app.clickup.com/t/85zrht8vb)

PR #5167, opened 2023-01-03
[V2-Beta] Fix: Bug on query create under two data sources in same kind

gsmithun4:


shubh22:
Task linked: [CU-85zrhbuj4 Query creation bug if created under datasources with same kind](https://app.clickup.com/t/85zrhbuj4)

kavinvenkatachalam (approved):

PR #5178, opened 2023-01-04
Viewer - Hide header for desktop mode when header is set to hidden

arpitnath:
Resolves: hide header for desktop mode when header is set to hidden


sherfin94 (approved):
Hey @arpitnath, can we adjust the

PR #5183, opened 2023-01-04
Bugfix :: Removed testimonial screen from onboarding pages

stepinfwd:
fix 
* removed testimonial section from every screen
* moved logo to center


kavinvenkatachalam (changes requested):
@stepinfwd As discussed, logo seems to be very small. Check with Jassem and update if it needs a diff image

image

kavinvenkatachalam (approved):

PR #5197, opened 2023-01-04
v2-improvement: select the first group by default and render the selected on by default

arpitnath:
resolves: select the first group by default and render the selected on by default


frontend/src/ManageGroupPermissions/ManageGroupPermissions.jsx
```diff
@@ -328,9 +329,14 @@ class ManageGroupPermissionsComponent extends React.Component {
                             ))}
                           </tbody>
                         ) : (
-                          <tbody>
+                          <tbody className={'user-group-table '}>
```
gsmithun4, line 332:
```suggestion
                          <tbody className="user-group-table">
```

kavinvenkatachalam (approved):

PR #5203, opened 2023-01-05
Bugfix :: Fixing scroll in dashboard and databases

stepinfwd:
*scroll was available even when dashboard has lesser apps similar in databases


frontend/src/_styles/theme.scss
```diff
@@ -4358,6 +4358,7 @@ input[type="text"] {
  */
 .folder-list {
   color: #292d37;
+  overflow-y: scroll;
```
gondar00, line 4361:
```suggestion
  overflow-y: auto;
```

stepinfwd: fixed @gondar00

shubh22:
Task linked: [CU-85zrhfuha The vertical scroll bar on the right hand panel does not work as expected](https://app.clickup.com/t/85zrhfuha)

kavinvenkatachalam (approved):

PR #5205, opened 2023-01-05
Bugfix :: Onboarded data saved to telemetry 

stepinfwd:
*onboarded data saving


frontend/src/_services/authentication.service.js
```diff
@@ -206,7 +224,7 @@ function resetPassword(params) {
 function logout() {
   clearUser();
   const loginPath = (window.public_config?.SUB_PATH || '/') + 'login';
-  window.location.href = loginPath + `?redirectTo=${window.location.pathname}`;
```
gsmithun4, line 209: is it related to this PR?

gsmithun4 (changes requested):

server/src/services/auth.service.ts
```diff
@@ -335,6 +337,10 @@ export class AuthService {
       await this.organizationUsersService.create(user, organization, false, manager);
       return this.generateLoginResultPayload(user, organization, false, true, manager);
     });
+
+    await this.metadataService.finishOnboarding(name, email, companyName, companySize, role);
```
gsmithun4, line 341: what if this url `https://hub.tooljet.io/subscribe` is not reachable from the instance? should add a try catch for finish onboarding function
https://github.com/ToolJet/ToolJet/pull/5205/files#diff-61222a9e424956861d68b088bbb08a464f1fa5d8a8c8fa7a3cfa116ae1f8170fR42

gsmithun4 (changes requested):

server/src/services/auth.service.ts
gsmithun4: and can call this function asynchronously

akshaysasidrn (approved):

PR #5230, opened 2023-01-06
fix: app card height while no edit access for user

gondar00:
- fix: app card height while no edit access for user
- fix: pagination width issue in mozilla browser


shubh22:
Task linked: [CU-85zrhg585 All the app cards should have the same padding.](https://app.clickup.com/t/85zrhg585)

arpitnath (approved):

PR #5252, opened 2023-01-09
Bugfix :: navbar in verification success screen

stepinfwd:
included navbar in verification success screen
fixed password overflow above eye button


kavinvenkatachalam (changes requested):
@stepinfwd Adding background to the eye icon affect the dark mode as well the when the saved password is used. Attaching the screenshot

image

image

kavinvenkatachalam (approved):

PR #5247, opened 2023-01-06
Multipage: component and current state is persisted for duplicate pages

arpitnath:
fixes: persisted state for components from duplicate pages

PR #5258, opened 2023-01-09
fixes: app crash for multi pages editor

arpitnath:
follow up: #5247


arpitnath:
Ref: https://github.com/ToolJet/ToolJet/pull/5260

PR #5259, opened 2023-01-09
[Bug]: Fixes Redirection to Dashboard On Switching Workspace

kavinvenkatachalam:
Fixes - When we switch the workspace from database page, it redirects to the dashboard instead of staying in the same page


akshaysasidrn (approved):
lgtm

PR #5263, opened 2023-01-09
fix: pick organization id from param for internal storage

gondar00:


server/src/controllers/tooljet_db.controller.ts
```diff
@@ -25,64 +24,54 @@ export class TooljetDbController {
...
+    return this.postgrestProxyService.perform(req, res, next, organizationId);
   }
 
   @Get('/:organizationId/tables')
   @UseGuards(TooljetDbGuard)
   @CheckPolicies((ability: TooljetDbAbility) => ability.can(Action.ViewTables, 'all'))
-  async tables(@User() user, @Param('organizationId') organizationId) {
-    const result = await this.tooljetDbService.perform(user, organizationId, 'view_tables');
+  async tables(@Param('organizationId') organizationId) {
+    const result = await this.tooljetDbService.perform(organizationId, 'view_tables');
```
gsmithun4, line 35: We should check
`user.organizationId === organizationId`
ability check is done using `user.organizationId`

gsmithun4: Same for all routes

gsmithun4 (changes requested):

gsmithun4 (approved):

PR #5288, opened 2023-01-10
[runpy] support to import external libraries

### LINKED ISSUE #5438: Runpy: Support to import external libraries
Moving to jsDelivr as a CDN for Pyodide could potentially improve its performance and reliability. jsDelivr is a widely used, highly reliable, and fast Content Delivery Network (CDN) that can provide faster and more consistent loading times for Pyodide's packages. Additionally, jsDelivr supports automatic failover and geographical load balancing, which can further improve the reliability of Pyodide's package delivery. Furthermore, using a widely used and well-established CDN like jsDelivr can also help ensure that Pyodide's packages are delivered even during times of heavy traffic or high demand.

The modules that are not currently supported in Pyodide are those that have C or C++ extensions that rely on system libraries. These modules cannot be used in Pyodide because it runs in a web browser, which does not have access to the underlying system libraries that the C or C++ extensions rely on. Additionally, Pyodide uses a version of Python that has been compiled to WebAssembly, which does not support the same system calls as a regular version of Python. Therefore, any module that requires access to system libraries or system calls will not work in Pyodide.
ref: https://github.com/pyodide/pyodide/issues/3034#issuecomment-1345496026

arpitnath:
Resolves #5438 
adds support to import external libraries


emidhun:
@arpitnath Completed the testing.
Please have a look into the libraries which are not working.

frontend/src/_helpers/utils.js
```diff
@@ -3,7 +3,7 @@ import moment from 'moment';
 import _ from 'lodash';
 import axios from 'axios';
 import JSON5 from 'json5';
-import urlJoin from 'url-join';
+// import urlJoin from 'url-join';
```
gondar00, line 6: This line can be removed

arpitnath:
The modules that are not currently supported in Pyodide are those that have C or C++ extensions that rely on system libraries. These modules cannot be used in Pyodide because it runs in a web browser, which does not have access to the underlying system libraries that the C or C++ extensions rely on. Additionally, Pyodide uses a version of Python that has been compiled to WebAssembly, which does not support the same system calls as a regular version of Python. Therefore, any module that requires access to system libraries or system calls will not work in Pyodide.
Ref: #5438

gondar00 (approved):

PR #5291, opened 2023-01-11
[improvement] Renamed the column read_on_dashboard to hide_from_dashboard 

### LINKED ISSUE #5292: [improvement] Rename the column name read_on_dashboard to hide_from_dashboard (avoid future confusion about the feature)

Currently, the hide from dashboard feature is working when the user turns on the feature from the permissions tab. but the value of that checkbox is stored in the column  `read_on_dashboard`. so that will confuse us in the future like we are storing the value `true` and hiding it from the dashboard  while the column name itself `read_on_dashboard`

I think changing the column name hide_from_dashboard will be a simple solution.

Yes

shah21:
Resolves #5292


gsmithun4:
@shah21 please rebase

gsmithun4 (approved):

PR #5320, opened 2023-01-12
[ Feature added ] :  [Table widget ] adding CSA to deselect the row

### LINKED ISSUE #5318: [ Feature Request ] : [ Table widget ] adding CSA to un select the row
Environment
Development

Version
v2.0.0-beta

No, it's not related to any bug.
In the current implementation, we have component-specific action for the Table widget to select the row. But, we don't have any action to unselect the selected row.

We can add one more action for the Table widget, where we can allow users to provide the key and value as parameters and if these key and value params are valid (means, this key, and value are associated with the row which is selected already) and if they are valid we can un select the selected row.

1. Create any new application
2. In the created application, drag the Table, and the button widget from the right sidebar onto the canvas.
4. Click on the button widget, to open the widget inspector for the button at the right sidebar itself.
5. Look for the Events section in the button widget inspector, and add one event or you can choose an existing event if any presents.
6. Once you click on the newly added event card or existing one, an event manager pop-up will appear to it's left side of the card.
7. In that, you will find the select input field for `Event`, for the same choose `onClick` event
8. Next, below that you will find the select input field for choosing `Action`, select `Control Component` present at the bottom of the options list.
9. Next will be `Component` choose the Table widget and after that, there will be one more time Action field, which will have two options, we have to focus and select `Select row`
10. After that, in order to select a particular row from the table data, we need to provide parameters such as key and value to select the particular row. For our simplicity, we have default data for the Table widget. Let's say out of that we have to select the row, whose id is 2. To do that, we need to provide `id` as a key and `{{2}}` as a value. Once done following till now, you can click anywhere on the canvas to close the event manager pop-up, as we are done with setting the event on button widget which will control the selection of rows in the Table widget.
11. Now, just to make life easy for us, let's click on the table widget to open the widget inspector in the right sidebar and look for the `options` section, in that section find for `highlight selected row` option and set it true or toggled it ON. So, that, once our row will be selected, will be able to distinguish it from the remaining rows.
12. Now, click on the button widget, as we added an event on the button widget, in the previous steps to select the row whose id is 2, you can observe the row with id 2 got selected/ highlighted. You can also verify that from the inspector present in the left sidebar. To do that open the inspector, look for components, and inside that, look for Table <name of your table widget>, inside that look for selectedRow, it will show all the data of the row which is selected.
13. Now, Similar to this we do not have any action to unselect the selected row. That's our problem statement here.
To have an action called UnSelect row, which will accept parameters as key and value, and if those are valid, when compared with the selected row, then we should be able to unselect it.

Yes

manishkushare:
Resolves #5318

### DIFF
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -499,6 +499,20 @@ export function Table({
     },
     [JSON.stringify(tableData), JSON.stringify(tableDetails.selectedRow)]
   );
+  registerAction(
+    'discardChanges',
+    async function () {
+      if (Object.keys(tableDetails.changeSet || {}).length > 0) {
+        setExposedVariables({
+          changeSet: {},
+          dataUpdates: [],
+        }).then(() => {
+          mergeToTableDetails({ dataUpdates: {}, changeSet: {} });
+        });
+      }
+    },
+    [JSON.stringify(tableData), JSON.stringify(tableDetails.changeSet)]
+  );
 
   useEffect(() => {
     const selectedRowsOriginalData = selectedFlatRows.map((row) => row.original);
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -486,6 +486,20 @@ export function Table({
     },
     [JSON.stringify(tableData), JSON.stringify(tableDetails.selectedRow)]
   );
+  registerAction(
+    'UnselectRow',
+    async function (key, value) {
+      if (key && value && tableDetails.selectedRow.hasOwnProperty(key) && tableDetails.selectedRow[key] === value) {
+        const selectedRowDetails = { selectedRow: {}, selectedRowId: {} };
+        mergeToTableDetails(selectedRowDetails);
+        setExposedVariables(selectedRowDetails).then(() => {
+          fireEvent('onRowClicked');
+        });
+      }
+      return;
+    },
+    [JSON.stringify(tableData), JSON.stringify(tableDetails.selectedRow)]
+  );
 
   useEffect(() => {
     const selectedRowsOriginalData = selectedFlatRows.map((row) => row.original);
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -489,8 +489,8 @@ export function Table({
   );
   registerAction(
     'deselectRow',
-    async function (key, value) {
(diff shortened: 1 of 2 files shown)


shubh22:
Task linked: [CU-85zrj01gc [Table widget] add csa to unselect the row](https://app.clickup.com/t/85zrj01gc)

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -486,6 +487,24 @@ export function Table({
     },
     [JSON.stringify(tableData), JSON.stringify(tableDetails.selectedRow)]
   );
+  registerAction(
+    'UnselectRow',
```
sherfin94, line 491:
```suggestion
    'deselectRow',
```

frontend/src/Editor/WidgetManager/widgetConfig.js
```diff
@@ -378,6 +378,14 @@ export const widgets = [
           { handle: 'value', displayName: 'Value' },
         ],
       },
+      {
+        handle: 'UnselectRow',
```
sherfin94, line 382:
```suggestion
        handle: 'deselectRow',
```

frontend/src/Editor/WidgetManager/widgetConfig.js
```diff
@@ -378,6 +378,14 @@ export const widgets = [
...
+        displayName: 'UnSelect row',
```
sherfin94, line 383:
```suggestion
        displayName: 'Deselect row',
```

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -486,6 +487,24 @@ export function Table({
...
+    async function (key, value) {
+      if (key && value && tableDetails.selectedRow.hasOwnProperty(key) && tableDetails.selectedRow[key] === value) {
+        const selectedRowDetails = { selectedRow: {}, selectedRowId: {} };
+        mergeToTableDetails(selectedRowDetails);
+        setExposedVariables(selectedRowDetails).then(() => {
+          fireEvent('onRowClicked');
```
sherfin94, line 497: There is no need to fire `onRowClicked`.

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -486,6 +487,24 @@ export function Table({
...
+        });
+      }
+      if (key && value && (!tableDetails.selectedRow.hasOwnProperty(key) || tableDetails.selectedRow[key] !== value)) {
+        toast.error('The row with given key and value was not selected');
```
sherfin94, line 501: Lets not fire any toast

sherfin94 (changes requested):
Suggested a few changes. Please take a look.

sherfin94 (approved):

PR #5342, opened 2023-01-16
[Bugfix] app crashes when delete op is ran for tj db query

### LINKED ISSUE #5341: [Tooljet database query] Running delete operation crashes the app
For a failed query an error toast should be fired with an error log in the debugger.

Delete query fails, and the app crashes

1. Add a tj database query
2. select the delete operation 
3. add filter fields and leave the limit empty

arpitnath:
Resolves #5341 

fixes: Validation of Limit input field


shubh22:
Task linked: [CU-85zrgrfxv Validation of Limit input field is missing](https://app.clickup.com/t/85zrgrfxv)

akshaysasidrn (approved):

PR #5343, opened 2023-01-16
Fixed default text widget bg color

### LINKED ISSUE #4207: Text widget: Default value for background color

Aviroop-001:
Resolves #4207


kavinvenkatachalam (changes requested):
@Aviroop-001 It add value to the property. We already have a `styles` key and `backgroundColor` inside it. The value of the backgroundColor is empty. You need to add the value `#fff00000` to it. Attaching the sceenshot,
cc: @arpitnath 
image

kavinvenkatachalam (changes requested):
Hey @Aviroop-001,

Can you make the suggested changes?

withshubh:
Hey @Aviroop-001 - Please update the PR with the changes suggested.

kavinvenkatachalam (approved):

PR #5362, opened 2023-01-18
[Fix] :: Copy to clipboard not visible in inspector after visiting Page panel from sidebar

### LINKED ISSUE #5361: [Bug] :: Open and reopen Page-panel, hides the copy button from the inspector.
### Version Information
V2

### Environment
MacOS

Copy path icon should be visible in the inspector if `enableCopyToClipboard` is passed as prop

`enableCopyToClipboard` is not being considered on initial render hence it is only shown once the component re-renders which is not certain at times.

Open and close the page panel and verify the copy icon on the inspector.

Yes

vjaris42:
Resolves #5361


shubh22:
Task linked: [CU-85zrht883 Open and reopen Page-panel, hides the copy button from the inspector.](https://app.clickup.com/t/85zrht883)

arpitnath (approved):

PR #5368, opened 2023-01-18
[ Bug fixed ] - [ Table widget ] showing blank spaces for undefined and null values in the respective cells

### LINKED ISSUE #5367: [ Bug ] : [ Table ] cells displaying undefined or null values as well instead of blank spaces
### Version Information
v2.0.0-beta
### Environment
Development

In the Table widget, for the respective cell, whose value is `undefined` or `null` should display blank spaces in that particular cell

In the current deployment, `undefined` or `null` values are displayed as it is in the individual cell.

1. Create a new application
2. Drag the Table widget from the right sidebar on the canvas.
3. Click anywhere on the Table widget or on the label of the table widget to open the widget inspector on the right sidebar.
4. Look for the `Table data ` column in the properties in the widget inspector. We have provided the default data in the same field.
5. In the data, any one key from the data with a value as `undefined` (undefined data type). For simplicity, let's change the first value of the `name` key in the first object to `undefined`
6. Once, done with the above step, please take a look at the table widget, in the first row, for the name column, the cell value is showing as undefined.
Which looks odd. That's our problem statement, we need to replace undefined and null values with blank spaces.

### Screenshots or Screencast

Screenshot 2023-01-18 at 5 00 44 PM

Yes

manishkushare:
Resolves #5367

### DIFF
--- a/frontend/src/Editor/Components/Table/columns/index.jsx
+++ b/frontend/src/Editor/Components/Table/columns/index.jsx
@@ -68,7 +68,7 @@ export default function generateColumnsData({
       isEditable: column.isEditable,
       Cell: function (cell) {
         const rowChangeSet = changeSet ? changeSet[cell.row.index] : null;
-        const cellValue = rowChangeSet ? rowChangeSet[column.name] || cell.value : cell.value;
+        let cellValue = rowChangeSet ? rowChangeSet[column.name] || cell.value : cell.value;
         const rowData = tableData[cell.row.index];
 
         if (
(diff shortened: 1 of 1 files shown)


shubh22:
Task linked: [CU-85zrjekmw [ Bug ] : [ Table ] cells displaying undefined or null values as well instead of blank spaces](https://app.clickup.com/t/85zrjekmw)

kavinvenkatachalam (approved):

PR #5383, opened 2023-01-20
[Fix] :: Check for subpaths in openapi

### LINKED ISSUE #5022: OpenAPI endpoints get decamelized
### Version Information
v1.30.0

### Environment
Docker / Linux 

An OpenAPI endpoint such as "/Path/With/Uppercase/Names"  should call that path with uppercase path elements preserved.

The response from the server get processed by decamelizedKeys()  so the the endpoint becomes "/_p_ath/_w_ith/_u_ppercase/_n_ames".

1. Paste an OpenAPI .json with an endpoint such as "/Path" into create datasource dialog
2. Create query using new datasource.
3. When choosing endpoint the options will show "/_p_ath"

src/controllers/data_sources.controller.ts:80

vjaris42:
Resolves #5022

### DIFF
--- a/server/src/controllers/data_sources.controller.ts
+++ b/server/src/controllers/data_sources.controller.ts
@@ -56,7 +56,11 @@ export class DataSourcesController {
         );
       }
     }
-    return decamelizeKeys({ data_sources: dataSources });
+    return decamelizeKeys({ data_sources: dataSources }, function (key, convert, options) {
+      //eslint-disable-next-line
+      const checkForKeysAsPath = /^(\/{0,1}(?!\/))[A-Za-z0-9\/\-_]+(\.([a-zA-Z]+))?$/gm;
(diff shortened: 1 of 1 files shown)


shah21 (dismissed):
LGTM.

server/src/controllers/data_sources.controller.ts
```diff
@@ -56,7 +56,11 @@ export class DataSourcesController {
         );
       }
     }
-    return decamelizeKeys({ data_sources: dataSources });
+    return decamelizeKeys({ data_sources: dataSources }, function (key, convert, options) {
+      //eslint-disable-next-line
```
gsmithun4, line 60: Shall we add a comment why lint check is disabled?

gsmithun4 (approved):

PR #5023, opened 2022-12-20
[feature] :: Datasource plugin for Azure blob storage

vishnurk6247:
closes #4484

PR #5398, opened 2023-01-23
[plugin]  Datasource plugin for Azure blob storage

arpitnath:
ref: #5023


akshaysasidrn (commented):
Changes look good, can we test them out?

akshaysasidrn (approved):

Issue #5447, opened 2023-01-30
Link widget causes recursion in the Inspector and fails


caerwynj:
### Version Information

tooljet/try:latest as of today

### Environment
Docker

### What is the expected behaviour?
Add  Link widget to the canvas and view the component in the Inspector

### What is the current behaviour?
Inspector displays "Something went wrong"
Error console shows: Range Error: Maximum call stack size exceeded

### How to reproduce the issue?

1.Create new application
2.Drag Link widget onto canvas
3.Open Inspector

### Any possible solutions?
Could it be recursion on the onClick event?

PR #5449, opened 2023-01-31
[Bug]: Fixed app crashes when items dropped into container inside listview

### LINKED ISSUE #5448: [Bug]: App Frozen When Items Dropped into Container Inside ListView
### Environment

ToolJet Cloud

Should be able to drop any widgets inside listview or container inside listview

The app is forzen when we drop the items to the container inside listView

1. Add a list view.
2. Delete all default components.
3. Add a container inside the List view.
4. Add a button inside the container. (or any other component/ button)

`frontend/src/Editor/Box.jsx`

Yes

kavinvenkatachalam:
Resolves #5448 #5447


shubh22:
Task linked: [CU-85zrhp91y App editor is Frozen.](https://app.clickup.com/t/85zrhp91y)

sherfin94 (approved):

Issue #5400, opened 2023-01-23
Add DTO validation for update organization API


gsmithun4:
### Describe the solution you'd like
Create DTO validation for update organization API
[PATCH] /api/organizations

name - Optional, string, max length 25
domain - optional, string, max length 250
enableSignUp - Optional, boolean
inheritSSO - Optional, boolean

### If the feature is approved, would you be willing to submit a PR?
No

Jittojoyes98:
Can I work on this issue?

withshubh:
Hey @Jittojoyes98 Assigned to you.

Jittojoyes98:
Heyy i have updated the file 
![Screenshot from 2023-01-31 11-36-19](https://user-images.githubusercontent.com/51293782/215679911-1cabd052-ca92-4e6e-a3a9-6eabcf2f9c61.png)
For the optional parameter does the (?) required along with the name of the field.

gsmithun4:
> Heyy i have updated the file ![Screenshot from 2023-01-31 11-36-19](https://user-images.githubusercontent.com/51293782/215679911-1cabd052-ca92-4e6e-a3a9-6eabcf2f9c61.png) For the optional parameter does the (?) required along with the name of the field.

Not mandatory, good to add. 
can use `@Transform(({ value }) => sanitizeInput(value))` for domain also
for name, shall we add @IsOptional followed by @IsString

PR #5456, opened 2023-02-01
Update dto API completed

Jittojoyes98:
Resolves https://github.com/ToolJet/ToolJet/issues/5400


server/src/dto/organization.ts
```diff
@@ -0,0 +1,34 @@
+import { Transform } from 'class-transformer';
```
gsmithun4, line 1: Shall we rename this file to `organization.dto.ts`

gsmithun4 (changes requested):

server/src/dto/organization.ts
Jittojoyes98: yeah ok, will fix

server/src/dto/organization.dto.ts
```diff
@@ -0,0 +1,34 @@
...
+export class OrganizationCreateDto {
+  @IsString()
+  @Transform(({ value }) => sanitizeInput(value))
+  @IsNotEmpty()
+  @MaxLength(25, { message: 'Name cannot be longer than 25 characters' })
+  name: string;
+}
+
+export class OrganizationUpdateDto {
+  @IsString()
```
gsmithun4, line 15: shall we change the order?
```
  @IsOptional()
  @IsString()
```
for all fields

gsmithun4 (changes requested):

gsmithun4 (approved):

gsmithun4:
Hi @Jittojoyes98 , the messages added is not showing in front end, please get the message from error object and set the toast
https://github.com/ToolJet/ToolJet/blob/improvement/workspace-id-scope-be/frontend/src/ManageSSO/GeneralSettings.jsx#L34
```
(err) => {
        setSaving(false);
        toast.error(err?.data?.message || 'Error while saving SSO configurations', {
          position: 'top-center',
        });
      }
```

gsmithun4:
Same here: https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/_components/OrganizationManager/EditOrganization.jsx#L25

gsmithun4 (changes requested):
display error messages

Jittojoyes98:
In the second toast message should i need to place somewhere, like the other toast which has the position on the  top.

gsmithun4 (approved):

PR #5461, opened 2023-02-01
[bug-fix] The public app is redirecting to the login page if the user token expired. 

### LINKED ISSUE #5462: The public app is redirecting to the login page when the user token expires
### Version Information
Latest
### Environment

All

Should not redirect to the login page. users should be able to see a public app without authentication 

The public app is redirecting to the login page when the user token expires

1. Create an app with queries and make it public (attach any query with some widgets also)
2. release the app. and run the sharable link on a private browser window
3. observe 

Yes

shah21:
Resolves #5462

### DIFF
--- a/frontend/src/App/App.jsx
+++ b/frontend/src/App/App.jsx
@@ -38,14 +38,12 @@ class App extends React.Component {
   }
 
   fetchMetadata = () => {
-    if (this.state.currentUser) {
-      tooljetService.fetchMetaData().then((data) => {
-        localStorage.setItem('currentVersion', data.installed_version);
-        if (data.latest_version && lt(data.installed_version, data.latest_version) && data.version_ignored === false) {
-          this.setState({ updateAvailable: true });
-        }
-      });
-    }
+    tooljetService.fetchMetaData().then((data) => {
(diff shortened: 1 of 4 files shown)


frontend/src/App/App.jsx
```diff
@@ -37,6 +37,25 @@ class App extends React.Component {
...
+      // call apps api and get the app type
+      appService.getAppBySlug(paths[length]).then((app) => {
+        if (app && !app.is_public) {
+          if (this.state.currentUser) {
+            this.fetchMetadata();
+          }
+        }
+      });
+    } else {
+      this.fetchMetadata();
```
gsmithun4, line 55: Shall we skip the metadata API call if route includes `applications` and add fetch metadata call in app viewer page?
- Advantage of this: we have the app details in app viewer page and can invoke fetch meta call if application is not public

shah21: ok, got it

gsmithun4 (changes requested):
Improvements suggested

gsmithun4 (approved):

gsmithun4 (approved):

PR #5475, opened 2023-02-03
Fix and modify common cypress specs for v2

ajith-k-v:
Fixes: #5441


shubh22:
Task linked: [CU-85zrkcp6a Fix and modify common cypress specs for v2](https://app.clickup.com/t/85zrkcp6a)

PR #5489, opened 2023-02-07
[BUG] Query event mapping to another query breaks on new version create

gsmithun4:


gsmithun4:
changing base branch to main for hot fix

PR #5548, opened 2023-02-13
Implementation of bounded box component

### LINKED ISSUE #5581: Feature request to add a new bounded box widget
No, But, as of now there is no widget which allows us to customize image annotation

So, we can create a widget which will help us to create customize image annotation.
(Describe your proposed solution here.)

Yes

manishkushare:
Resolves #5581


shubh22:
Task linked: [CU-85zrmw0bv Feature request : Bounded Box widget](https://app.clickup.com/t/85zrmw0bv)

kavinvenkatachalam (changes requested):
@manishkushare Here are the few suggestions,

- I see `BoundedBox` is written as a class component. Can you make it as a functional component?
- Can you make the dropdown same size when we select the value initially and when I try to update the value again?
- Delete button will not be visible for some images. Can you have the same background as dropdown?
- The dropdown is not visible when we move out of the selected portion. Can we make it visible all the time?

kavinvenkatachalam (changes requested):
@manishkushare Code looks good! There are few suggestions in the UI and functionality,

1. `onChange` was not triggered for every action. For example, it should trigger whenever I draw the box, when I change the dropdown value, delete the selection etc
2. When I select point as a type, after selection I see the green box drawn instead of point
3. When I try to resize the widget, the image goes out of the boundary. Can you check that?
4. Can we compress the default image which we are using or a smaller size image?
5. Can we use `uuid` or something else instead of generating random number for the box ID? Random decimal number feels not user friendly
6. If possible try to reduce the padding of the select input

kavinvenkatachalam (changes requested):
@manishkushare Noticed few things,

1. If I decrease the widget size, the image goes out of the box
2. `onChange` event is triggered when I start drawing the box. It should happen only when I finish drawing it
3. One more thing I noticed is that exposed value has two diff key. One is `data` and another one is `geometry`. Can we make it as a single key so that it will be easy for the users to fetch the data

kavinvenkatachalam (approved):

PR #5612, opened 2023-02-21
[AppWrite] - Upgrade to latest version

byawitz:
Upgrade the AppWrite engine package from 5 to 8 As of today AppWrite required that the Database ID will be sent in each request. That is because today AppWrote projects can have more then on Database.


neozhu:
hello, this issue always exists, in my testing throw database was not found use appwrite

PR #5637, opened 2023-02-27
changes in dockerfile for non-sudo user

adishM98:
made changes for the dockerfile to build image using non-sudo user

### DIFF
--- a/docker/production.Dockerfile
+++ b/docker/production.Dockerfile
@@ -6,42 +6,33 @@ ENV NODE_OPTIONS="--max-old-space-size=4096"
 RUN npm i -g npm@8.11.0
 RUN mkdir -p /app
 
-# Create a non-sudo user and group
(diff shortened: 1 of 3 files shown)


docker/production.Dockerfile
```diff
@@ -6,33 +6,42 @@ ENV NODE_OPTIONS="--max-old-space-size=4096"
 RUN npm i -g npm@8.11.0
 RUN mkdir -p /app
 
+# Create a non-sudo user and group
+RUN addgroup --gid 1001 tooljetgroup \
```
akshaysasidrn, line 10: Wouldn't we have to make use of the user in the last stage give this is a multi user stage setup??

akshaysasidrn: I think what you can do is to define non sudo user in the last stage and then make that user own the copied directories from the prev stages.

akshaysasidrn (approved):

PR #5642, opened 2023-02-27
Add feature to programatically enable/disable, the make editable field in the tables column

### LINKED ISSUE #5404: [Table ] - Add option to allow a column to be editable dynamically

manishkushare:
Resolves #5404


kavinvenkatachalam:
It causes slowness of the app as we need to pass complete currentState to the table column. So, adding this PR `On Hold`

PR #5696, opened 2023-03-03
fixes cursor default to pointer

### LINKED ISSUE #5066: [icon component] cursor should become pointer when `onClick` event handler is added
When an onClick event is added to the icon component, the cursor doesn't change to a pointer which doesn't make it clear to the user that action can be performed on clicking the icon.

When an onClick event handler is added to the icon component, the cursor should become a pointer on hovering the icon so that it is apparent to the user that the action can be performed on clicking.

1. Add an icon component
2. Add an onClick event handler to icon
3. Preview the app and hover over the icon 

### Screenshots or Screencast
Dec-27-2022 15-16-49

shauryag2002:
fixes #5066 
fixes cursor default to pointer
before fix:
https://user-images.githubusercontent.com/77102885/222711581-09ce0c1f-1b03-4172-b80a-a6d797513fdf.mp4

after fix:
https://user-images.githubusercontent.com/77102885/222711635-43f6f63a-969a-4795-b408-c1ced4adc29f.mp4


frontend/src/_styles/theme.scss
arunava6M, line 1: Hey, please correct me if am wrong.
The pointer should come only when onClick prop has been passed to the icon component. But it seems with this change, every time the cursor will be pointer even if no onClick is passed.
Can you make this dependent on the onClick prop ?

kavinvenkatachalam (changes requested):
@shauryag2002 Thanks for the PR

The cursor should change to pointer only when the `icon` has `onClick` event. Can you please take a pull from latest develop before making changes.

frontend/src/_styles/theme.scss
```diff
@@ -5909,7 +5910,7 @@ input.hide-input-arrows {
     .popover-header {
       background-color: #232e3c;
       border-bottom: 1px solid #324156;
-      ;
+      cursor: pointer;
```
kavinvenkatachalam, line 5913:
```suggestion
```

kavinvenkatachalam (changes requested):
Adding `cursor: pointer` to the `popover-header` is not needed in this case. Can you please remove it?

kavinvenkatachalam (approved):

PR #5706, opened 2023-03-05
change width for column

Karmin24:
Changing the width of columns in a table


docker-compose.yaml
```diff
@@ -66,6 +66,7 @@ services:
       - 5432:5432
     volumes:
       - postgres:/data/postgres
+    user: postgres
```
sgaluza, line 69: I think this change is irrelevant to this PR and should be excluded

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -155,7 +155,7 @@ export function Table({
 
   const defaultColumn = React.useMemo(
     () => ({
-      minWidth: 60,
+      minWidth: 10,
```
sgaluza, line 158: This looks a bit aggressive for the width of other column types. Can we figure out a better way to reduce the image size instead of changing the default column width, which may affect other column types?

sgaluza:
@Karmin24 please review the comments and also please remove unrelated files from the PR (lock/json).

PR #5717, opened 2023-03-07
Feature: Add rows to tables

### LINKED ISSUE #269: Feature: Add rows to tables
The cells of tables can be edited but there isn't a way to add a new row.

manishkushare:
Resolves #269


sherfin94:
There seems to be some kind of conflict between adding rows and editing rows at the same time. Please check

sherfin94:
Putting newly added rows inside `changeSet` and `dataUpdates` can break backwards compatibility.

For example, those two values are often connected to `UPDATE` queries, whereas newly added rows will require `INSERT` queries.

Therefore, it will be safer to expose newly added rows as `newlyAddedRows`.

Please confirm with @shubh22 

cc: @kavinvenkatachalam

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -46,6 +46,64 @@ import GenerateEachCellValue from './GenerateEachCellValue';
...
+// utility function
+function removeAllAddedElements(array, addedElements) {
+  for (let i = 0; i < addedElements.length; i++) {
+    const index = array.indexOf(addedElements[i]);
+    if (index !== -1) {
+      array.splice(index, 1);
+    }
+  }
+}
+function utilityToUpdateTableDetails(
```
sherfin94, line 58: See if you can rename this function to an easily understandable name.

Also try to see if you can avoid mutation and use pure functions.

sherfin94 (changes requested):

PR #5721, opened 2023-03-08
Feature to support pagination in the listview widget

### LINKED ISSUE #5720: Feature to add pagination in the listview widget
In the listview component, we don't have pagination so that we can view the lists in the pages

Support the pagination in the listview widget

YEs

manishkushare:
Resolves #5720


frontend/src/Editor/Components/Listview.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { useRef, useState, useEffect } from 'react';
 import { SubContainer } from '../SubContainer';
 import _ from 'lodash';
+import { Pagination } from '../../_components/Pagination';
```
kavinvenkatachalam, line 4:
```suggestion
import { Pagination } from '@/_components/Pagination';
```

kavinvenkatachalam (changes requested):
@manishkushare Please find some suggestions below,

- App crashes when I try to select the listView on the older versions
- Scrollbar seems behind the pagination. Can you make it scrollable only till the thumb reaches pagination div?

frontend/src/Editor/Components/Listview.jsx
```diff
@@ -111,6 +130,19 @@ export const Listview = function Listview({
           </div>
         ))}
       </div>
+      {_.isArray(data) && enablePagination && (
```
kavinvenkatachalam, line 133:
```suggestion
      enablePagination && {_.isArray(data) && (
```

kavinvenkatachalam (changes requested):
@manishkushare Suggesting a very minor change,

1. Have a `enablePagination` check before check whether it is an array. In this way, array check will run only when the `enablePagination` toggle is turned on.
2. The scrollbar thumb should end above pagination div. Attaching the screenshot of table widget that's how it needs be.


image

kavinvenkatachalam (approved):

kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

kavinvenkatachalam (approved):

kavinvenkatachalam (approved):

Issue #4974, opened 2022-12-13
[number input] allows to set default value more than the max value


withshubh:
### What is the current behaviour?

In number-input component, when we set the max value and then set the default value more than the max value the component allows it to be set.

### What is the expected behaviour?

Ideally when max value is set, user should not be able to set the default value > max value

### How to reproduce the issue?

1. Add number input component on the canvas 
2. set max value 2000
3. set default value 4000

### Screenshots or Screencast

https://user-images.githubusercontent.com/25361949/207323893-636468be-e479-4a59-81fa-57b5de780aa7.MOV

harshmange44:
@withshubh Hey I would like to fix this.

AldrinoS:
Hi @withshubh, I would like to work on this issue

Aviroop-001:
I'd like to fix this issue. Please assign me the issue.

shubh22:
Assigned to you, @Aviroop-001

Ubaid-Manzoor:
@shubh22 Seem nobody has worked on it yet, please do assign it to me.

Sonali-Behera:
Hey, I would love to work on this issue. Please assign this to me.

withshubh:
Hey @Sonali-Behera - The PR for this issue is under review. We have already assigned an issue to you: https://github.com/ToolJet/ToolJet/issues/5981

MajidBabaeifar:
If it's still open, I'm eager to fix it (¬_¬ )

PriteshKiri:
Recreated the issue https://github.com/ToolJet/ToolJet/issues/10324

PriteshKiri:
@MajidBabaeifar you can comment on https://github.com/ToolJet/ToolJet/issues/10324 if you are still interested in this issue.

PR #5734, opened 2023-03-11
Merge Request - Fix default value out of bound issue #4974

Ubaid-Manzoor:
I have made 3 check
1. If the default value is < min or > max then set the default value to min/max respectively.
2. If the min value is changed and the default value < new min, then set the default value to the new min 
3. same as point 2 for when max value changes


Ubaid-Manzoor:
Branch name was not as per guidelines

PR #5735, opened 2023-03-12
Fix default value out of bound issue #4974

Ubaid-Manzoor:
I have made 3 check

If the default value is < min or > max then set the default value to min/max respectively.
If the min value is changed and the default value < new min, then set the default value to the new min
same as point 2 for when max value changes

### DIFF
--- a/frontend/src/Editor/Components/NumberInput.jsx
+++ b/frontend/src/Editor/Components/NumberInput.jsx
@@ -42,6 +42,13 @@ export const NumberInput = function NumberInput({
     if (value > maxValue) setValue(parseInt(properties.maxValue));
   }, [properties.maxValue]);
 
+  useEffect(() => {
+    if (!isNaN(value)) {
+      setExposedVariable('value', value);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [value]);
+
   const computedStyles = {
     height,
     display: visibility ? '' : 'none',
--- a/frontend/src/Editor/Components/NumberInput.jsx
+++ b/frontend/src/Editor/Components/NumberInput.jsx
@@ -15,45 +15,30 @@ export const NumberInput = function NumberInput({
 
   const [value, setValue] = React.useState(parseInt(properties.value));
 
-  const handleChange = (e) => {
+  const handleValueChange = (newValue) => {
     if (
       !isNaN(parseInt(properties.minValue)) &&
       !isNaN(parseInt(properties.maxValue)) &&
       parseInt(properties.minValue) > parseInt(properties.maxValue)
     ) {
       setValue(parseInt(properties.maxValue));
-    } else if (!isNaN(parseInt(properties.maxValue)) && parseInt(e.target.value) > parseInt(properties.maxValue)) {
+    } else if (!isNaN(parseInt(properties.maxValue)) && parseInt(newValue) > parseInt(properties.maxValue)) {
       setValue(parseInt(properties.maxValue));
-    } else if (!isNaN(parseInt(properties.minValue)) && parseInt(e.target.value) < parseInt(properties.minValue)) {
+    } else if (!isNaN(parseInt(properties.minValue)) && parseInt(newValue) < parseInt(properties.minValue)) {
       setValue(parseInt(properties.minValue));
     } else {
-      setValue(parseInt(e.target.value));
+      setValue(parseInt(newValue));
     }
+  };
+
+  const handleChange = (e) => {
+    handleValueChange(e.target.value);
     fireEvent('onChange');
   };
 
-  const handleMinValuChange = () => {
-    const { minValue } = properties;
-    if (value < minValue) setValue(parseInt(properties.minValue));
-  };
-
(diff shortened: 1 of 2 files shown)


kavinvenkatachalam (changes requested):
@Ubaid-Manzoor Thanks for the PR. 

`Inspector.jsx` is the common file for all the widget components. So, it is not recommended to write a logic to specify the value for the single widget. Can you please move the logic to `NumberInput.jsx` file which is specifc to Number Input widget?

Ubaid-Manzoor:
@kavinvenkatachalam It is not a NumberInput issue,
Let me be more clear
1. when we click on the number input it opens the popup on the right which I found was the Inspector component.
2. Now when I change a value(defaultValue/minValue/MaxValue) in the inspector element, not in the number input in the inspector element which opens in right It was not correct. And one more thing when I change the value in NumberInput it is working fine as there is already logic in the NumberInput component to set the correct value.

So if we change from NumberInput field It works fine,
but if it goes like Inspector --> NumberInput (Value in Inspector and NumberInput were both incorrect).

I AM TALKING ABOUT WHEN CHANGING HERE IT WAS NOT WORKING
Screenshot 2023-03-14 at 10 39 12 AM

Please let me know if I am wrong.

kavinvenkatachalam:
@Ubaid-Manzoor I'm very clear what's the issue is. We cannot make changes to the `Inspector.jsx` as it is common component for all the widget. The changes related to specific widget needs to be done on the paticular widget. In this case, have the same logic inside `NumberInput.jsx` and update the value if it is less than min and greater than max.

We already have a function `handleChange` which checks whether the value is within the range. Reuse the function and set the value of the number input state during the initial render and whenever there is a change in default, min and max value

withshubh:
fixes #4974

Ubaid-Manzoor:
@kavinvenkatachalam 
sure will do that

Ubaid-Manzoor:
Hi @kavinvenkatachalam ,
I have made changes in `NumberInput` but How can I make the properties rerender from inside of NumberInput, as NumberInput has only one state (value)?
Are there any props of NumberInput that I can use here?

kavinvenkatachalam:
@Ubaid-Manzoor There are no additional props. In the existing `useEffect`, add additional dependencies like `minValue` `maxValue` and write a logic inside it to check whether the default value is less than/greater than minValue or maxValue. If so, then set the `value` state to new value

Ubaid-Manzoor:
@kavinvenkatachalam Please check now.

kavinvenkatachalam (changes requested):
Hey @Ubaid-Manzoor,

When I drop the number input widget, the default value is not set. Attaching the screenshot.

image


Instead of creating a new function use the existing function as follows,

image

Ubaid-Manzoor:
sure @kavinvenkatachalam

Ubaid-Manzoor:
Done, but please let me know what the function `fireEvent('onChange')` does

kavinvenkatachalam:
> Done, but please let me know what the function `fireEvent('onChange')` does

It will trigger the `on change` event attached to the `NumberInput`

frontend/src/Editor/Components/NumberInput.jsx
```diff
@@ -15,26 +15,30 @@ export const NumberInput = function NumberInput({
...
+
+  const handleChange = (e) => {
+    handleValueChange(e.target.value);
     fireEvent('onChange');
   };
 
   useEffect(() => {
-    setValue(parseInt(properties.value));
-  }, [properties.value]);
+    handleValueChange(value);
```
kavinvenkatachalam, line 40:
```suggestion
    handleValueChange(properties.value);
```

kavinvenkatachalam (changes requested):
@Ubaid-Manzoor Requested a small change. We need to pass the new value to the `handleChange` function

frontend/src/Editor/Components/NumberInput.jsx
Ubaid-Manzoor: @kavinvenkatachalam, I have already tried using `properties.value`, but It was not working as expected. 
The Problem is `properties.value` is a `default` value, and whenever I change the default value it will change the `value in NumberInput` which is not correct.

And the default value is supposed to be set when there is no value right? like at the start of NumberInput creation.
And what you are saying is to pass the new value to `handleChange`, but the default value is not a new value, we will be normally changing the value which is the `state` of `NumberInput`.

Please correct me if I a wrong somewhere

kavinvenkatachalam: @Ubaid-Manzoor Agree with you. Can you remove `properties.value` from the dependency array below?

Ubaid-Manzoor: Done @kavinvenkatachalam

kavinvenkatachalam (approved):

Nandinisaha13:
As per the internal discussion,  error handling in the properties is on the roadmap. This PR might get affected by the change. So, adding on hold tag.

Issue #5749, opened 2023-03-14
`DB_HOST` is not getting introspected by the `DATABASE_URL` environment variable


rhuanbarreto:
### Version Information
Latest

### Environment
Tooljet on Docker

### What is the expected behaviour?
Given that the `DATABASE_URL` environment variable is set, there's no need to declare `PG_` variables once they come from the previous one. But at startup, I get an error:

```
tooljet-tooljet-1    | Error: you need to provide a host and port to test.
tooljet-tooljet-1    | Usage:
tooljet-tooljet-1    |     wait-for-it.sh host:port [-s] [-t timeout] [-- command args]
tooljet-tooljet-1    |     -h HOST | --host=HOST       Host or IP under test
tooljet-tooljet-1    |     -p PORT | --port=PORT       TCP port under test
tooljet-tooljet-1    |                                 Alternatively, you specify the host and port as host:port
tooljet-tooljet-1    |     -s | --strict               Only execute subcommand if the test succeeds
tooljet-tooljet-1    |     -q | --quiet                Don't output any status messages
tooljet-tooljet-1    |     -t TIMEOUT | --timeout=TIMEOUT
tooljet-tooljet-1    |                                 Timeout in seconds, zero for no timeout
tooljet-tooljet-1    |     -- COMMAND ARGS             Execute command with args after the test finishes
tooljet-tooljet-1 exited with code 1
```

### How to reproduce the issue?

This doesn't work:

```
version: '3'

services:
  tooljet:
    image: tooljet/tooljet-ce:latest
    ports:
      - 3000:3000
    environment:
      DATABASE_URL: postgres://postgres:postgres@postgres1/postgres
    command: npm run start:prod
  postgres1:
    image: postgres:13-alpine
    restart: always
    environment:
      - POSTGRES_PASSWORD=postgres
```

This works:

```
version: '3'

services:
  tooljet:
    image: tooljet/tooljet-ce:latest
    ports:
      - 3000:3000
    environment:
      DATABASE_URL: postgres://postgres:postgres@postgres1/postgres
      PG_HOST: postgres1
    command: npm run start:prod
  postgres1:
    image: postgres:13-alpine
    restart: always
    environment:
      - POSTGRES_PASSWORD=postgres
```

### Any possible solutions?

[...]

PR #5756, opened 2023-03-15
fix for database url

adishM98:
made changes in entrypoint.sh file
Added the info on env.example file
fixes - #5749 and #5768

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,6 +16,11 @@ PG_USER=<db username>
 PG_HOST=<db host>
 PG_PASS=<db password>
 
+
+#DATABASE CONFIG using string
+#Example: postgres://username:password@hostname:port/database?sslmode=disable
+
+
 # TOOLJET DATABASE
 ENABLE_TOOLJET_DB=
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -16,6 +16,12 @@ PG_USER=<db username>
 PG_HOST=<db host>
 PG_PASS=<db password>
 
+
+#DATABASE CONFIG using string
+# If you intent you use the string please use the below format
```
akshaysasidrn, line 21: * if the connection does not support ssl

adishM98: Have made the corrections.

server/ormconfig.ts
```diff
@@ -30,7 +48,7 @@ function buildConnectionOptions(data): TypeOrmModuleOptions {
     extra: {
       max: 25,
     },
-    ...sslConfig(data),
+    ...databaseurl(data),
```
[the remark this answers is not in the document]
adishM98: renamed

akshaysasidrn (approved):

PR #5784, opened 2023-03-20
Upon saving the  edited changes in the table, it is not mutating query object

### LINKED ISSUE #5501: Table edit and save changing the referenced data.

1. Add a table.
2. Use any queries and use the query data as the table data.
3. Make any of the table data editable.
4. Edit data and save.
5 Verify the query data from inspector.

### Screenshots or Screencast

https://user-images.githubusercontent.com/20367955/217478793-78a23f56-ad47-4b9e-980f-cf7ba3658420.mp4

manishkushare:
Resolves #5501


kavinvenkatachalam (dismissed):

kavinvenkatachalam (changes requested):
@manishkushare `useDeferredValue` is not working on React 17. We might need a solution that works for 17 as well

kavinvenkatachalam (changes requested):
@manishkushare Seems like `updatedData` in the inspector is getting reset after saving the changes. 

We're already cloning the table data in the function `handleExistingRowCellValueChange`, can you check the possibility of reusing or assigning the cloned data instead of cloning again?

kavinvenkatachalam (approved):

PR #5808, opened 2023-03-21
Fix for Tooljetdburl

adishM98:
Fix for not taking TOOLJET_DB default value tooljet_db when passed as url

### DIFF
--- a/server/scripts/database-config-utils.ts
+++ b/server/scripts/database-config-utils.ts
@@ -32,15 +32,21 @@ export function getEnvVars() {
 
 function buildDbConfigFromDatabaseURL(data): any {
   const config = buildDbConfigFromUrl(data.DATABASE_URL)
-  const TJDBconfig = buildDbConfigFromUrl(data.TOOLJET_DB_URL)
+  const TJDBconfig = data.TOOLJET_DB_URL ? buildDbConfigFromUrl(data.TOOLJET_DB_URL) : {};
+
+  // const config = data.TOOLJET_DB_URL ? buildDbConfigFromUrl(data.DATABASE_URL) : {};
+  // let TJDBconfig;
+  // if (data.TOOLJET_DB_URL) {
+  //   TJDBconfig = buildDbConfigFromUrl(data.TOOLJET_DB_URL)
+  // }
 
   const { value: dbConfig, error } = validateDatabaseConfig({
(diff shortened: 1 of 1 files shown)


server/scripts/database-config-utils.ts
```diff
@@ -21,17 +21,18 @@ export function getEnvVars() {
   if (fs.existsSync(envVarsFilePath)) {
     data = { ...data, ...dotenv.parse(fs.readFileSync(envVarsFilePath)) };
   }
+
   data = {
     ...data,
     ...(data.DATABASE_URL && buildDbConfigFromDatabaseURL(data)),
+    ...(data.TOOLJET_DB_URL && buildDbConfigFromDatabaseURL(data))
```
akshaysasidrn, line 28: we can combine the conditional check here:

`...((data.DATABASE_URL || data.TOOLJET_DB_URL) && buildDbConfigFromDatabaseURL(data))`

server/scripts/database-config-utils.ts
```diff
@@ -45,6 +46,24 @@ function buildDbConfigFromDatabaseURL(data): any {
     if (parsedUrl.port) config.port = parsedUrl.port;
   }
 
+  let TJDBconfig: any;
+  if (data.TOOLJET_DB_URL) {
+    const parsedTJDBUrl = url.parse(data.TOOLJET_DB_URL, false, true);
```
akshaysasidrn, line 51: can you extract url parsing into a separate function and reuse it here?

server/scripts/database-config-utils.ts
```diff
@@ -85,7 +103,7 @@ function validateDatabaseConfig(dbConfig: any): Joi.ValidationResult {
       PG_USER: Joi.string().required(),
       PG_DB: Joi.string().default('tooljet_production'),
       PG_DB_OWNER: Joi.string().default('true'),
-      ...(dbConfig.ENABLE_TOOLJET_DB === 'true' && {
+      ...((dbConfig.ENABLE_TOOLJET_DB === 'true' || dbConfig.TOOLJET_DB_URL) && {
```
akshaysasidrn, line 106: just checking the enable flag would be enough here right?

akshaysasidrn (changes requested):

akshaysasidrn (approved):

PR #5818, opened 2023-03-22
Allow configuration of MySQL connections via socket path

### LINKED ISSUE #5600: Datasource: connect to MySQL on Cloud SQL via Unix socket

The data we would like to process in ToolJet is stored in a restricted Cloud SQL instance.

In our setup, ToolJet runs on Google Cloud Run and we would like to be able to connect to a database running on Cloud SQL. Cloud Run connects to the Cloud SQL instance via the container settings. According to the [documentation](https://cloud.google.com/sql/docs/mysql/connect-run#connect_to), the connection should be established via Unix sockets.

The current UI to configure a MySQL connection only provides inputs for host, port, and database name.

image

Trying to replace the host with the socket connection string (`/cloudsql/...`) results in an `ENOTFOUND` error

Add an option to the MySQL configuration dialog that allows to set the socket connection details. Something like `/cloudsql/INSTANCE_IDENTIFIER` should be sufficient.

- Use a similar mechanism as for postgres as it seems this is supported to connect the ToolJet database
        See [ToolJet Setup | Deploying ToolJet on Google Cloud Run](https://docs.tooljet.com/docs/setup/google-cloud-run/#deploying-tooljet-application), list item 5, _Tip_:
    > If you are using [Public IP](https://cloud.google.com/sql/docs/postgres/connect-run) for Cloud SQL, then database host connection (value for `PG_HOST`) needs to be set using unix socket format, `/cloudsql/<CLOUD_SQL_CONNECTION_NAME>`.
- If this is already supported but not exposed via a UI, we could change the configuration in the table `data_source_options`
- Add an SSH tunnel to a machine on the same network as the Cloud SQL instance
- It is possible to connect to Cloud SQL instances via a [Cloud SQL connector](https://cloud.google.com/sql/docs/mysql/connect-connectors). I assume this also means additional integration efforts.

Yes

kariem:
Add `socket_path` property for MySQL plugin

- Add new property `socket_path`
- If `socket_path` is set, ignore host/port and ssl settings when setting up the connection

Resolves #5600

### DIFF
--- a/plugins/packages/mysql/lib/index.ts
+++ b/plugins/packages/mysql/lib/index.ts
@@ -66,16 +66,22 @@ export default class MysqlQueryService implements QueryService {
   }
 
   async buildConnection(sourceOptions: SourceOptions) {
+    // either use socket_path or host/port + ssl
+    const props = sourceOptions.socket_path
+      ? { socketPath: sourceOptions.socket_path }
+      : {
+          host: sourceOptions.host,
+          port: +sourceOptions.port,
+          ssl: sourceOptions.ssl_enabled ?? false, // Disabling by default for backward compatibility
+        };
     const config: Knex.Config = {
       client: 'mysql',
       connection: {
-        host: sourceOptions.host,
+        ...props,
         user: sourceOptions.username,
         password: sourceOptions.password,
         database: sourceOptions.database,
-        port: +sourceOptions.port,
         multipleStatements: true,
-        ssl: sourceOptions.ssl_enabled ?? false, // Disabling by default for backward compatibility
       },
     };
 
--- a/plugins/packages/mysql/lib/types.ts
+++ b/plugins/packages/mysql/lib/types.ts
@@ -2,6 +2,7 @@ export type SourceOptions = {
   database: string;
   host: string;
   port: string;
+  socket_path: string;
   username: string;
   password: string;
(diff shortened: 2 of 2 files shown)


withshubh:
Hey @kariem - Thanks for contributing to ToolJet. 🙌 
We will review the PR soon!

arpitnath (changes requested):
To build the connection form and obtain the socket_path from the user for connecting to MySQL, you'll need to make the following changes to the manifest.json file:


`../../mysql/lib/manifest.json`

Add a new option for the socket_path. This option should allow the user to enter the socket path required to connect to MySQL. Here's an example of how it can be added:

`      {
        "label": "Socket Path",
        "type": "text",
        "name": "socket_path",
        "placeholder": "Enter the socket path"
      }`

kariem:
@arpitnath Thank you for the feedback!

I have now added `socket_path` to `mysql/lib/manifest.json`. In addition to the changes outlined by you, I have also made changes to the sections `source.options` and `defaults`.

How about the `required` section? Adding the `socket_path` option would make the currently required options for `host` and `port` redundant. Should we mention this in the description for `socket_path`?

Suggestion

> `Enter the socket path. (Set host=localhost, port=3306)`

akshaysasidrn:
@kariem Thank you for your contribution!

PR #5824, opened 2023-03-23
Fix for entrypoint.sh file to take default port value 5432 in database_url

adishM98:
- The issue was that when we pass a string without the port value (eg: postgre://postgres:postgres@postgres/tooljet_development), It was not taking a default port value to 5432. Also instead of using sed, changed it to awk and pulled the value according to the field.

- Also added the information in env-vars document.

- Fixed the issue lint action in ci

### DIFF
--- a/docs/versioned_docs/version-2.2.0/setup/env-vars.md
+++ b/docs/versioned_docs/version-2.2.0/setup/env-vars.md
@@ -46,6 +46,11 @@ ToolJet server uses PostgreSQL as the database.
 If you are using docker-compose setup, you can set PG_HOST as `postgres` which will be DNS resolved by docker
 :::
 
+:::info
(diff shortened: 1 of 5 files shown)


docs/versioned_docs/version-2.2.0/setup/env-vars.md
```diff
@@ -98,6 +103,11 @@ When this feature is enabled, the database name provided for `TOOLJET_DB` will b
 Incase you want to trigger it manually, use the command `npm run db:create` on ToolJet server.
 :::
 
+:::info
+If you intent you use the string and if the connection does not support ssl. Please use the below format using the variable TOOLJET_DB_URL.
```
akshaysasidrn, line 107: ~use the string~ can reword as db connection URL.

PR #5830, opened 2023-03-23
[Fix] Global datasources & Environment Id issue

shah21:


server/src/services/data_sources.service.ts
```diff
@@ -85,23 +85,25 @@ export class DataSourcesService {
...
+        environmentId = env?.id;
+      } else {
+        throw new NotAcceptableException('Environment id should not be empty');
+      }
     }
 
     if (environmentId) {
       dataSource.options = (
-        await this.appEnvironmentService.getOptions(dataSourceId, organizationId, environmentId)
+        await this.appEnvironmentService.getOptions(dataSourceId, dataSource.organizationId, environmentId)
```
gsmithun4, line 106: organization id in data source is not mandatory field. will be present only for global data sources.

gsmithun4 (changes requested):

PR #5832, opened 2023-03-24
Fixed modal closing and widget selection issue inside modal

### LINKED ISSUE #5831: Not able select widgets inside modal and not able to close the modal
### Environment

Latest Develop

- When the widgets are dropped to modal, the users are able to select the widgets by clicking on the widget.
- Opened modal should be closed irrespective of the widgets selected inside the modal

- Not able to select any widgets inside the modal
- When the modal close button is clicked after selecting any of the child widget inside modal, users are not able to close the modal

1. Drop a modal widget into canvas
2. Open the modal widget
3. Drop any two widgets into the opened modal
4. Try to select the first widget we dropped. It cannot be selected

### Any possible solutions?

`onClick` event can be added to the `Rnd` parent and upon clicking widgets we can call the `setSelectedComponent` function and pass the clicked widget ID. After fixing this, encountered another issue - not able to close the modal once I click the widgets inside the modal. This is due to an issue with state update on `Editor.jsx`. 

Widget position is changing a bit when we close the modal. So, `appDefinition` is getting changed along with the `currentState`. As a quick fix, as we're calculating the current state during `componentDidUpdate`, we can remove computing current state when appDefinition is changed.

`frontend/src/Editor/Components/Modal.jsx`

Yes

kavinvenkatachalam:
Fixes #5831

### DIFF
--- a/frontend/src/Editor/Components/Modal.jsx
+++ b/frontend/src/Editor/Components/Modal.jsx
@@ -53,23 +53,15 @@ export const Modal = function Modal({
   );
 
   useEffect(() => {
-    if (exposedVariables.show !== showModal) {
-      setExposedVariable('show', showModal).then(() => fireEvent(showModal ? 'onOpen' : 'onClose'));
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [showModal]);
-
-  useEffect(() => {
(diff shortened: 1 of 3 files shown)


arpitnath (approved):
LGTM

sherfin94 (approved):

PR #5838, opened 2023-03-24
Added option to programatically change value of cells of a column in table

### LINKED ISSUE #5801: Add option to programatically change value of cells of a column in table
Assuming the column is of type number and getting data from a query. 
Business logic requires this number to be added with a constant 5, then the developer should be able to specify at the column level to add 5 in all the cell values. 
e.g. - {{current + 5}}

manishkushare:
Resolves #5801


frontend/src/Editor/Inspector/Components/Table/Table.jsx
```diff
@@ -165,6 +165,7 @@ class TableComponent extends React.Component {
       { name: '+12:00', value: 'Etc/GMT-12' },
       { name: '+13:00', value: 'Pacific/Auckland' },
     ];
+    console.log('table-- column', column);
```
kavinvenkatachalam, line 168:
```suggestion
```

kavinvenkatachalam (changes requested):
@manishkushare Sorting is not working properly. Can you check it?

PR #5857, opened 2023-03-28
Fixes for integration issues

### LINKED ISSUE #5856: Query is not running on preview mode, if two runJS or runPy is running on page load.
### Environment
Develop 

If queries's run on page load toggle is on, it should be running on preview.
even if there is more similar queries.

On load queries are not running if there is two similar queries (two runJS or runPy)

1. Add Two runJS queries with run on page load. 
2. Use the data of any of the queries on a component.
3. Preview the app

kavinvenkatachalam:
Fixes 
#5856 
#5860
#5864


emidhun:
@kavinvenkatachalam  Tested and verified the functionality working perfectly for me.
Will retest if there is any review changes.

arpitnath (approved):

PR #5877, opened 2023-03-29
Adding scroll for container widget

### LINKED ISSUE #2301: Scrollbars in Container widgets

It seems sensable to make container widgets scrollable. Use-case: I want to use a modal to fill in a form. But the form will be very limited without the ability to scroll through a container.

It would be nice if the Widget-area in a Container is infinitely expandable downwards. That makes it possible to display more information, e.g. in modals, or handle overflow.

 I can use pagination or tabs, but that's a cumbersome work-around.

I'm not a programmer by trade, I'm sorry. But I'm trying to help by giving feedback on usability and doing bug reports.

stepinfwd:
closes #2301


kavinvenkatachalam (changes requested):
@stepinfwd Can you implement the same way how modal/form works?

kavinvenkatachalam (dismissed):

Issue #5732, opened 2023-03-10
New version creation flow enhancement while building an app


thehacklady:
### Version Information
v2 

### What is the expected behaviour?

1. New version is being asked to be created from 1st version but should be from current version.
2. Released version shouldn’t be editable. It should be frozen with a permanent info message that it can’t be edited.
3. ‘Create new version popup’ on released version also need to ask to create version from the released version and not from 1st version. (this automatic popup should not be there at all if we implement 2)
4. Switching to released version gives popup to create new version. Should be removed

### Screenshots or Screencast

Related to point 1 above: 

<img width="1439" alt="point1" src="https://user-images.githubusercontent.com/32910986/224339731-66d067a6-9ec1-4b19-85ed-78fada5ed99a.png">

shreyashghadge11:
@shubh22  claim

shreyashghadge11:
I have raised a pr

PR #5912, opened 2023-04-02
New version creation flow enhancement while building an app #5732 Solved

shreyashghadge11:
Made changes to the file https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/Editor/AppVersionsManager/CreateVersionModal.jsx on line 100. The dialog was always showing the first version and not the latest one. 
The problem was that the default value was set to the last element of array which was the first element as the array was passed in descending order. Just had to put default value as first element of array.


withshubh:
Hey @shreyashghadge11 - Thanks for contributing to ToolJet. We will review the PR soon.

withshubh:
Hey @shreyashghadge11 - Sorry to inform you that the issue you picked was not open for community members. It is being worked upon internally. We have to close this PR, please feel to comment on any issue with a `good-first-issue` or `up-for-grabs` label.

PR #5947, opened 2023-04-05
App version creation flow enhancement - New released version warning popup - Current version issues

### LINKED ISSUE #5570: First version name populated while editing another version name.
Version name should be populated with the same version name while editing the version name instead of the first version name.

Version name is populated with the first version name.

1. Create an application.
2. Create another version.
3. Try & edit the version name & observe.

https://user-images.githubusercontent.com/41302917/219020498-d54e6c95-420a-4eef-8e83-c1adbf809e04.mov

### LINKED ISSUE #5711: Delete icon overlaps with version name in version list
The delete icon should not overlap with the version name.

The delete icon overlaps with the version name.

1. Login with the valid credentials
2. Create an application
3. Click on the version field, having the current version as the value
4. Click on the create new version link and create a new version with 25 characters
5. Click on the version field and move the mouse cursor to the newly created version

### Screenshots or Screencast
Screenshot 2023-03-06 at 3 54 25 PM

### LINKED ISSUE #5732: New version creation flow enhancement while building an app

### LINKED ISSUE #5869: Versioning related observations

- User should be able to see the last saved version on deleting the latest one.
- App could not be saved error should not appear on deleting latest version and adding some new components on the screen.
- User should be able to see the selected version name while renaming, and not the prior updated version name.
- User should be able to see the rename icon besides every version name.

**Case 1:** Create version v3 from v2, delete v3 version. The selected version should be assigned to the last saved one, currently it is not showing any version as selected, and now if user adds any new component system will show error as 'app could not be saved'
**Case 2:** Select version v2 and rename it (v2 rename). Select any other version (v3) and try to rename that. Ideally the latest selected name should appear in the rename text-box (v3), whereas system shows the last re labeled name only (v2 rename).
**Case 3:** Delete version confirmation pop up has black background (in light mode), which is not consistent with other delete confirmation pop ups.

1. Create new version
2. Delete the version
3. Try to rename the existing version names and notice the above mentioned cases

### Screenshots or Screencast
Screenshot 2023-03-29 at 3 07 58 PM

### Can you identify the location in 

the driver source code where the problem exists?

shah21:
Resolves #5732
Resolves #5869
Resolves #5711
Resolves #5570


kavinvenkatachalam (changes requested):
@shah21 Few things I noticed,

1. The message UI looks better if it is a straight line. Current one looks bit different
2. If I release a version and then if I try to switch to unreleased version, I'm not able to edit the name of the query or delete the query
3. Instead of checking `isVersionReleased` every where, I would suggest to have some class appended to the body and with CSS we can disable all the input boxes, button etc

kavinvenkatachalam:
@shah21 Can we wait for sometime until we merge the global state? This needs to reworked when we merge the state management.

shah21:
> @shah21 Can we wait for sometime until we merge the global state? This needs to reworked when we merge the state management.

@shubh22 @thehacklady ⬆️  Can we wait ?

kavinvenkatachalam:
Seems like state management will take some time and we can proceed this PR.

@shah21 Can you resolve the conflicts?

frontend/src/Editor/AppVersionsManager/ReleasedVersionError.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import cx from 'classnames';
+
+export const ReleasedVersionError = ({ isUserEditingTheVersion, changeBackTheState }) => {
+  React.useState(() => {
+    setInterval(() => changeBackTheState(), 2000);
+  }, [isUserEditingTheVersion]);
```
kavinvenkatachalam, line 7: Clear the `setInterval` when the component is unmounted. Even if we go back to the dashboard, the `setInterval` keep on running.

kavinvenkatachalam (changes requested):
@shah21 Everything looks good! Requested a small change, please take a look

kavinvenkatachalam (approved):

server/src/services/plugins.service.ts
```diff
@@ -238,8 +238,8 @@ export class PluginsService {
 
     try {
       // validate manifest and operations as JSON files
-      const validManifest = JSON.parse(manifest.toString()) ? manifest : null;
-      const validOperations = JSON.parse(operations.toString()) ? operations : null;
+      // const validManifest = JSON.parse(manifest.toString()) ? manifest : null;
```
gsmithun4, line 241: shall we remove the entire try catch if its not used?

server/src/dto/version-edit.dto.ts
```diff
@@ -11,4 +11,8 @@ export class VersionEditDto {
 
   @IsOptional()
   definition: any;
+
+  @IsBoolean()
+  @IsOptional()
```
gsmithun4, line 16: Should this be 
```
  @IsOptional()
  @IsBoolean()
```
?
so that it checks if value is boolean if its not null

gsmithun4 (changes requested):

server/src/services/plugins.service.ts
shah21: I am not very sure about this part. I commented this code only because the lint was failing

shah21: I will confirm with @arpitnath

shah21: I will remove the comments and will rebase with the `develop` once @arpitnath's lint PR merged

kavinvenkatachalam (approved):

gsmithun4 (approved):

PR #5970, opened 2023-04-08
CI (workflows): bump actions version, fix Node deprecation warning

kbdharun:
## Changes

- Bump the `actions/checkout` version from `v2` (Node 12) to `v3` (Node 16).
- Bump the `actions/setup-node` version from `v2` (Node 12) to `v3` (Node 16).
- Bump the `actions/cache` version from `v2` (Node 12) to `v3` (Node 16).
- Bump the `actions/upload-artifact` version from `v2` (Node 12) to `v3` (Node 16).
- Bump aws-actions/configure-aws-credentials to `v2`.

This PR fixes the Node 12 deprecation warnings under Annotations in Action runs by bumping the version of the workflow to `v3` which uses Node 16.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,15 +36,15 @@ jobs:
 
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Use Node.js 18.3.0
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: 18.3.0
 
       - name: Cache node modules
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         env:
           cache-name: cache-node-modules
         with:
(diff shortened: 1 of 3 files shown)


adishM98:
Hi @kbdharun, 

Thank you for the input. But since we are using node v18.x, I guess the below code block fixes that - 

```
      - name: Use Node.js 18.3.0
        uses: actions/setup-node@v2
        with:
          node-version: 18.3.0
```

kbdharun:
> Thank you for the input. But since we are using node v18.x, I guess the below code block fixes that - 
> 
> ```
>       - name: Use Node.js 18.3.0
>         uses: actions/setup-node@v2
>         with:
>           node-version: 18.3.0
> ```

The `18.3.0` version definition, is for the build rather than the action itself, so as long as `setup-node@v2` is used some script will still be using Node 12. More information here https://github.com/actions/setup-node/releases/tag/v3.0.0.

Since, we are bumping the actions/checkout version to v3, having the same in setup-node will provide better compatibility.

adishM98:
Understood, but irrespective of changing the version setup-node@v3. I would still need to mentioned to use the verison 18.3.0, isn't that right ?

kbdharun:
> Understood, but irrespective of changing the version setup-node@v3. I would still need to mentioned to use the verison 18.3.0, isn't that right ?

Yes, if the build workflow here requires the specific version then you need to specify it irrespective of the version of the action. As you can see here https://github.com/actions/setup-node, it is optional but it is recommended to do so.

akshaysasidrn (approved):

adishM98 (approved):

adishM98:
@kbdharun We have tested the PR and looks good and approved it as well.

PR #5980, opened 2023-04-10
Introduced global state management

### LINKED ISSUE #5979: Global State Management

Currently, the state is maintained at the `Editor` component and the props are drilled to the component where it is really needed. This causes unwanted render and scalability issues

Introduce `Zustand` to maintain the global state and move the `dataQueries` & `dataSources` state from Editor component to Zustand store. Similar to hooks, data can be consumed using useStore().

NA

Yes

kavinvenkatachalam:
Resolves #5979


sherfin94 (approved):

sherfin94 (dismissed):

kavinvenkatachalam:
@sherfin94 Figured out the following bugs and fixed it -

1. After deleting the last query, the query panel doesn't move the create state (It was still showing the last deleted query values).
2. Query in the inspector was not updated properly (currentState updation issue)
3. After deleting the query, the query needs to be selected by default or if selected query is different from the deleted query, then the selected query should be selected.

cc: @Nandinisaha13

sherfin94 (changes requested):
Few issues that I've bumped into:

* If a query is renamed, all the run query events lose their connection
* When a query is renamed before Create is clicked, the name goes away after create is clicked.

Please check

sherfin94 (approved):

sherfin94 (approved):

PR #6019, opened 2023-04-12
Fixed logout event issue on editor

### LINKED ISSUE #6018: Logout event is not working
### Version Information

2.4.0

### Environment

Develop

Logout event should work when we attach the event to any action and user should be log out from the editor.

When the logout event is triggered, the user was redirected to login page and immediately, the user was taken to the dashboard without login

1. Drop a button on the canvas
2. Add an onClick event with logout action
3. Click the button 

### Screenshots or Screencast

NA

We're clearing the local storage data when the event is triggered. This got changed recently

### Any possible solutions?

Need to call the `authenticationService` logout function 

`frontend/src/_helpers/appUtils.js`

Yes

kavinvenkatachalam:
Fixes #6018


shah21 (approved):
LGTM

PR #6036, opened 2023-04-14
Fix ::  View remains in same position during page switch in multi page apps

### LINKED ISSUE #5758: Switching pages via event handler doesn't move the cursor on top of the next page

While using an event handler (say on click) and adding switch page actions, the cursor should move on top of another page.

When a user tries to create multiple pages using multipage feature, and move from first page to second by using an event handler and then adding switch page action to it, then the cursor remains at the exact same point in the second page as it was in the first page. User has to scroll upwards/downwards depending on the components and might also end up in an empty space which does not give a good experience.

1. Create two pages
2. Add a component in the very bottom of the first page, need to scroll down.
3. Add a component in the second page in the top.

After this, go to the component in the first page and try to switch to the second tab by adding an event handler.
You'd be able to see that the page isn't moving to top on switch.

stepinfwd:
closes #5758


kavinvenkatachalam (changes requested):
@stepinfwd It working only on the editor mode. Can you implement the same on viewer mode?

kavinvenkatachalam (approved):

PR #6038, opened 2023-04-14
Fix :: Additional button is getting added in Button Group component by changing value in labels

### LINKED ISSUE #5876: Additional button is getting added in Button Group component by changing value in labels field
The count of buttons in Button Group component should not change on changing the labels values.

Based on user inputs in labels, the default count of number of buttons within Button Group component keeps changing.

1. Drag Button Group component
2. Update the value of labels field multiple times, enter value like {{[22]}}, change it to {{[2]}} again change it to other value, the count of button will increase
3. Change input in values field, enter {{[1,2,3,4]}}, set the label value as {{[2]}}, now keep the labels input as blank, additional button will be displayed in the group

### Screenshots or Screencast
Screenshot 2023-03-29 at 5 23 15 PM

Screenshot 2023-03-29 at 5 23 47 PM

stepinfwd:
closes #5876


frontend/src/Editor/Components/ButtonGroup.jsx
```diff
@@ -34,20 +34,21 @@ export const ButtonGroup = function Button({
...
-  // data is used as state to show what to display , club of label+values / values
+  const [data, setData] = useState(values);
+
   useEffect(() => {
     setDefaultActive(defaultSelected);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [JSON.stringify(defaultSelected)]);
 
   useEffect(() => {
+    setData(values);
```
kavinvenkatachalam, line 45:
```suggestion
```

kavinvenkatachalam (changes requested):
@stepinfwd There is a change in behaviour. 

In the cloud if we add 5 labels and only 3 values, then total 5 buttons are displayed. Now in this PR, it has been changed to 3 as we're displaying buttons based only on the values. Can you please take a look?

frontend/src/Editor/Components/ButtonGroup.jsx
stepinfwd: > 
Below this line we are checking for length of labels vs values and setting data to labels if length is more

frontend/src/Editor/Components/ButtonGroup.jsx
```diff
@@ -34,20 +34,21 @@ export const ButtonGroup = function Button({
...
+    const dataset = values;
     if (labels?.length < values?.length) {
-      // eslint-disable-next-line no-unsafe-optional-chaining
-      setData([...labels, ...values?.slice(labels?.length)]);
+      labels.map((item, index) => {
+        dataset[index] = item;
+      });
+      setData(dataset);
     } else {
       setData(labels);
```
stepinfwd, line 53: Here @kavinvenkatachalam

kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

PR #6040, opened 2023-04-14
Fixed issue on running app load queries multiple time for the released app

kavinvenkatachalam:
Fixed running query multiple time on app load for the released app


shah21 (approved):
LGTM

PR #6069, opened 2023-04-19
Fix: replaced this by page name

### LINKED ISSUE #5991: Add page name in delete page confirmation popup
When a user is trying to delete a page, they see a popup to confirm the deletion.
In this popup, a user should be shown the name of the page as well and not just the generic message to make sure that user is deleting the right page.

Screen Shot 2023-04-07 at 15 04 29

The message should say - 
Do you really want to delete <b>{{page name}}</b> page?

apshada:
Closes #5991

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -1402,12 +1402,13 @@ class EditorComponent extends React.Component {
     );
   };
 
-  deletePageRequest = (pageId, isHomePage = false) => {
+  deletePageRequest = (pageId, isHomePage = false, pageName = '') => {
     this.setState({
       showPageDeletionConfirmation: {
         isOpen: true,
         pageId,
         isHomePage,
+        pageName,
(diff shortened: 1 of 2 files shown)


withshubh:
Hey @apshada - Thanks for creating the PR, we will review it soon.

kavinvenkatachalam (changes requested):
Hey @apshada,

Thanks for the PR. If we have multiple pages and try to delete a page other than the current page, the modal shows the current page name. It will be confusing for the users. Attaching the GIF

page






I would suggest to pass a page name (`page.name`) to the delete function in line 57 `frontend/src/Editor/LeftSidebar/SidebarPageSelector/PageHandler.jsx` and use it to set the `showPageDeletionConfirmation` state. Then you can use the name in the delete confirmation modal. Attaching the screenshot for your reference.

image
image
image
image

kavinvenkatachalam (approved):

PR #6074, opened 2023-04-19
Feature - ToolJet Copilot 🚀

arpitnath:
ToolJet Copilot is designed to help you write code faster and more efficiently, all while maintaining the highest standards of quality and functionality.

It enables you to tap into expert guidance and support, streamlining your coding process and helping you to work more efficiently. Regardless of whether you're developing custom JavaScript functions or crafting complex SQL queries, our Copilot feature is there to provide assistance every step of the way.


server/src/services/copilot.service.ts
```diff
@@ -0,0 +1,71 @@
...
+    const response = await fetch('https://0p94cxsi3g.execute-api.us-west-1.amazonaws.com/Prod/api-key', options);
+    const { data } = await response.json();
+
+    return {
+      data: data,
+      status: response.status,
+    };
+  }
+
+  async getCopilotAPIKey(userId: string) {
```
akshaysasidrn, line 54: Since it is tied to the workspace, I'd think it's better to scope it as such rather than having it tied to the user. So we can have the feature on workspace basis.

server/src/controllers/copilot.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+  @Post('api-key')
+  async addUpdateCopilotAPIKey(@User() user, @Body() body: AddUpdateCopilitAPIKeyDto) {
+    const { key } = body;
+    const userId = user.id;
+    return await this.copilotService.addUpdateCopilotAPIKey(key, userId);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Get('api-key')
+  async getCopilotAPIKey(@User() user) {
```
akshaysasidrn, line 48: Is it necessary to have this endpoint actually? If it is to just check API key is present or not, we can give the response as such. Stating because it is generally not a good practice to give unencrypted API key back to the client as it can then be exploited if breached.

akshaysasidrn (changes requested):

akshaysasidrn (approved):

akshaysasidrn (approved):

PR #6082, opened 2023-04-20
Change default server listen addr to '::' for dual IPv6/IPv4 support

jpwarren:
Add a `LISTEN_ADDR` environment variable for configuration of the IP address/hostname the server binds to.

`LISTEN_ADDR` defaults to `'::'` so the server supports both IPv6 and IPv4 by default.


akshaysasidrn (approved):
Thank you for your contribution

PR #6128, opened 2023-04-25
bug-fix: 🐛 - handle when a filter exist for equal & null

### LINKED ISSUE #5616: Tooljet database: Update filter doesn't works when user keeps "Null" value in condition for specific row until user applies second condition

1. Login with valid credentials and navigate to the Database page
2. Create a table with appropriate columns and add values in it.(Keep some values as Null/Empty)
3. Go to the application page, create an application, Add a query with the tooljet database.
4. In the query builder, Select the table name created in step 2. 
5. Select the operation as an update row. Apply the filter such that:
- select the column name
- selection operation as "equals" 
- Enter the value as Null.
6. Inside the column section, select a column and provide the value. 
7. Run the query.

vikotar:
- This code change handles when a user selects `equal` filter with `null` value. It stops the query execution and thorws an error message with the text `Query can not run, because app can not perform equal operation with null value`

Closes #5616

### DIFF
--- a/frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/util.js
+++ b/frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/util.js
@@ -21,13 +21,13 @@ import { get } from 'lodash';
  *
  * const result = hasEqualWithNull(queryOptions); // true
  */
-export const hasEqualWithNull = (queryOptions) => {
-  const filters = get(queryOptions, 'list_rows.where_filters');
+export const hasEqualWithNull = (queryOptions, operation) => {
+  const filters = get(queryOptions, `${operation}.where_filters`);
   if (filters) {
     const filterKeys = Object.keys(filters);
     for (let i = 0; i < filterKeys.length; i++) {
       const filter = filters[filterKeys[i]];
-      if (filter.operator === 'eq' && filter.value === '{{null}}') {
+      if ((filter.operator === 'eq' || filter.operator === 'neq') && filter.value === null) {
         return true;
       }
     }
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/util.js
```diff
@@ -0,0 +1,36 @@
...
+ *         value: '{{null}}',
+ *       },
+ *     },
+ *   },
+ * };
+ *
+ * const result = hasEqualWithNull(queryOptions); // true
+ */
+export const hasEqualWithNull = (queryOptions) => {
+  const filters = get(queryOptions, 'list_rows.where_filters');
```
akshaysasidrn, line 25: We need to do this check for the other operations where filters are applied, not only for list_rows.

frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/util.js
```diff
@@ -0,0 +1,36 @@
...
+  if (filters) {
+    const filterKeys = Object.keys(filters);
+    for (let i = 0; i < filterKeys.length; i++) {
+      const filter = filters[filterKeys[i]];
+      if (filter.operator === 'eq' && filter.value === '{{null}}') {
```
akshaysasidrn, line 30: I'd also think it's better to do this check after the query options are resolved such that we don't have to check for hardcoded js code (`{{null}}`) before its resolved. Wdyt?

vikotar: It totally makes sense. @akshaysasidrn 
Thanks for the hint. Let me change the code.

akshaysasidrn (dismissed):

akshaysasidrn (approved):

PR #6137, opened 2023-04-25
Fix :: Repeated toast bug after adding new row and opening a drawer in  TJDB

### LINKED ISSUE #6104: The toast message was displayed repeatedly.
The toast message should display only after action.

The toast message was displayed repeatedly.

2. Navigate to the database page 
3. Create a table with a valid column name
4. Click on the "Add new row" button and enter valid input values in the fields.
5. Click on the save button
6. Repeat step no. 4 for the next row.

### Screenshots or Screencast

https://user-images.githubusercontent.com/59684099/233964837-e852651a-cddf-46a2-99a6-e7c2af8f9455.mov

stepinfwd:
closes #6104 
* fixes small ui bug in create table where footer is not fixed to bottom

### DIFF
--- a/frontend/src/TooljetDatabase/Drawers/CreateRowDrawer/index.jsx
+++ b/frontend/src/TooljetDatabase/Drawers/CreateRowDrawer/index.jsx
@@ -12,7 +12,9 @@ const CreateRowDrawer = ({ isCreateRowDrawerOpen, setIsCreateRowDrawerOpen }) =>
   return (
     <>
       <button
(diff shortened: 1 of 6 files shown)


kavinvenkatachalam (approved):

akshaysasidrn (approved):
Verified it works. Still having to dismiss all toasts doesn't seems pragmatic for when the components re-renders. I did come across an issue related here: https://github.com/timolins/react-hot-toast/issues/272
If we give id to our toasts I think this should be handled. Just sharing the reference here.

PR #5759, opened 2023-03-15
Tree component is not reflecting data correctly

manishkushare:
Resolves #5684 
Resolves #5774
Resolves #5789
Resolves #5790
Along with the issues mentioned in the referred issue, tree select has the same behavior, when checekdData and expandedData properties are not valid, this PR addresses those issues as well.

PR #6138, opened 2023-04-25
Bugfix :: Query driven tabs fail with no data 

### LINKED ISSUE #5533: Query driven tabs fail with no data and cannot be recovered
### Version Information
2.1.1

### Environment
Linux

If a dataset cannot be generated or returns no rows, the Tabs component should just display no tabs.   

It generates a "Something went wrong" text and that cannot be recovered from.   Neither, quitting the app and returning nor a hard browser refresh  fixes the problem.   

1.  Create a dataset that returns a proper tab data structure, but one that does not refresh on page start.
2.  Save the query again or leave the app and return.

### Screenshots or Screencast
image

### Any possible solutions?
One solution is to check for undefined within the path navigation in the tabs data definition and default to something that is displayable.   I.e.:

{{queries?.myQuery?.data?.events  || [ { title: 'None', id: '0'}  ]}}

No.

No.

stepinfwd:
closes #5533

### DIFF
--- a/frontend/src/Editor/Components/Tabs.jsx
+++ b/frontend/src/Editor/Components/Tabs.jsx
@@ -147,7 +147,7 @@ export const Tabs = function Tabs({
           margin: '-1px',
         }}
       >
-        {parsedTabs.map((tab) => (
+        {parsedTabs?.map((tab) => (
           <li
             className="nav-item"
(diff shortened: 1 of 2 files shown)


kavinvenkatachalam:
This is already working fine even if we pass `undefined` value. It might fixed by #5759

@manishkushare Can you confirm this?

manishkushare:
Yes , @kavinvenkatachalam , it is working fine in latest develop

PR #6202, opened 2023-05-01
fix: Tooljet is not loading the released apps which has some specific texts in sharable link path

shah21:


frontend/src/App/App.jsx
```diff
@@ -75,7 +75,9 @@ class AppComponent extends React.Component {
       'confirm',
       'confirm-invite',
     ];
-    return existedPaths.find((path) => window.location.pathname.includes(path));
+
+    const pathnames = window.location.pathname.split('/')?.filter((path) => path != '');
+    return pathnames?.length > 0 ? existedPaths.find((path) => pathnames[0] === path) : false;
```
akshaysasidrn, line 80: nitpick: Can we make this to return boolean? Will return `undefined` when find has no matches.

frontend/src/App/App.jsx
```diff
@@ -75,7 +75,12 @@ class AppComponent extends React.Component {
       'confirm',
       'confirm-invite',
     ];
-    return existedPaths.find((path) => window.location.pathname.includes(path));
+
+    const subpath = getSubpath();
+    const subpathArray = subpath ? subpath.split('/').filter((path) => path != '') : [];
+    const pathnames = window.location.pathname.split('/')?.filter((path) => path != '');
```
akshaysasidrn, line 81: Do you think it'd be simple to reason about if we get the subpath and then split it out from `window.location.pathname` and go about the normal logic?

akshaysasidrn: can keep for refactoring later

akshaysasidrn (approved):

PR #6211, opened 2023-05-02
GRPC Datasource

### LINKED ISSUE #6218: Add 'gRPC' plugin for connecting to gRPC server APIs
## Proposal
Currently, ToolJet provides plugins/datasources for REST APIs and GraphQL.
The proposed plugin will allow users to easily connect to gRPC APIs, send and receive data from the server, and work with protobuf messages. This will make it easier for developers to build applications using gRPC and ToolJet.

This plugin will be available for self-hosted deployments and should support unary requests and responses.

arpitnath:
Resolves #6218


server/src/controllers/global_data_sources.controller.ts
```diff
@@ -46,6 +47,10 @@ export class GlobalDataSourcesController {
     }
 
     const decamelizedDatasources = dataSources.map((dataSource) => {
+      if (dataSource.pluginId) {
```
akshaysasidrn, line 50: why do we have to skip plugins now as compared to earlier?

server/src/controllers/global_data_sources.controller.ts
```diff
@@ -72,6 +77,15 @@ export class GlobalDataSourcesController {
...
     }
 
+    if (kind === 'grpc') {
+      const rootDir = process.cwd().split('/').slice(0, -1).join('/');
+      const protoFilePath = `${rootDir}/protos/service.proto`;
+      const fs = require('fs');
+
+      const filecontent = fs.readFileSync(protoFilePath, 'utf8');
+      const rcps = await getServiceAndRpcNames(filecontent);
+      options.find((option) => option['key'] === 'protobuf').value = JSON.stringify(rcps, null, 2);
```
akshaysasidrn, line 87: I'd suppose when we add in upload capability these, preprocessing code would go onto the client side right?

plugins/packages/grpc/lib/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    if (!sourceOptions.url) {
+      throw new QueryError('Missing URL', {}, {});
+    }
+
+    const cwd = process.cwd();
+    const rootDir = cwd.split('/').slice(0, -1).join('/');
+    const protoFilePath = `${rootDir}/protos/service.proto`;
+
+    if (!sourceOptions.protobuf || sourceOptions.protobuf.length === 0) {
```
akshaysasidrn, line 19: This will always be empty string currently yes? No need to remove, can be placeholder for check later.

frontend/src/Editor/DataSourceManager/DataSourceManager.jsx
```diff
@@ -105,7 +105,7 @@ class DataSourceManagerComponent extends React.Component {
     this.setState({
       dataSourceMeta: source.manifestFile?.data?.source ?? source,
       selectedDataSource: source.manifestFile?.data?.source ?? source,
-      selectedDataSourceIcon: source.iconFile?.data,
+      selectedDataSourceIcon: source?.iconFile?.data,
```
akshaysasidrn, line 108: I suppose the source would never be missing here, can skip optional chaining?

server/src/controllers/global_data_sources.controller.ts
arpitnath: Yes

akshaysasidrn (approved):

arpitnath:
docs: #6242

PR #6222, opened 2023-05-02
Add Option to Use Existing Secret for Helm Chart

jonathon2nd:
This PR adds the option to use an existing secret for ToolJet's Helm chart, which prevents secret environment variables from being exposed when viewing the deployed app's values.

Changes:
- Update `deploy/helm/README.md` to include instructions for secret management using an existing secret
- Modify `deploy/helm/templates/tooljet/secret.yaml` to conditionally create a secret only if `.Values.apps.tooljet.useExistingSecret` is set to `false`
- Add `useExistingSecret: false` to `deploy/helm/values.yaml` under `apps` section

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -3,6 +3,15 @@
 This chart can be used to install ToolJet in a Kubernetes Cluster via [Helm v3](https://helm.sh).\
 This setup is very rudimentary and comes with an included PostgreSQL server out of the box.
 
+## Secret management
+You can use an existing secret to configure psql auth and additional ToolJet secret env vars.
+`.Values.apps.tooljet.useExistingSecret: true`
+
+See deploy/helm/templates/tooljet/secret.yaml for desired format.
+`.Values.apps.tooljet.secret.name` must be the name of the manually created secret.
+
+If `.Values.apps.tooljet.useExistingSecret: true`, `.Values.apps.tooljet.secret.data` will not be used by helm chart.
+
 ## Installation
 
 To install, follow these steps:
--- a/deploy/helm/templates/tooljet/secret.yaml
+++ b/deploy/helm/templates/tooljet/secret.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.apps.tooljet.useExistingSecret -}}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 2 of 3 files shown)


adishM98:
Hi @jonathon2nd, 

Thank you for your input. Just to get a better understanding. You have added this check (.Values.apps.tooljet.useExistingSecret: true`) so that it gives the user to use their own DB or connect to the existing DB.

jonathon2nd:
With the usage of the postgres sub-chart, not really sure on a sane way to integrate a user generated secret for it.
So with the default `useExistingSecret: false` , this would continue to support `postgresql.enabled: true`

Maybe it would be best to specify that using `useExistingSecret: true` assumes and requires an already existing DB. Which makes sense, as if you are ok for those values being in `postgresql.auth` in the values for the built in subchart, you are probably fine with them in `apps.tooljet.secret.data`

What do you think?

akshaysasidrn:
Hi @jonathon2nd, I'm wondering if this would be necessary to have though. Current chart already encompasses the necessary setup for configuring Tooljet out of the box. So if any further medication is needed, couldn't be modified at deployment.yaml rather as secrets are inferred from there?
Please also note, we are setting up helm setup in a different repo and would deprecate it here. Welcoming your contributions [there](https://github.com/ToolJet/helm-charts).

jonathon2nd:
While I understand the perspective that changes could potentially be made in the deployment.yaml, this approach presents a significant problem. Every time the Helm chart is updated, any manual modifications made at deployment.yaml would be overwritten, leading to the loss of those changes. This creates an unnecessary maintenance burden and increases the risk of deployment issues caused by lost or forgotten modifications.

The intent behind this PR is to provide a sustainable, maintainable way to manage secrets, without having to worry about those changes being lost during updates. This change would allow users to leverage existing secrets which could be more aligned with their overall Kubernetes secret management strategy.

Furthermore, promoting the usage of existing secrets would provide more flexibility and security to users, allowing them to employ secrets management solutions such as AWS Secret Manager, Azure Key Vault, or even Kubernetes' own native secrets.

Also, about the deprecation and setup of the helm in a different repo, it looks like it is the exact same situation there. If accepted this PR could then be added there.

akshaysasidrn:
@jonathon2nd Sure, that makes sense. Could you open a PR for this in the helm charts repo?

PR #6255, opened 2023-05-04
Fix :: App breaking during version switch 

### LINKED ISSUE #6256: App breaking during version switch
App should not break during version switch

App breaking during version switch 

one way is to follow below step till 3 and try switching versions fast

1. Create an APP
2. Create multiple versions.
3. Create a query on one of the version.
4. Open version modal,
5. Type new version name.
6. Change the from version to the one with query.
7. Again change the from version to another,

### Screenshots or Screencast
https://user-images.githubusercontent.com/31006028/236190471-c684b685-061d-4b77-9c2e-5577cb8f266b.mov

yes

stepinfwd:
fixes #6256


stepinfwd:
closing this one for hotfixing

PR #6268, opened 2023-05-05
Bugfix - handles plugins to move from gds to lds when importing app to different workspaces

### LINKED ISSUE #6269: On importing apps which has queries built from Marketplace plugins as GDS, the app crashes
On importing apps which has queries built from Marketplace plugins as GDS, the app crashes

The imported app, with queries built from Global datasources should be handled by moving it to local data sources

1. Install a plugin from marketplace
2. create gds from it
3. create an app and create queries.
4. export the app
5. create a new workspace and import the app

arpitnath:
Fixes #6269


vjaris42 (dismissed):

vjaris42 (dismissed):

server/src/services/app_import_export.service.ts
```diff
@@ -385,6 +385,7 @@ export class AppImportExportService {
           kind: source.kind,
           type: source.type || DataSourceTypes.DEFAULT,
           appVersionId: appVersionMapping[appVersion.id],
+          pluginId: source?.pluginId || null,
```
gsmithun4, line 388: will it work for inter-instance imports?
the `source?.pluginId` of an instance won't be available with other

gsmithun4 (changes requested):

gsmithun4 (approved):

gsmithun4 (approved):

PR #6286, opened 2023-05-08
Added bar type option to chart widget

### LINKED ISSUE #4437: Plotly layout barchart stacked doesn't work
### Environment
Docker

Setting "layout": { "barmode": "stack" } should display stacked bars.

Nothing happens.

Create data with 2 bar objects like https://plotly.com/javascript/bar-charts/

Add barmode
{
"data": {{queries.mssql7.data}},
"layout": { "barmode": "stack" }
}

See that it's still using grouped.

### Any possible solutions?
Not sure if the layout prop is actually passed down.

https://github.com/ToolJet/ToolJet/blob/2bf4707624940f33867fe053bc26d058b1f58d85/frontend/src/Editor/Components/Chart.jsx

Not sure

stepinfwd:
closes #4437


johnsoncherian (approved):

frontend/src/Editor/Inspector/Components/Chart.jsx
```diff
@@ -98,6 +97,21 @@ class Chart extends React.Component {
         currentState
       ),
     });
+    {
+      resolveReferences(this.state.component.component.definition.properties.plotFromJson?.value, currentState) &&
```
johnsoncherian, line 101: Don't we already have this value computed in line 64. Can't we reuse the same here.?

johnsoncherian (approved):

withshubh:
docs: https://github.com/ToolJet/ToolJet/pull/6879

PR #6287, opened 2023-05-08
Page clone feature

### LINKED ISSUE #6297: Implement a clone/duplicate option for pages
## Proposal

Provide an option to clone or duplicate a page. User should be able to create an exact copy of a page which he/she can rename, update the url or add, delete or edit components.  The duplicated page should copy all the components and queries linked from the original page.

## Potential advantages
This should help the users to reduce development time for page that are similar to already created pages in same app. User can use the feature to quickly create a copy and make the necessary changes rather than creating the whole page from scratch.

johnsoncherian:
Fixes #6297


kavinvenkatachalam (changes requested):
@johnsoncherian One thing I noticed is, when I try to switch the page with button component, the event appears to be the same and not changing properly. We can deselect the component or we can update the correct value. Attaching the GIF
clone

kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

PR #6299, opened 2023-05-09
Workspace name sanitisation

### LINKED ISSUE #5661: Sanitising a workspace name while creating or editing

Currently, user can give a name to a workspace but the name has no defined properties which breaks the UI at some points. For example, the length of the name is defined as '25 characters' and because of this the UI breaks (check ss for reference)

Workspace admins should be able to create and edit workspace names. 

In case of creating a new workspace and adding name to it : 
1) Name should be unique.
3) Name should be mandatory. 

In case of editing workspace name : 
1) Name should be unique.
3) Name should be mandatory. 

Screenshot 2023-03-01 at 7 28 30 PM

shah21:
Resolves #5661


server/migrations/1683136077244-AddUniqueConstraintToWorkspaceName.ts
```diff
@@ -0,0 +1,36 @@
...
+      'organizations',
+      new TableUnique({
+        name: 'name_organizations_unique',
+        columnNames: ['name'],
+      })
+    );
+  }
+
+  public async migrateWorkspaceNames(entityManager: EntityManager) {
+    const workspaces = await entityManager.find(Organization);
```
gsmithun4, line 18: we don't want to get all workspaces and iterate 
```
select sub_query.name from (select count(*) as name_count, name from organizations group by name) sub_query where name_count > 1
```
use this query to get names which repeats and iterate the array

server/migrations/1683136077244-AddUniqueConstraintToWorkspaceName.ts
```diff
@@ -0,0 +1,36 @@
+import { Organization } from 'src/entities/organization.entity';
+import { EntityManager, MigrationInterface, QueryRunner, TableUnique } from 'typeorm';
+
+export class AddUniqueConstraintToWorkspaceName1683136077244 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const entityManager = queryRunner.manager;
+    await this.migrateWorkspaceNames(entityManager);
+    await queryRunner.createUniqueConstraint(
```
gsmithun4, line 8: can declare this in entity also

server/src/services/organizations.service.ts
```diff
@@ -66,6 +66,12 @@ export class OrganizationsService {
   async create(name: string, user?: User, manager?: EntityManager): Promise<Organization> {
     let organization: Organization;
     await dbTransactionWrap(async (manager: EntityManager) => {
+      const ifWorkspaceNameExists = await manager.findOne(Organization, { name });
+
+      if (ifWorkspaceNameExists) {
+        throw new BadRequestException('Workspace name already exists.');
```
gsmithun4, line 72: Should throw `ConflictException`

gsmithun4 (changes requested):

frontend/src/_helpers/handle-response.js
```diff
@@ -28,3 +28,13 @@ export function handleResponseWithoutValidation(response) {
     return data;
   });
 }
+
+export const handleErrConnections = (error) => {
```
gsmithun4, line 32: shall we do this for all APIs using some global configs, we can do it in different PR

gsmithun4 (changes requested):

frontend/src/_helpers/handle-response.js
shah21: Sure. I will remove this Part

gsmithun4 (approved):

PR #6301, opened 2023-05-09
Detach modal button height property from modal height

### LINKED ISSUE #6306: Detach height of button box from height of modal widget
Button box wrapper which opens modal height is linked to modal height 

Add a separate height property for modal 
By default height of button box wrapper should be just more than the height of the button

Yes

stepinfwd:
close #6306


kavinvenkatachalam (changes requested):
@stepinfwd Existing modals are getting affected. Can you check for the backward compatibility?

stepinfwd:
> @stepinfwd Seems like `%` is not supported. Can you check for the possibilities?

we can use other relative units like em , vw etc

kavinvenkatachalam (changes requested):
@stepinfwd Even with the latest change, it is affecting the existing modal height. I would suggest to keep the existing `height` property and can create a property with diff key like `modalHeight`. So, if the height value is available in the existing widget, we can use the height and display accordingly. If it is not available, then we can use the new logic and display with `modalHeight`

kavinvenkatachalam (dismissed):

withshubh:
docs here: https://github.com/ToolJet/ToolJet/pull/6566

PR #6309, opened 2023-05-09
Feat: modal options

vikotar:


frontend/src/WorkflowEditor/FlowBuilder/index.jsx
```diff
@@ -49,54 +52,66 @@ function FlowBuilder(_props) {
...
+        x,
+        y,
+        startingNodeId,
+        startingNodeHandleId,
+        clientX: event.clientX,
+        clientY: event.clientY,
+      });
       setEditingActivity({ type: 'IDLE' });
 
       if (event.target.className === 'react-flow__pane') {
```
vikotar, line 70: should we remove this condition, I am not using it.

sherfin94 (approved):

frontend/src/WorkflowEditor/FlowBuilder/index.jsx
sherfin94: The idea of the condition is to not create a new node if the dragging end point is another node or node handle. `react-flow__pane` represents the empty pane.

vikotar: Coolio. I will move the condition back into it. :)

PR #6319, opened 2023-05-09
render workflow changes which helps to delete pr DB

adishM98:
added the code to delete the db when destroy label is triggered

### DIFF
--- a/.github/workflows/render-preview-deploy.yml
+++ b/.github/workflows/render-preview-deploy.yml
@@ -244,9 +244,9 @@ jobs:
 
       - name: Drop PostgreSQL PR database
         env:
-          PGHOST: dpg-cf4metg2i3mtiflq3dvg-a.oregon-postgres.render.com
+          PGHOST: ${{ secrets.RENDER_PG_HOST }}
           PGPORT: 5432
-          PGUSER: preview_env_db_user
+          PGUSER: ${{ secrets.RENDER_PG_PASS }}
           PGPASSWORD: ${{ secrets.POSTGRES_PASSWORD }}
           PGDATABASE: ${{ env.PR_NUMBER }}
         run: |
--- a/.github/workflows/render-preview-deploy.yml
+++ b/.github/workflows/render-preview-deploy.yml
@@ -242,6 +242,16 @@ jobs:
               console.log(e)
             }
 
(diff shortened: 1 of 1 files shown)


.github/workflows/render-preview-deploy.yml
```diff
@@ -242,6 +242,16 @@ jobs:
...
 
+      - name: Drop PostgreSQL PR database
+        env:
+          PGHOST: dpg-cf4metg2i3mtiflq3dvg-a.oregon-postgres.render.com
+          PGPORT: 5432
+          PGUSER: preview_env_db_user
+          PGPASSWORD: ${{ secrets.POSTGRES_PASSWORD }}
+          PGDATABASE: ${{ env.PR_NUMBER }}
+        run: |
+          psql -h $PGHOST -p $PGPORT -U $PGUSER -d postgres -c "DROP DATABASE $PGDATABASE;"
```
akshaysasidrn, line 253: can we infer this from `${{ secrets.RENDER_PG_HOST }}`

.github/workflows/render-preview-deploy.yml
```diff
@@ -242,6 +242,16 @@ jobs:
               console.log(e)
             }
 
+      - name: Drop PostgreSQL PR database
+        env:
+          PGHOST: dpg-cf4metg2i3mtiflq3dvg-a.oregon-postgres.render.com
+          PGPORT: 5432
+          PGUSER: preview_env_db_user
+          PGPASSWORD: ${{ secrets.POSTGRES_PASSWORD }}
```
akshaysasidrn, line 250: ${{ secrets.RENDER_PG_PASS }}

akshaysasidrn (changes requested):

.github/workflows/render-preview-deploy.yml
```diff
@@ -242,6 +242,16 @@ jobs:
               console.log(e)
             }
 
+      - name: Drop PostgreSQL PR database
```
akshaysasidrn, line 245: Also I think we should remove the label after deleting the db

adishM98: Since i have added it under the label destroy-review-app it will be removed automatically.

akshaysasidrn (approved):

PR #6325, opened 2023-05-09
Added feature to toggle the visibility of columns programmatically

### LINKED ISSUE #6289: Ability to control Table columns visibility programmatically
Require an ability to set the column visibility of a table component programmatically.   

Expecting a field in the column options of a Table component that can set the visibility of the selected column, the `fx` of this field should enable users to change the behavior dynamically.

For example, An option just like `make editable` should be helpful.
Screenshot 2023-05-08 at 7 47 48 PM

manishkushare:
Resolves #6289


kavinvenkatachalam:
@manishkushare Can you take a pull from latest develop?

frontend/src/Editor/Inspector/Components/Table/ProgramaticallyHandleToggleSwitch.jsx
```diff
@@ -14,10 +14,36 @@ export const ProgramaticallyHandleToggleSwitch = ({
...
+        value = props.disableActionButton;
+        break;
+      case 'columnVisibility':
+        value = props.columnVisibility;
+        break;
+      default:
+        break;
+    }
+    return value;
+  };
```
kavinvenkatachalam, line 33:
```suggestion
  const getValueBasedOnProperty = (property, props) => {
    switch (property) {
      case 'isEditable':
        return props.isEditable;
      case 'disableActionButton':
        return props.disableActionButton;
      case 'columnVisibility':
        return props.columnVisibility;
      default:
        return value;
    }
  };
```

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -368,6 +368,12 @@ export function Table({
     darkMode,
   });
 
+  columnData = columnData.filter((column) => {
+    if (resolveReferences(column.columnVisibility, currentState)) {
+      return column;
+    }
+  });
```
kavinvenkatachalam, line 375:
```suggestion
  columnData = useMemo(
    () =>
      columnData.filter((column) => {
        if (resolveReferences(column.columnVisibility, currentState)) {
          return column;
        }
      }),
    [columnData, currentState]
  );
```

kavinvenkatachalam (changes requested):
@manishkushare Requested few changes. Please take a look. On the implementation wise everything looks good.

frontend/src/Editor/Inspector/Components/Table/ProgramaticallyHandleToggleSwitch.jsx
```diff
@@ -14,10 +14,36 @@ export const ProgramaticallyHandleToggleSwitch = ({
...
+
+  const getInitialValue = (property, definition) => {
+    if (property === 'columnVisibility') {
+      return definition?.value ?? `{{true}}`;
+    } else {
+      return definition?.value ?? `{{false}}`;
+    }
+  };
```
kavinvenkatachalam, line 41:
```suggestion
    if (property === 'columnVisibility') {
      return definition?.value ?? `{{true}}`;
    }
      return definition?.value ?? `{{false}}`;
  };
```

kavinvenkatachalam (changes requested):
Missed to add this

kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

PR #6356, opened 2023-05-11
Fixed vertical alignment issue for cells in the table component whose column type are action, multiselect,datepicker

### LINKED ISSUE #6355: Vertical alignment bug in the table cell for column types datepicker, multiselect, and action buttons
### Version Information
v2.5.0

### Environment
Development

The content inside each cell should be vertically aligned

The content inside each cell should be for column types date picker, multi-select, and action button are not vertically aligned

1. Create a new app or open an existing one
2. Drag a table component from the right sidebar of the editor onto the canvas.
3. Click anywhere on the table to open the table widget inspector in the right sidebar.
4. Search for columns sections in the table inspector under the properties tab
5. Add a new column, and click on the  newly added column button, it will open the pop-up window to manage the particular column settings,
6. Look for column type, the first field in the pop-up window. and select `Multiselect` as the type.
7. Now look at the table on the canvas, look for the newly added column and you can see `Multiselect` element inside each cell under the respective column, closely observe it, it's not vertically aligned. We need to make it vertically aligned.
8. Now change the column type to date picker and again observe vertical alignment, we need to fix the same here as well.
9. Now, In the table inspector, look for the `actions` column, and add a new action button. Right after this, you can observe action button column is being introduced in the table in the canvas. Observe its vertical alignment as well, we need to make it to the center.

### Screenshots or Screencast
[Tooljet - Dashboard - 11 May 2023 - Watch Video](https://www.loom.com/share/ba4f74be516a42fa8900891b77f69597)

Yes

manishkushare:
Resolves #6355


kavinvenkatachalam (dismissed):

kavinvenkatachalam:
@manishkushare Can you resolve the conflicts?

kavinvenkatachalam (approved):

PR #6389, opened 2023-05-13
improvement - changing scope of API key from user to workspace

arpitnath:
ref #6074

### DIFF
--- a/server/src/controllers/copilot.controller.ts
+++ b/server/src/controllers/copilot.controller.ts
@@ -31,6 +31,6 @@ export class CopilotController {
   @UseGuards(JwtAuthGuard)
   @Post('api-key')
   async validateCopilotAPIKey(@User() user, @Body() body: { secretKey: string }) {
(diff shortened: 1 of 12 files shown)


server/src/controllers/copilot.controller.ts
```diff
@@ -31,6 +31,6 @@ export class CopilotController {
   @UseGuards(JwtAuthGuard)
   @Post('api-key')
   async validateCopilotAPIKey(@User() user, @Body() body: { secretKey: string }) {
-    return await this.copilotService.validateCopilotAPIKey(user.id, body.secretKey);
+    return await this.copilotService.validateCopilotAPIKey(user.organizationId, body.secretKey);
```
akshaysasidrn, line 34: It is better to have current organization Id to be passed from frontend, as inferring organization id from the user can err when org switch isn't reflected incase of concurrent requests.

akshaysasidrn (dismissed):

akshaysasidrn (approved):

PR #6391, opened 2023-05-15
feat: conditional execution of even handler functions

### LINKED ISSUE #6304: Conditional execution of actions attached to events
### What problem are you trying to solve?
Currently, users have limited control over when an element's event handler action should be executed. For example, they may want to run a button's onclick's actions only if a certain condition is met (e.g. a checkbox is checked), but there's no built-in way to do this.

### What solution do you propose?
Add a "conditional action" feature that allows users to specify a condition for when an event's action should be executed. This feature would involve adding a new attribute to each component called "runOnlyIf", which takes an expression as input.

The expression is evaluated before the action attached to the event runs in response to an event. If the expression returns a truthy value, the action is executed/run. But if it returns a falsy value, the action is skipped.

johnsoncherian:
Fixes  #6304


kavinvenkatachalam (approved):

withshubh:
docs updated here: https://github.com/ToolJet/ToolJet/pull/6868

Issue #6364, opened 2023-05-11
Add a new dropdown widget 


stepinfwd:
### Is your proposal related to a problem?
Current dropdown widget has few drawbacks 
Below mentioned enhancements can be done is same dropdown or as a new dropdown widget

### Describe the solution you'd like
Dropdown widget should be able to take in data as array of object with following format
`{{[	{label: 'One',value: 1,disable: false,visible: true},{label: 'Two',value: 2,disable: false,visible: true},{label: 'Three',value: 3,disable: false,visible: true}	]}}`

Add option to include placeholder to dropdown widget
option to hide or disable individual options

### If the feature is approved, would you be willing to submit a PR?
yes

PR #6392, opened 2023-05-15
Feature :: Dropdown widget enhancement

### LINKED ISSUE #5886: Add the Placeholder property to the Dropdown component
The component "Dropdown" do not offer the property to set a placeholder text.
image

Add the Placeholder property to the "Dropdown" component.

For inspiration, check out the "Text input" component's Placeholder property, expecting the same for the Dropdown component.

stepinfwd:
closes #5886 #6364


kavinvenkatachalam (changes requested):
@stepinfwd Instead of creating a new component inside the `Inspector` for dynamic properties, we can add one more property in the `widgetConfig`. Something like `conditionalRender` and pass on the properties to it which it needs to depend. We can have a call and discuss this.

frontend/src/Editor/Components/DropDown.jsx
```diff
@@ -16,22 +17,61 @@ export const DropDown = function DropDown({
...
+      setValues(schema.map((item) => item.value));
+      setDisabledItems(schema.map((item) => item.disable));
+      setVisibleItems(schema.map((item) => item.visible));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [advanced, schema]);
+
   if (!_.isArray(values)) {
-    values = [];
+    setValues([]);
```
kavinvenkatachalam, line 46:
```suggestion
  if (advanced) {
    values = schema.map((item) => item.value);
    display_values = schema.map((item) => item.label);
  } else if (!_.isArray(values)) {
    values = [];
  }
```

frontend/src/Editor/Components/DropDown.jsx
```diff
@@ -16,22 +17,61 @@ export const DropDown = function DropDown({
...
+                visible: visibleItems[index],
+              }
+            : {};
+        })
+        .filter((element) => {
+          if (Object.keys(element).length !== 0) {
+            return true;
+          }
+          return false;
+        }),
```
kavinvenkatachalam, line 74:
```suggestion
    selectOptions = advanced
      ? [...schema.filter((data) => data.visible).map((value) => value)]
      : [
          ...values.map((value, index) => {
            return { label: display_values[index], value: value };
          }),
        ];
```

kavinvenkatachalam (changes requested):
@stepinfwd Instead of all the changes in dropdown, the suggested changes alone is enough to work both advanced and older mode. You can revert the complete `DropDown.jsx` changes and add only the suggested changes. Please test all the scenarios

frontend/src/Editor/Components/DropDown.jsx
```diff
@@ -16,23 +16,44 @@ export const DropDown = function DropDown({
...
-  if (!_.isArray(values)) {
+  function findDefaultItem(schema) {
+    const foundItem = schema?.find((item) => item.default === true);
+    return foundItem ? foundItem.value : null;
+  }
+
+  if (advanced) {
+    values = schema?.map((item) => item.value);
+    display_values = schema?.map((item) => item.label);
+    value = findDefaultItem(schema);
```
kavinvenkatachalam, line 34: This is getting called every time when re-render happens. This can be avoided by wrapping with `useRef` and setting it to false after the first time. Then whenever the `schema` changes, this can be recalculated

frontend/src/Editor/Components/DropDown.jsx
```diff
@@ -109,9 +130,12 @@ export const DropDown = function DropDown({
...
 
   useEffect(() => {
-    setExposedVariable('optionLabels', display_values);
+    setExposedVariable(
+      'optionLabels',
+      schema?.filter((item) => item.visible)?.map((item) => item.label)
+    );
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [JSON.stringify(display_values)]);
+  }, [JSON.stringify(schema)]);
```
kavinvenkatachalam, line 138: We should have `display_values` and `advanced` as well. Whenever anything changes, we need expose `display_values` if `advanced` is false

frontend/src/Editor/Inspector/Utils.js
```diff
@@ -28,10 +29,24 @@ export function renderElement(
...
   const definition = paramTypeDefinition[param] || {};
 
+  if (componentConfig.component == 'DropDown') {
+    const paramTypeConfig = componentMeta[paramType] || {};
+    const paramConfig = paramTypeConfig[param] || {};
+    const { conditionallyRender = null } = paramConfig;
+
+    if (conditionallyRender) {
+      const { key, value } = conditionallyRender;
+      if (paramTypeDefinition?.[key]) {
```
kavinvenkatachalam, line 44:
```suggestion
      if (paramTypeDefinition?.[key] ?? value) {
```

Without value, it is rendering both `schema` and other values for older widgets

kavinvenkatachalam (changes requested):
@stepinfwd Few code suggestions. 

When the `advanced` toggle is turned off, the `optionLabels` exposed value is not getting updated. Take a look at this after the suggested changes

kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

PR #6394, opened 2023-05-15
[docs] global datasource restapi revamp

withshubh:


docs/docs/data-sources/restapi.md
```diff
@@ -3,44 +3,150 @@ id: restapi
...
 :::
+
+<div style={{textAlign: 'center'}}>
+
+<img className="screenshot-full" src="/img/datasource-reference/rest-api/preview.png" alt="ToolJet - Data source - REST API" />
+
+</div>
+
+## Response types
+
```
gsmithun4, line 66: Should mention this:
if response content-type is image, the response will be base64 string

docs/docs/data-sources/restapi.md
```diff
@@ -3,44 +3,150 @@ id: restapi
...
+REST APIs can return data in a variety of formats, including **JSON** and **Base64**. JSON is a common format used for data exchange in REST APIs, while Base64 is often used for encoding binary data, such as images or video, within a JSON response.
```
withshubh, line 67:
```suggestion
REST APIs can return data in a variety of formats, including **JSON** and **Base64**. JSON is a common format used for data exchange in REST APIs, while Base64 is often used for encoding binary data, such as images or video, within a JSON response.  
When the response `content-type` is **image**, the response will be a `base64` string.
```

gsmithun4 (approved):

PR #6396, opened 2023-05-15
Sanitising workspace folder name while creating or editing a folder

### LINKED ISSUE #5663: Sanitising workspace folder name while creating or editing a folder

Currently, user can give a name to a workspace folder but the name has defined length of 25 characters. 

Workspace admins should be able to add and edit workspace folder names. The name length needs to be of same standard across the product. 

In case of creating a new workspace folder and adding name to it : 
1) Name should be unique.
3) Name should be mandatory. 

In case of editing workspace folder name : 
1) Name should be unique.
3) Name should be mandatory. 

shah21:
Resolves #5663


server/src/dto/create-folder.dto.ts
```diff
@@ -1,12 +1,15 @@
...
   @IsString()
   @IsNotEmpty()
-  @Transform(({ value }) => sanitizeInput(value))
-  @MaxLength(25, { message: 'Folder name cannot be longer than 25 characters' })
-  @MinLength(0, { message: 'Folder name cannot be empty' })
+  @Transform(({ value }) => {
+    const newValue = sanitizeInput(value);
+    return newValue.trim();
+  })
+  @Matches('^[A-Za-z0-9 ]+$', '', { message: 'Folder name must contain only letters and numbers' })
```
gsmithun4, line 12: shall we add `_` and `-`?

frontend/src/HomePage/Folders.jsx
```diff
@@ -129,24 +129,17 @@ export const Folders = function Folders({
...
-    if (newFolderName?.trim().length > 25) {
-      toast.error('Folder name cannot be longer than 25 characters.');
-      return false;
-    }
-    return true;
-  }
-
   function executeEditFolder() {
-    if (validateName()) {
+    const folderName = newFolderName.trim();
```
gsmithun4, line 133:
```suggestion
    const folderName = newFolderName?.trim();
```

gsmithun4 (changes requested):

server/src/services/folders.service.ts
```diff
@@ -33,6 +38,11 @@ export class FoldersService {
   }
 
   async update(folderId: string, folderName: string): Promise<UpdateResult> {
+    const existed = await this.foldersRepository.findOne({ name: folderName });
+    if (existed) {
+      throw new ConflictException('Workspace name is already taken.');
```
gsmithun4, line 43: should be `Folder name is already taken`

server/src/services/folders.service.ts
```diff
@@ -22,6 +22,11 @@ export class FoldersService {
   ) {}
 
   async create(user: User, folderName): Promise<Folder> {
+    const existed = await this.foldersRepository.findOne({ name: folderName });
+    if (existed) {
+      throw new ConflictException('Workspace name is already taken.');
```
gsmithun4, line 27: should be `Folder name is already taken`

gsmithun4 (changes requested):

gsmithun4 (approved):

PR #6423, opened 2023-05-16
[enhancement] Sanitising app name while creating or editing an app

### LINKED ISSUE #5669: Sanitising app name while creating or editing an app

Currently, user can give a name to a workspace app but it breaks in certain cases. For example, when the length of name is massive. 

Workspace admins should be able to add and edit workspace apps names. The name length needs to be of same standard across the product. 

In case of creating a new workspace app and adding name to it : 
1) Name should be unique.
3) Name should be mandatory. 

In case of editing workspace app name : 

_Note - While renaming an app, the existing app name should be present.   Name should get updated to the new one after changing it. (currently clicking on the pencil icon doesn't do anything - need modification)_

1) Name should be unique.
3) Name should be mandatory. 

shah21:
Resolves #5669


server/src/services/apps.service.ts
```diff
@@ -186,12 +186,19 @@ export class AppsService {
     return await viewableAppsQb.getMany();
   }
 
-  async update(appId: string, appUpdateDto: AppUpdateDto, manager?: EntityManager) {
+  async update(appId: string, appUpdateDto: AppUpdateDto, organizationId: string, manager?: EntityManager) {
```
gsmithun4, line 189: can revert this changes if we adapt this
https://github.com/ToolJet/ToolJet/pull/6581

gsmithun4 (changes requested):

gsmithun4 (approved):

PR #6436, opened 2023-05-17
Row selection revamp

### LINKED ISSUE #5406: [Table] - Row selection revamp
The current Implementation of the row selection in the Table is confusing with highlighted and checkbox themes. So, we need to enhance the behavior of the row selection feature.

1. There will be two themes in the row selection one is the highlight theme and the second one is the checkbox theme. (these are the existing theme only)
2. There will be 3 properties related to row selection in the options inside table inspector properties. 2 out of 3 will be existing properties and will add one new property named  `Allow selection` which will be true by default. 
The `Highlight selected row` property will be accessible only when `Allow selection` will be true, else it will be hidden in the options. The `bulk selector` property will behave in an existing fashion.
3. Whenever `allow selection` is `true` and `highlighted selected row` is `false`, will have a checkbox theme, which will allow you to select only one row at a time.
4. When `allow selection` is true and `highlight selected row ` is also true at that time, will use the highlight theme, which means the checkbox will not be available, user can select a single row, upon clicking the respective row, and that entire row will be highlighted.
5. When `allow Selection` is true, `highlight selected row ` is true, and `bulk selector` is also true, at that time will allow the user to highlight multiple rows at the same time.
6. When allow selection is true, highlight selected row is false and bulk selection is true, then will have a checkbox theme and the user can select multiple checkboxes.
7. if allow selection is false and the bulk selector is true, will show a checkbox theme and the user can select multiple rows
8. Exposed Variables - 
selectedRows (current) - Array of objects - should show selected row data for both the themes and  single/bulk as well
selectedRowsId (new) - Array of object - should have indexes of all selected rows (same order as in selectedRows)
selectedRowId (current) - current behavior
selectedRow (current) - current behavior

Yes

manishkushare:
Resolves #5406


kavinvenkatachalam (changes requested):
@manishkushare I'm not able to unselect the row once it is selected. It's particularly on `Highlight selected row`. Can you check it out?

kavinvenkatachalam (changes requested):
@manishkushare Noticed one more thing, If I enable `Highlight selected row` and selected few rows. I can see the values exposed. But when I turn off highlight selected row, I can see the exposed value becomes null but the checkbox are selected. Similar case for checkbox to highlight. Can you check that?

image

frontend/src/Editor/Inspector/Components/Table/Table.jsx
```diff
@@ -923,7 +923,10 @@ class TableComponent extends React.Component {
     const useDynamicColumn = component.component.definition.properties.useDynamicColumn?.value
       ? resolveReferences(component.component.definition.properties.useDynamicColumn?.value, currentState) ?? false
       : false;
-
+    const allowSelection = component.component.definition.properties?.allowSelection?.value
+      ? resolveReferences(component.component.definition.properties.allowSelection?.value, currentState)
+      : resolveReferences(component.component.definition.properties.highlightSelectedRow.value, currentState) ||
+        resolveReferences(component.component.definition.properties.showBulkSelector.value, currentState);
```
kavinvenkatachalam, line 929: Instead of ternary we can use nullish coalescing. If the value of `allowSelection` is null then it will go to else statement
```suggestion
    const allowSelection = component.component.definition.properties?.allowSelection?.value
      ?? resolveReferences(component.component.definition.properties.highlightSelectedRow.value, currentState) ||
        resolveReferences(component.component.definition.properties.showBulkSelector.value, currentState);
```

frontend/src/Editor/Inspector/Elements/Code.jsx
```diff
@@ -14,7 +16,24 @@ export const Code = ({
   fxActive,
   component,
 }) => {
-  const initialValue = definition ? definition.value : '';
+  const getDefinitionForNewProps = (param) => {
+    if (['showAddNewRowButton', 'allowSelection'].includes(param)) {
+      if (param === 'allowSelection') {
```
kavinvenkatachalam, line 21: Need to check `allowSelection` is undefined here. Else `allowSelection` value will always be depend on `bulkSelection` and `highlightSelectedRow`

kavinvenkatachalam (changes requested):
@manishkushare Added few code changes. One more issue with the functionality,

If I disable `allowSelection` still I'm able to see the `selectedRowsId` in the exposed values. We need to restrict that

kavinvenkatachalam (approved):

kavinvenkatachalam (approved):

PR #6454, opened 2023-05-17
Bugfix - source options are lost on save in client

### LINKED ISSUE #6004: User should not have to enter client secrets or any other value again while editing a datasource

arpitnath:
Fixes #6004


akshaysasidrn:
@arpitnath won't the manifest changes break for existing apps that might be using twilio?

akshaysasidrn (approved):

PR #6464, opened 2023-05-18
guards to plugins API endpoints based on env

arpitnath:
Adds guards to plugins API endpoints based on the value of the **ENABLE_MARKETPLACE_FEATURE** environment variable. This ensures that the API is only accessible when the variable is set to 'true'.


arpitnath:
Ref: #6579

PR #6470, opened 2023-05-18
[fix] Incorrect edit informations is displayed on app card

### LINKED ISSUE #5554: [Dashboard screen]- Incorrect edit informations is displayed

If user is editing the existing application, the last edit information displayed on dashboard should reflect the same.
e.g. if user has created an application 4 days ago and making some edits today, the information should be displayed as follows:
"Edited 2 hours ago by XYZ"

  Instead of showing the latest updates specific information, system is showing the application created time as:
"Edited 4 days ago by XYZ"

1. create an application and verify the default information on dashboard
2. edit the application later and verify the same information
3. information should reflect the latest changes

shah21:
Resolves #5554

### DIFF
--- a/frontend/src/HomePage/AppCard.jsx
+++ b/frontend/src/HomePage/AppCard.jsx
@@ -52,7 +52,8 @@ export default function AppCard({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isHovered]);
 
-  const updated = moment(app.created_at).fromNow(true);
+  const updated_at = app?.editing_version?.updated_at || app?.updated_at;
+  const updated = moment(updated_at).fromNow(true);
   const darkMode = localStorage.getItem('darkMode') === 'true';
 
   let AppIcon;


frontend/src/HomePage/AppCard.jsx
```diff
@@ -52,7 +52,8 @@ export default function AppCard({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isHovered]);
 
-  const updated = moment(app.created_at).fromNow(true);
+  const updated_at = app?.editing_version?.updated_at || app?.updated_at;
```
gsmithun4, line 55: Question: are we showing last updated user?
Shows in UI that
updated just now by `A`.
where `A` is app created user, and recently edited by other user

gsmithun4 (changes requested):

frontend/src/HomePage/AppCard.jsx
shah21: @thehacklady

gsmithun4 (approved):

Issue #6437, opened 2023-05-17
Set a default country code on onboarding form


withshubh:
### What is the current behaviour?
On onboarding form for new users, the country code dropdown is blank on the mobile number page. 

Fix the issue by setting a default country code in the dropdown.

### Steps to reproduce
- Sign up on ToolJet: https://www.tooljet.com/signup
- Open the link received on your registered email
- The onboarding form will open up, on the last step user is asked to enter phone number 

### Screenshots or Screencast
![image (12)](https://github.com/ToolJet/ToolJet/assets/25361949/ea660cbc-9e8e-4509-9d4e-c1c9e30810af)

FourAndHalf:
As far as i looked it up, the only change needed is to change the useState condition in function OnBoardingForm

```js
function OnBoardingForm({ userDetails = {}, token = '', organizationToken = '', password, darkMode }) {
  const Logo = darkMode ? LogoDarkMode : LogoLightMode;
  const [page, setPage] = useState(0);
  const [completed, setCompleted] = useState(false);
  const [isLoading, setIsLoading] = useState(false);
  const [formData, setFormData] = useState({
    companyName: '',
    role: '',
    companySize: '',
```
```diff
-phoneNumber: '',
+phoneNumber: '+1',
```
```js
  });
```

if you think this may solve the problem, can you assign this issue to me.

withshubh:
The change is required in these two files.
![Screenshot 2023-05-17 at 2 33 36 PM](https://github.com/ToolJet/ToolJet/assets/25361949/2ea43ebb-4436-49aa-a142-aaf5f5be4a42)

wasimTQ:
@withshubh Instead of hardcoding values. Should we use [country.is/](https://country.is/) api and put the user's country in the `PhoneInput`?

CVamsi27:
> @withshubh Instead of hardcoding values. Should we use [country.is](https://country.is/) api and put the user's country in the `PhoneInput`?

I would like to work in this.

iSatVeerSingh:
@FourAndHalf I tried your fix and as we can see that your fix is not working. Here is a screenshot below.


![Screenshot from 2023-05-29 19-25-56](https://github.com/ToolJet/ToolJet/assets/85070570/8ef2461f-aa31-4faa-a1fa-3171f2f7ed39)

I found the solution for this and I am working on it.

Navaneeth-pk:
[...]

PR #6478, opened 2023-05-19
Modified the null placeholder in phonenumber field to +1 in both lightMode and darkMode pages

FourAndHalf:
Added default country code to the previously empty box
Fixes issue #6437

**Changed the values assigned as default from ' ' to '+1' on both pages** 
 
```jsx
  const [formData, setFormData] = useState({
    companyName: '',
    role: '',
    companySize: '',
```
```diff
- phoneNumber: ' ',
+ phoneNumber: '+1',
```
```jsx
   )};
```

The place for correction that was shared before in the earlier screenshot was actually on the page's form titles
```jsx
const FORM_TITLES = [
    `Where do you work ${userDetails?.name}?`,
    'What best describes your role?',
    'What is the size of your company?',
    'Enter your phone number',
    'Enter your phone number', //dummy for styling
  ];
```

I also looked up how to change the displayed flag, but it was set using a dependency library. I don't anything can be done about it by me. 
```jsx
import PhoneInput from 'react-phone-input-2';
import 'react-phone-input-2/lib/style.css';
```

### DIFF
--- a/plugins/packages/snowflake/package-lock.json
+++ b/plugins/packages/snowflake/package-lock.json
@@ -9,9 +9,9 @@
       "version": "1.0.0",
       "dependencies": {
         "@tooljet-plugins/common": "file:../common",
-        "@types/snowflake-sdk": "^1.6.2",
+        "@types/snowflake-sdk": "^1.6.12",
         "react": "^17.0.2",
-        "snowflake-sdk": "github:ToolJet/snowflake-connector-nodejs#chore/update-https-agent"
+        "snowflake-sdk": "^1.6.22"
       }
     },
     "../common": {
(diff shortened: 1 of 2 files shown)


frontend/src/OnBoardingForm/OnBoardingForm.jsx
```diff
@@ -22,7 +22,7 @@ function OnBoardingForm({ userDetails = {}, token = '', organizationToken = '',
     companyName: '',
     role: '',
     companySize: '',
-    phoneNumber: '',
+    phoneNumber: '+1',
```
iSatVeerSingh, line 25: I tried it and this is not working. You should set a default county as "US" in phone input.

frontend/src/OnBoardingForm/OnbboardingFromSH.jsx
```diff
@@ -28,7 +28,7 @@ function OnbboardingFromSH({ darkMode }) {
     email: '',
     password: '',
     workspace: '',
-    phoneNumber: '',
+    phoneNumber: '+1',
```
iSatVeerSingh, line 31: Also for this.

iSatVeerSingh:
Also update the line "Fixes issue https://github.com/ToolJet/ToolJet/issues/6437" to "Fixes #6437". So the issue can automatically close when PR get merge.

johnsoncherian:
@FourAndHalf this solution does not seem to be working. Try passing country directly in input field as mentioned in previous comments. Refer this documentation for better understanding:
https://www.npmjs.com/package/react-phone-input-2

johnsoncherian (changes requested):
@FourAndHalf this solution does not seem to be working. Try passing country directly in input field as mentioned in previous comments. Refer this documentation for better understanding:
https://www.npmjs.com/package/react-phone-input-2

PR #6489, opened 2023-05-22
Add support for  one level of nested data in the table component

### LINKED ISSUE #5926: [Table] Add support for nested data

When data with nested objects are added to the table, the table does not show the columns of the nested keys.

Adding the data with the nested objects to the table should also show columns of the nested objects.

1. Add a table on the canvas
2. On the table data property, add the following data:
```js
[
  {
    "student": {
      "name": "Anakin Skywalker",
      "saber": "blue"
    },
    "teacher": {
      "name": "Obi wan Kenobi",
      "saber": "blue"
    }
  },
  ...
]
```
3. The table will not load any columns

manishkushare:
Resolves #5926


kavinvenkatachalam (approved):

kavinvenkatachalam (changes requested):
@manishkushare Please change the variable name and the array exposed as discussed over the call

kavinvenkatachalam (approved):

PR #6559, opened 2023-05-26
fix: added duplicate check for page name and handle

### LINKED ISSUE #6503: 'Page handle already exists' error is displayed even after renaming the page
'Page with same handle already exists' error should not appear on the screen if user renames existing page and tries to add new page with same name

Error is being displayed on the screen even after page rename is performed by the user.

1. create multi page application
2. try to enter same page name and verify if error message is displayed on the screen
3. rename the page and try to add new page with the older name

### Screenshots or Screencast

https://github.com/ToolJet/ToolJet/assets/125240229/adc1c492-7b74-423a-b9a6-f935b5775f20

johnsoncherian:
Fixes #6503

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -1004,11 +1004,19 @@ class EditorComponent extends React.Component {
 
   addNewPage = ({ name, handle }) => {
     // check for unique page handles
-    const pageExists = Object.values(this.state.appDefinition.pages).some((page) => page.handle === handle);
+    const pageExists = Object.values(this.state.appDefinition.pages).some((page) => page.name === name);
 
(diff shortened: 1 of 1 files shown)


kavinvenkatachalam (changes requested):
@johnsoncherian Can we change the error being displayed when we try with the same name? User might not know the handle when they try to add a page. Error can be something like `'Page name already exists'`

johnsoncherian:
> @johnsoncherian Can we change the error being displayed when we try with the same name? User might not know the handle when they try to add a page. Error can be something like `'Page name already exists'`

This is sorted @kavinvenkatachalam. Please do take a look

kavinvenkatachalam (approved):

PR #6593, opened 2023-05-30
Move Editor's version manager regarding states to zustand store

### LINKED ISSUE #6628: [refactor]: Move version regarding states from editor to zustand store
### What area of the codebase is to be refactored?

Editor component

### What are the potential advantages of this proposed refactoring?

This refactor will help in reducing prop drilling at multiple nested level components

nakulnagargade:
Closes #6628


sherfin94:
Hey @nakulnagargade , please resolve the conflicts.

frontend/src/Editor/Comments.jsx
```diff
@@ -1,16 +1,16 @@
...
 import { commentsService } from '@/_services';
-
+import { useAppVersionStore } from '@/_stores/appVersionStore';
 import useRouter from '@/_hooks/use-router';
 
-const Comments = ({ newThread = {}, appVersionsId, socket, canvasWidth, currentPageId }) => {
+const Comments = ({ newThread = {}, socket, canvasWidth, currentPageId }) => {
   const [threads, setThreads] = React.useState([]);
   const router = useRouter();
+  const { appVersionsId } = useAppVersionStore((state) => ({ appVersionsId: state?.editingVersion?.id }), shallow);
```
johnsoncherian, line 13: do we need shallow here.? 
const appVersionsId = useAppVersionStore((state) => state?.editingVersion?.id);

We need to return only the version id string here right.?

nakulnagargade: yes, We won't need a shallow here since this is only a string. Kept it in case if we"ll derive more states objects from useAppVersionStore, But yeah we can add it when needed. Removing it.

frontend/src/Editor/Editor.jsx
```diff
@@ -243,25 +242,21 @@ class EditorComponent extends React.Component {
     }
   }
 
-  isVersionReleased = (version = this.state.editingVersion) => {
+  isVersionReleased = (version = useAppVersionStore?.getState()?.editingVersion) => {
```
johnsoncherian, line 245: I don't see this function used anywhere. Please check once.

johnsoncherian (approved):

johnsoncherian:
@nakulnagargade you may have to merge with latest development branch and resolve conflicts

kavinvenkatachalam:
@nakulnagargade Can you resolve the conflicts?

frontend/src/Editor/Container.jsx
```diff
@@ -563,9 +558,8 @@ export const Container = ({
                 currentPageId,
                 childComponents,
                 isVersionReleased,
-                setReleasedVersionPopupState,
+                enableReleasedVersionPopupState,
```
kavinvenkatachalam, line 561: It can be removed here. These props are being passed to the `SubContainer` but we're fetching it from the store already in `SubContainer`

kavinvenkatachalam (changes requested):
@nakulnagargade Few suggestions,

- While removing the components from the inspector, the release message is static and not animated. Every action on left side bar doesn't trigger the animation (`enableReleasedVersionPopupState` function is not getting called)
- `editingVersion` is being widely used in the Editor. It can be passed to HOC and can be used as a prop instead of fetching from the store every time. We can pass all the version items as props instead of fetching every time from the store. I feel it won't cause any unwanted re-renders
- `shallow` check is needed where ever you import the complete store. If not then the component will get re-render if some other state in the store changes

kavinvenkatachalam (changes requested):
@nakulnagargade Still few places were left out with shallow check. Can you check it?

nakulnagargade:
> 



> @nakulnagargade Still few places were left out with shallow check. Can you check it?

@kavinvenkatachalam Done

kavinvenkatachalam (changes requested):
@nakulnagargade There is some issue with editing the released app. When I edit the released app, I'm not able to see the`This is a released app. Create a new version to make changes.` message and getting an error could not save. Please take a look

frontend/src/_stores/appVersionStore.js
```diff
@@ -0,0 +1,25 @@
...
+        set({ editingVersion: version, isVersionReleased: get().releasedVersionId === version?.id }),
+      enableReleasedVersionPopupState: () => set({ isUserEditingTheVersion: true }),
+      disableReleasedVersionPopupState: () => set({ isUserEditingTheVersion: false }),
+      updateReleasedVersionId: (versionId) =>
+        set({
+          releasedVersionId: versionId,
+          isVersionReleased: get().editingVersion?.id ? get().editingVersion?.id === versionId : false,
+        }),
+    },
+  }))
```
kavinvenkatachalam, line 24: Including the sore name helps to debug faster
`{ name: 'App Version Store' }`

frontend/src/Editor/AppVersionsManager/ReleasedVersionError.jsx
```diff
@@ -1,11 +1,24 @@
...
+      isUserEditingTheVersion: state.isUserEditingTheVersion,
+      disableReleasedVersionPopupState: state.actions.disableReleasedVersionPopupState,
+    }),
+    shallow
+  );
+  const changeBackTheState = useCallback(() => {
+    isUserEditingTheVersion && disableReleasedVersionPopupState();
+  }, [isUserEditingTheVersion, disableReleasedVersionPopupState]);
+
+  React.useEffect(() => {
```
kavinvenkatachalam, line 18: Instead of using `React.useEffect` we can import the `useEffect` directly and use it for better readability. We can follow this as a standard.

kavinvenkatachalam (commented):
@nakulnagargade Requested a small change

frontend/src/_stores/appVersionStore.js
nakulnagargade: Done

frontend/src/Editor/AppVersionsManager/ReleasedVersionError.jsx
nakulnagargade: Done.

kavinvenkatachalam (approved):

PR #6608, opened 2023-05-30
Default value for annotation in bounded box widget

### LINKED ISSUE #5929: Bounding box cannot be prepopulated
### Version Information
v2.2.2
### Environment
Docker

I'm using BoundedBox. I want to populate it with bounding boxes I already have.

Then users can:
- leave the correct ones
- modify/delete the existing ones
- add new bounding boxes

When I programmatically add annotation object to the BoundBox object, it isn't shown on the image. I'd expect that it's shown on imagen (bug 1)

After programmatically adding an image to the annotations, if a new boundind box is drawn, it overwrites the programmatically driven one (bug 2).

1. Add Bound Box
2. Run javascript query to add an annotation programmatically:
```
var new_annotation = {}
new_annotation.type = 'POINT'
new_annotation.x = 20.0
new_annotation.y = 20.0
new_annotation.width = 10.0
new_annotation.height = 10.0
new_annotation.text = 'Tree'
new_annotation.id = 'd6b0e0ff-358f-4e34-b6e4-7d2fe5304ae5'

components.boundedbox_2.annotations.push(new_annotation)
```
3. The component now has the new annotation in it, but it's not shown on the image (Screenshot 1). This is the first bug.
4. Draw new bounding box on the image.
5. The new bounding box is shown on the image, but the programmatically added bounding box is removed from annotations (Screenshot 2).

### Screenshots or Screencast
Screenshot 1:
image

Screenshot 2:
image

stepinfwd:
closes #5929


kavinvenkatachalam (changes requested):
@stepinfwd We should have the default data property. Advanced and schema is not needed in this case

kavinvenkatachalam (changes requested):
@stepinfwd Changes in the `Utils` file can be removed and not needed

kavinvenkatachalam (changes requested):
@stepinfwd Few more things I noticed,

- If image value is removed, the default value is shown on the broken image. We can avoid that
- We can allow the empty value in the default value field. As of now, it accepts only array type

cc: @manishkushare

frontend/src/Editor/Components/BoundedBox/BoundedBox.jsx
```diff
@@ -18,6 +19,8 @@ export const BoundedBox = ({ properties, fireEvent, darkMode, setExposedVariable
         }),
       ]
     : [];
+  let outerDiv = document.getElementsByClassName('lmGPCf');
```
kavinvenkatachalam, line 22: This can be declared as `const` and moved inside the else case in useEffect

kavinvenkatachalam (changes requested):
@stepinfwd Requested a small change. Rest evenrything looks good to me

kavinvenkatachalam (approved):

kavinvenkatachalam (approved):

PR #6618, opened 2023-05-31
Bug fixed : export with two identical column titles fills data in wrong column

### LINKED ISSUE #6440: Table PDF export with two identical column titles fills data in wrong column
### Version Information
v2.5.0

### Environment
Docker

When exporting a `table` with two columns that have the identical column title to PDF, the data is filled in the wrong PDF column.
It seems that the code just searches for the first column with a matching title and fills it in there. Even overwriting old data.
Here is an example:

Table:
image

PDF:
image

Put data into correct column.

manishkushare:
Resolves #6440
Also, when the column key contains ` .` like `company.name`, in such cases entire columns are exported with empty cells (values).

### DIFF
--- a/frontend/src/Editor/Components/Table/Table.jsx
+++ b/frontend/src/Editor/Components/Table/Table.jsx
@@ -250,17 +250,17 @@ export function Table({
     let headers = columns.map((column) => {
       return { exportValue: String(column.exportValue), key: column.key ? String(column.key) : column.key };
     });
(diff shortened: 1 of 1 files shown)


withshubh:
Just commenting for notifcations

sherfin94 (approved):

Nandinisaha13:
The new add row does not work properly, needs to be tested again.

PR #6619, opened 2023-05-31
Default row selection in table on page reload

### LINKED ISSUE #3962: [Table]: Add an option/property to programmatically pre-select a table row on page load
Provide an option in the table widget that allows users to preselect a row when app initially loads

We've no such action available as of now.

1. Visit the [link](https://app.tooljet.com/) and create a new application.
2. Drag and drop table widget from the widgets inventory on the left
3. Now inspect the widget to access its properties
4. Toggle ON "Highlight SelectedRow"
5. If we click on any table row, it will be highlighted
6. If we refresh the app, selected row is disappears
7. Enable an option to select a row in the table when app initially loads  
### Screenshots or Screencast

Screenshot 2022-09-06 at 10 06 15 AM

manishkushare:
Resolves #3962


kavinvenkatachalam (changes requested):
@manishkushare Couple of issues I found,

1. The default selected row is not highlighted
2. Debugger throws an error if I pass an object but checked the code on `Table.jsx` and seems like it uses an object to select the row
3. It should be enabled only when `rowSelection` is TRUE

frontend/src/Editor/Inspector/Elements/Code.jsx
```diff
@@ -17,14 +17,17 @@ export const Code = ({
...
       if (param === 'allowSelection') {
         const highlightSelectedRow = component?.component?.definition?.properties?.highlightSelectedRow?.value ?? false;
         const showBulkSelector = component?.component?.definition?.properties?.showBulkSelector?.value ?? false;
         const allowSelection =
           resolveReferences(highlightSelectedRow, currentState) || resolveReferences(showBulkSelector, currentState);
 
         return '{{' + `${allowSelection}` + '}}';
+      } else if (param === 'defaultSelectedRow') {
+        console.log('kavin');
+        return `{{manish}}`;
```
kavinvenkatachalam, line 30: You can remove this

frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -783,6 +792,25 @@ export function Table({
...
+            toggleRowSelected(selectedRowId, true);
+            mergeToTableDetails({ selectedRow: selectedRow, selectedRowId: selectedRowId });
+          });
+        } else {
+          toggleRowSelected(selectedRowId, true);
+        }
+      }
+    }
+    //hack : in the initial render, data is undefined since, upon feeding data to the table from some query, query inside current state is {}. Hence we added data in the dependency array, now question is should we add data or rows?
+  }, [JSON.stringify(defaultSelectedRow), JSON.stringify(data)]);
```
kavinvenkatachalam, line 812: I think page data is enough instead of complete table data unless if there is any edgecase

kavinvenkatachalam (changes requested):
@manishkushare Few issues I noticed,

- When I turn on `Highlight selected row`, the default selected row got deselected
- When the default selected row is not in first page, it is not getting selected

kavinvenkatachalam (approved):

Issue #6603, opened 2023-05-30
Remove the text about query for the Toogle swtich column type in a Table component 


tejakummarikuntla:
### How to find the text to remove
1. Create a ToolJet app
2. Add a Table component from the components manager
3. Select the table and get into the properties sections of the table
4. In the Columns section, select any column to open the column properties.
5. Inside column properties, set the `Column type` to `Toggle switch`
6. You should be able to find the text `This query doesn't have any event handlers`

### Solution
Remove the text `This query doesn't have any event handlers`
![image](https://github.com/ToolJet/ToolJet/assets/34749692/44e3ce62-3a2f-45eb-a3f7-b36b083ba561)

gantanikhilraj:
I can take this @tejakummarikuntla , assign it to me.

tejakummarikuntla:
Hey @gantanikhilraj, there was an issue assigned to you, let's wait for others to pick this up.

Avi98:
“Tejakummarikuntla, would it be alright if I take care of this up? It seems straightforward.”

tejakummarikuntla:
Hey @Avi98, Thanks for your interest; assigned this issue to you.

PR #6620, opened 2023-05-31
removed event manager text

Avi98:
Fixes [#6603](https://github.com/ToolJet/ToolJet/issues/6603)

### DIFF
--- a/frontend/src/Editor/Inspector/EventManager.jsx
+++ b/frontend/src/Editor/Inspector/EventManager.jsx
@@ -26,6 +26,7 @@ export const EventManager = ({
   popOverCallback,
   popoverPlacement,
   pages,
+  hideEmptyEventsAlert,
 }) => {
   const dataQueries = useDataQueries();
   const [events, setEvents] = useState(() => component.component.definition.events || []);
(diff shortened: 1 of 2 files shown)


withshubh:
Hey @Avi98 - Please sign the CLA

Avi98:
@withshubh  i have signed CLA but still its in pending state. Please let me know what im missing

frontend/src/Editor/Inspector/EventManager.jsx
```diff
@@ -907,8 +907,6 @@ export const EventManager = ({
     );
   };
 
-  const componentName = componentMeta.name ? componentMeta.name : 'query';
-
   if (events.length === 0) {
```
johnsoncherian, line 910: Hi @Avi98 the ask here is to remove the text only in the column properties widget. This is a common component which is reused across app. So this change will cause that text to disappear everywhere rather than only in aforementioned widget. You can introduce a new prop to this component (eg: like `hideEmptyEventAlert`) and if its value is true/false hide the warning text. Then from the config popup of table u can pass it as true and make it hidden only for the specific widget.

Avi98: Thanks @johnsoncherian, done

johnsoncherian (approved):

kavinvenkatachalam (approved):

PR #6665, opened 2023-06-06
Using tab to navigate through editable cells on the table

### LINKED ISSUE #6490: Using `tab` to navigate through various fields in the  cells on the table

manishkushare:
Resolves #6490


kavinvenkatachalam (changes requested):
@manishkushare Seems like tab switching is not properly work on text and number data type. Can you take a look?

frontend/src/Editor/Components/Table/GenerateEachCellValue.jsx
```diff
@@ -100,13 +101,21 @@ export default function GenerateEachCellValue({
       }}
       className="w-100 h-100"
     >
-      {!isColumnTypeAction && columnTypeAllowToRenderMarkElement.includes(columnType) && showHighlightedCells ? (
+      {showHighlightedCells &&
+      globalFilter &&
```
kavinvenkatachalam, line 105: Why do we need this condition here?

kavinvenkatachalam (changes requested):
@manishkushare Able to navigate with `tab` key but I'm not able to change the value. I need to click on the cell again to update the value. Please take a look

kavinvenkatachalam (approved):

Issue #3981, opened 2022-09-08
can't connect to Postgresql provided by nhost.io


yureckey:
there is nhost.io service which provides Postgresql in the Amazon cloud, I'm not sure exactly what they use (not RDS, maybe K8S)
when I'm trying to connect to this db from other dashboard-building tools (Retool, Appsmith, Nocodb) it's working fine. But connecting from Tooljet isn't working

### Version Information
latest 

### Environment
ToolJet Cloud

### What is the expected behaviour?
working connection to Postgresql instance

### What is the current behaviour?
see screenshot of error

### How to reproduce the issue?

1. signup and create new project in the nhost.io
2. get connection string
3. try to connect 

### Screenshots or Screencast

![ScreenClip  147](https://user-images.githubusercontent.com/535466/189044575-47d81156-d463-4ca9-b071-1562ad60b490.png)

mslourens:
the startup argument `statement_timeout` is not supported by the Postgresql version used by Nhost.io, but it is passed by the Tooljet Postgresql plugin.

shivangT:
Hey @yureckey ,

Let me discuss this with the team. Getting back shortly. 
What PostgreSQL version are you using?

Thanks for waiting

yureckey:
it's PostgreSQL 14.5

piperubio:
Hello,

I am following this issue as I am also experiencing a similar problem with Postgres 15.3. Like @yureckey, I am unable to connect to Postgres provided by Digital Ocean Managed DB using ToolJet.

Is there any update on this issue? It would be great if I could get some guidance on how to resolve it or if there is any plan for a fix in a future update.

Thank you for your time and assistance.

PR #6700, opened 2023-06-12
Delete static postgres connection option "statement_timeout"

piperubio:
### Summary

This pull request removes the static `statement_timeout` option from the Postgres connection configuration in the `plugins/packages/postgresql/lib/index.ts` file.

### Motivation

The `statement_timeout` option was causing compatibility issues with certain versions of Postgres and certain managed Postgres services, as not all of them support this option. This issue was reported in [#3981](https://github.com/ToolJet/ToolJet/issues/3981). By removing this option, we aim to increase the compatibility of our tool with various Postgres setups.

### Changes

- Removed the `statement_timeout` option from the Postgres connection configuration.

### Testing

Please ensure that connections to various Postgres versions and services are still functioning as expected after this change.

### DIFF
--- a/plugins/packages/postgresql/lib/index.ts
+++ b/plugins/packages/postgresql/lib/index.ts
@@ -90,7 +90,6 @@ export default class PostgresqlQueryService implements QueryService {
       database: sourceOptions.database,
       password: sourceOptions.password,
       port: sourceOptions.port,
-      statement_timeout: 10000,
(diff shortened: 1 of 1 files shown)


akshaysasidrn:
@piperubio Sorry, this has been pending review for a while. We can't completely avoid statement timeout as the default value might not suffice for users. Can we rather make this filed configurable? ie. have the use set the value in datasource connection options?

PR #6729, opened 2023-06-14
bug fixed : changeSet getting updated without changing the value in the table

### LINKED ISSUE #6728: [ Table ] : save / discard button is accessible even if we the cell value is not changed upon on blur
### Version Information
v2.6.2
### Environment
Development

When any column is editable in the table, in that particular column when clicking inside the particular cell, and instead of changing any value, we simply drag the mouse away from it (without making any changes in value), then changeSet should not get updated, and thus save and discard button should not be accessible.

In the above-mentioned scenario for column-type text, even if we don't change the value inside the cell, still change set gets updated and thus save and discard button is accessible even if the value is not edited.

1. Create a new app.
2. Drag a table from the right sidebar onto the canvas
3. Now click anywhere on the table to open the table widget inspector in the right sidebar itself.
4. Look for the columns section and click on any one of the cards in the section to open its settings pop-up window.
5. In that window, change the respective type of column to text
6. Now, on that canvas, click inside any cell from that particular column whose type is changed to text and then move away the cursor from that particular cell, as soon as you did that, you can see the save and discard button being accessible. Those buttons are only accessible if we edit the value in any one if the cell

### Screenshots or Screencast
[My app 1686754744072 - Tooljet - 14 June 2023 - Watch Video](https://www.loom.com/share/d220160cb0e14ffe9f362fdd96db6227)

Yes

manishkushare:
Resolves #6728

### DIFF
--- a/frontend/src/Editor/Components/Table/columns/index.jsx
+++ b/frontend/src/Editor/Components/Table/columns/index.jsx
@@ -164,10 +164,7 @@ export default function generateColumnsData({
                       }
                     }}
                     onBlur={(e) => {
-                      if (
-                        e.target.defaultValue !== e.target.value &&
-                        e.target.defaultValue.replace(/\n/g, '') !== e.target.value // do we require this, I have a doubt, need to add this condition, only when we change the column data type to string from text
-                      ) {
+                      if (e.target.defaultValue !== e.target.value) {
                         handleCellValueChange(
                           cell.row.index,
                           column.key || column.name,
--- a/frontend/src/Editor/Components/Table/columns/index.jsx
+++ b/frontend/src/Editor/Components/Table/columns/index.jsx
@@ -164,7 +164,10 @@ export default function generateColumnsData({
                       }
                     }}
                     onBlur={(e) => {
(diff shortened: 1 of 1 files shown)


frontend/src/Editor/Components/Table/columns/index.jsx
```diff
@@ -164,7 +164,10 @@ export default function generateColumnsData({
                       }
                     }}
                     onBlur={(e) => {
-                      if (e.target.defaultValue !== e.target.value) {
+                      if (
+                        e.target.defaultValue !== e.target.value &&
+                        e.target.defaultValue.replace(/\n/g, '') !== e.target.value // do we require this, I have a doubt, need to add this condition, only when we change the column data type to string from text
```
kavinvenkatachalam, line 169: I checked for the string and default datatype. Seems we're getting the right value and this can be removed

kavinvenkatachalam (changes requested):
@manishkushare Added a comment. Let me know if there is any concern

kavinvenkatachalam (approved):

PR #6735, opened 2023-06-15
Feature/harperdb plugin

ganesh8056:
New Plugin for HarperDB. In this PR we have enabled only the SQL option.


marketplace/plugins/harperdb/lib/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  async getConnection(sourceOptions: SourceOptions): Promise<any> {
+    const { host, username, password } = sourceOptions
+
+    const DB_CONFIG = {
+      harperHost: host,
+      username: username,
+      password: password,
+      token: '',
```
akshaysasidrn, line 50: we can skip setting token and schema here since we don't provide that option at the UI end

marketplace/plugins/harperdb/lib/index.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (mode === 'sql') {
+        const { sql_query } = queryOptions;
+        const sql_operation_response = await harperdbClient.query(sql_query);
+        result = sql_operation_response?.data;
+      }
+
+      if (mode === 'nosql') {
+        switch (operation) {
+          case 'insert':
+            const insert_response = await harperdbClient.insert({
```
akshaysasidrn, line 22: here as well

marketplace/plugins/harperdb/lib/index.ts
```diff
@@ -0,0 +1,75 @@
...
+export default class Harperdb implements QueryService {
+  async run(sourceOptions: SourceOptions, queryOptions: QueryOptions, dataSourceId: string): Promise<QueryResult> {
+    const harperdbClient = await this.getConnection(sourceOptions);
+    const { operation, mode } = queryOptions;
+    let result = {};
+
+    try {
+      if (mode === 'sql') {
+        const { sql_query } = queryOptions;
+        const sql_operation_response = await harperdbClient.query(sql_query);
```
akshaysasidrn, line 15: please camel cases

akshaysasidrn (changes requested):

marketplace/plugins/harperdb/lib/index.ts
ganesh8056: Screenshot 2023-06-16 at 5 26 12 PM

In Harperive SDK docs they have given either pass token / username& password, and Schema is optional. But in Typedefinition it is mandatory. So token and schema is given as Empty string.

Screenshot 2023-06-16 at 5 34 18 PM

marketplace/plugins/harperdb/lib/index.ts
ganesh8056: I have updated the Logic bit. Like instead of storing response in a `variable` - `sql_operation_response `, now directly storing it in result. Only on final return data has been taken and sent like `data: result?.data ?? {}`

marketplace/plugins/harperdb/lib/manifest.json
```diff
@@ -0,0 +1,53 @@
...
+        "type": "string"
+      },
+      "password": {
+        "type": "string",
+        "encrypted": true
+      }
+    }
+  },
+  "defaults": {},
+  "properties": {
```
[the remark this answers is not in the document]
ganesh8056: My understanding is ( We need separate option in UI  for host, port and protocol )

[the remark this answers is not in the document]
akshaysasidrn: We already have host right? 
So for protocol since its only HTTP we can add an SSL toggle and another field for specifying port.

marketplace/plugins/harperdb/lib/index.ts
akshaysasidrn: I thought giving an empty string might make it pick those values up on authentication. Ignore my comment. 👍🏾

akshaysasidrn (approved):

PR #6751, opened 2023-06-15
feature added for determining maximum cell height in the table

### LINKED ISSUE #6284: [table] set maximum height for the cell

There should be an option to set the maximum height for the table cell, so that whenever the text is large it gets overflow.

Currently, there is no option to set the maximum height for the table cell, causing the cell size to get big if the text is large. 

1. drag a table
2. create a query that returns the data that has large strings
3. load it on the table

### Screenshots or Screencast
Screenshot 2023-05-08 at 5 09 28 PM

manishkushare:
Resolves #6284


sherfin94:
Hey @manishkushare , the margin between scrollbar and the cell should be zero or minimal

PR #6753, opened 2023-06-15
Allow multi-line editing of Text column in Table component

dlebech:
This PR adds support for multi-line editing of the Text type column in the Table component by:

1. When holding the Shift key, hitting Enter adds a new line instead of saving the changes. This is a common pattern in many text inputs (especially chat inputs), where Enter is used to save the current changes by default.
2. Adjusting the CSS `white-space` for <textarea> so newlines are visible. Before, it would add the newline, but all the text was collapsed and it was not visible.

Thank you.


manishkushare (approved):

dlebech:
Thank you for the review last month @manishkushare. Is there anything else you need before merging this?

PR #6787, opened 2023-06-20
Feature :: Form widget with JSON input

### LINKED ISSUE #5821: Ability to generate fields of the Form component with JSON schema
Based on requirements, the fields of the form get changed, need to set the fields of the form with a JSON schema so that the JSON can be dynamic and so the form. 

Add an ability to Form components to take JSON schema as input and generate fields in the form accordingly.

stepinfwd:
closes #5821


kavinvenkatachalam (changes requested):
@stepinfwd Can you test all the components once? Seems like it's not working properly (Tried with `DatePicker` & `CheckBox`)

sherfin94:
Hey @stepinfwd , when I tried out, it seemed to crash the existing form widgets, see if you're bumping into the same issue.

To reproduce, goto develop, add a form widget, switch to your branch.

sherfin94:
Hey @stepinfwd , lets call this `Use custom schema` instead of `Advanced`: 
image

sherfin94:
Also @stepinfwd , adding a ` ` after the starting `{{` seems to highlight the entire default schema well, lets do that also: 
image

frontend/src/Editor/Components/Form/Form.jsx
```diff
@@ -0,0 +1,273 @@
...
+  useEffect(() => {
+    setUIComponents(generateUIComponents(JSONSchema, advanced, componentTypes));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [JSON.stringify(JSONSchema), advanced]);
+
+  const checkJsonChildrenValidtion = () => {
+    for (const key in childrenData) {
+      if (childrenData[key]?.isValid == false) return false;
+    }
+    return true;
```
kavinvenkatachalam, line 97: `every` method can be used here

frontend/src/Editor/Components/Form/Form.jsx
```diff
@@ -0,0 +1,273 @@
...
+          onEvent('onSubmit', { component }).then(() => resetComponent());
+        } else {
+          fireEvent('onInvalid');
+        }
+      }
+    } else if (buttonComponentId == uiComponents.length - 1 && JSONSchema.hasOwnProperty('submitButton')) {
+      if (isValid) {
+        onEvent('onSubmit', { component }).then(() => resetComponent());
+      } else {
+        fireEvent('onInvalid');
```
kavinvenkatachalam, line 172: Can we have a common function instead of repeating the logic?

frontend/src/Editor/Components/Form/Form.jsx
```diff
@@ -0,0 +1,273 @@
...
+          delete result[textKey].name;
+        }
+      }
+    }
+
+    return result;
+  };
+
+  useEffect(() => {
+    setUIComponents(generateUIComponents(JSONSchema, advanced, componentTypes));
```
kavinvenkatachalam, line 89:
```suggestion
    setUIComponents(generateUIComponents(JSONSchema, advanced));
```

`componentTypes` can be imported directly in the utils file

kavinvenkatachalam (changes requested):
@stepinfwd If I try to delete any style then it is vreaking the form. We can make the styles as optional. Please take a look.

And remove the single quote from the value key to differentiate between the value and key

kavinvenkatachalam (changes requested):
@stepinfwd Few minor suggestions,

1. Please remove the console statement in `Form.jsx`
2. Seems like the background color on canvas is changed in this PR. Please look into it

kavinvenkatachalam (approved):

kavinvenkatachalam (changes requested):
@stepinfwd Can we have a common function to test the visibility?

kavinvenkatachalam (approved):

PR #6802, opened 2023-06-21
Bugfix :: Box shadow is getting applied to the surrounding box 

### LINKED ISSUE #5959: box-shadow should be applicable on the border radius and not the actual component shape

When we change the border radius of a component(let's say button) the box shadow should be applied on the current shape and not the actual shape of the component.

When we change the border radius of a component the box shadow should is being applied to the actual shape of the component.

1. Drag button on the canvas
2. Change border radius
3. Add a box-shadow

### Screenshots or Screencast
Screenshot 2023-04-06 at 7 43 56 PM

stepinfwd:
closes #5959

### DIFF
--- a/frontend/src/Editor/Box.jsx
+++ b/frontend/src/Editor/Box.jsx
@@ -284,13 +284,13 @@ export const Box = function Box({
         style={{
           ...styles,
           backgroundColor,
-          boxShadow: validatedGeneralStyles?.boxShadow,
         }}
         role={preview ? 'BoxPreview' : 'Box'}
       >
         {inCanvas ? (
           !resetComponent ? (
             <ComponentToRender
+              boxShadow={validatedGeneralStyles?.boxShadow}
               onComponentClick={onComponentClick}
               onComponentOptionChanged={onComponentOptionChanged}
               currentState={currentState}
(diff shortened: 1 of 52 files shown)


frontend/src/Editor/Box.jsx
```diff
@@ -284,13 +284,13 @@ export const Box = function Box({
...
           ...styles,
           backgroundColor,
-          boxShadow: validatedGeneralStyles?.boxShadow,
         }}
         role={preview ? 'BoxPreview' : 'Box'}
       >
         {inCanvas ? (
           !resetComponent ? (
             <ComponentToRender
+              boxShadow={validatedGeneralStyles?.boxShadow}
```
nakulnagargade, line 293: @stepinfwd just curious can we send boxShadow as a part of styles props itself since styles is already being sent to every component as prop. This can save us from passing another prop (boxShadow).

<ComponentToRender
 styles = {{...validatedStyles, boxShadow : validatedGeneralStyles?.boxShadow}}
 />

nakulnagargade (approved):

sherfin94:
@stepinfwd , few issues that I've bumped into:

For datepicker, the box shadow is not correct (not adapting to border radius): 
image

For datepicker, the box shadow is not working:
image

For dropdown, the box shadow is not correctly changing according to its border radius:
image

Lets make sure that, at least for the commonly used components, box shadow adapts correctly the border radius

nakulnagargade (approved):

PR #6805, opened 2023-06-21
Fix for nested widget issue on cloned pages

### LINKED ISSUE #6297: Implement a clone/duplicate option for pages

johnsoncherian:
Fixes #6297

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -1352,9 +1352,9 @@ class EditorComponent extends React.Component {
 
     if (!name || !handle) return;
 
-    const queryParamsString = queryParams.map(([key, value]) => `${key}=${value} `).join('&');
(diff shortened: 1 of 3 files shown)


kavinvenkatachalam (changes requested):
@johnsoncherian This is not working for Tabs and Calendar component. We have a similar implementation on copy/cut components. You can refer here - https://github.com/ToolJet/ToolJet/blob/f67e3a929a8faaa0e00c869a6768a2a0d5a384c3/frontend/src/_helpers/appUtils.js#L1378

kavinvenkatachalam (approved):

Issue #6801, opened 2023-06-21
[docs]: move `visibility` and `disable` to Styles table


withshubh:
### Summary

Move the Visibility and disable's title and description on the Styles table.
![Screenshot 2023-06-21 at 5 59 09 PM](https://github.com/ToolJet/ToolJet/assets/25361949/43b6fe51-7f97-4fa5-92b3-a517c3dff521)

Note: The following change is required in all the versions of the docs: [All the versions](https://github.com/ToolJet/ToolJet/tree/develop/docs/versioned_docs) from `1.x.x` to `2.7.0` and the base [docs/docs](https://github.com/ToolJet/ToolJet/tree/develop/docs/docs)

### Issue Type

Documentation bug

### The entire URL of the documentation with the issue

https://docs.tooljet.com/docs/widgets/text/#visibility

### If the issue is confirmed, would you be willing to submit a pull request?

No

### Code of Conduct

- [X] I agree to follow the ToolJet Code of Conduct

ShivangRawat30:
I would like to take this one

Lovepreet013:
I would like to take this issue

PR #6807, opened 2023-06-21
moved Visibility and disable to Styles table

ShivangRawat30:
solves #6801 
Moved `Visibility` and `disable` to Styles table


withshubh (changes requested):
Please make the same changes in the docs/versioned_docs folder in each version

Issue #5981, opened 2023-04-10
Add support to generate `PDF` file using generate file action


withshubh:
### Is your proposal related to a problem?

When we add an action to generate a file to the event handler, we only get two options: `CSV` and `Text`.  It should also include an option to generate a `PDF` file from the provided data(array of objects)

### Describe the solution you'd like

- Add a button on the canvas
- Add an event handler -> select `Generate File action`
- On the `Type` dropdown, you'll find `CSV` and `Text`

Add an option to select the Type as `PDF` and provide the name and data dynamically.

### If the feature is approved, would you be willing to submit a PR?
No

sumitshinde-84:
Can you provide playground link ?

withshubh:
> Can you provide playground link ?

Hey @sumitshinde-84 - I did not understand, could you please tell what's playground link?

tejakummarikuntla:
If you are asking about the ToolJet application builder, you should be able to find it once your sign in and create a new app,
Could you elaborate on the question?

sumitshinde-84:
![toolJet](https://user-images.githubusercontent.com/110285294/231458020-d97f5d2b-2271-4259-84c9-e0be7f1cecb8.png)

Is it correct that this button needs a "Generate PDF" feature?

Sonali-Behera:
Hey, I would love to work on this issue. Please assign this to me.

withshubh:
Hey @Sonali-Behera - Thanks for showing interest in contributing to ToolJet. Assigned the issue to you.
Please feel free to join our [slack community](https://tooljet.com/slack) if you have any questions.

mud1tx:
@withshubh I can work on this issue can you please assign me?

withshubh:
Hey @Sonali-Behera - Any update on the PR?

withshubh:
Unassigned @Sonali-Behera since there was no update for a long time.

nemarugommula:
Hey @withshubh, I'd like to work on this issue, I have changes ready in my branch, Could you assign it to me? Thanks!

PR #6809, opened 2023-06-21
Issue : Adding support to generate PDF file using generate file action

nemarugommula:
Resolves : https://github.com/ToolJet/ToolJet/issues/5981
 - Added an option to show PDF option to generate files as pdf.
 - Modified existing methods in appUtils.js and generae_fill.js to support pdf type 
 
 Image : 
 
Screenshot 2023-06-22 at 7 17 55 PM

Screenshot 2023-06-22 at 7 13 56 PM

Screenshot 2023-06-22 at 7 17 13 PM

Screenshot 2023-06-22 at 7 17 26 PM

### DIFF
--- a/frontend/src/Editor/Inspector/EventManager.jsx
+++ b/frontend/src/Editor/Inspector/EventManager.jsx
@@ -467,6 +467,7 @@ export const EventManager = ({
                       options={[
                         { name: 'CSV', value: 'csv' },
                         { name: 'Text', value: 'plaintext' },
+                        { name: 'PDF', value: 'pdf' }
                       ]}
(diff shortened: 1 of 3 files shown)


manishkushare:
@nemarugommula , when we feed table data as an input to the data field of the pdf generator, it's not working as intended, can you please check it? Please look at how the CSV file generation works, we want similar functionality while generating pdf files. Right now, your current implementation will only generate pdf files, if we provide string data type to the data field of pdf file generation. Data can be anything string or array of objects.

manishkushare (approved):

PR #6810, opened 2023-06-21
Feature - Workspace environment constants

### LINKED ISSUE #6726: Workspace Variable / Constants Revamp - Phase 2
Description
Workspace constants will be environment-specific and resolved server-side.

Environment-specific means that the value of a workspace constant can be different for different deployment environments. For example, the value of the API_KEY constant might be different for the production environment and the staging environment. This allows users to define different values for constants depending on the environment in which their application is running.
Resolved server-side means that the value of a workspace constant is not resolved until it is accessed on the server-side. This means that the value of the constant is not exposed to the client-side, which helps to improve security.
Workspace admins can create/update/delete constants.
Workspace constants can be used in different apps and GDS connection forms across the same Workspace.
Integration with Global Data Sources connection forms: Integrate constant resolution functionality into the GDS connection form, enabling users to seamlessly utilize constants when building applications.

arpitnath:
Resolves #6726


server/src/services/organization_constants.service.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  async allEnvironmentConstants(organizationId: string): Promise<OrganizationConstant[]> {
+    return await dbTransactionWrap(async (manager: EntityManager) => {
+      const query = manager
+        .createQueryBuilder(OrganizationConstant, 'organization_constants')
+        .leftJoinAndSelect('organization_constants.orgEnvironmentConstantValues', 'org_environment_constant_values')
+        .where('organization_constants.organization_id = :organizationId', { organizationId });
+      const result = await query.getMany();
+
+      const appEnvironments = await this.appEnvironmentService.getAll(organizationId, null);
```
gsmithun4, line 28:
```suggestion
      const appEnvironments = await this.appEnvironmentService.getAll(organizationId);
```

gsmithun4: since the function accepts manager, and manager is available here, we can pass manager obj
```
const appEnvironments = await this.appEnvironmentService.getAll(organizationId, manager);
```

gsmithun4 (changes requested):

gsmithun4 (approved):

Issue #6759, opened 2023-06-16
On pressing enter should save the new workspace name


withshubh:
### What is the issue?
When a user edits a workspace name, a modal pops up where the user enters the `new workspace name` and clicks on the `save` button to rename the workspace.
Ideally, The user should be able to save the new name by pressing the enter key after entering the new workspace name.

### How to reproduce the issue?

1. Go to the ToolJet dashboard
2. From the workspaces dropdown at the bottom, click on the edit button next to the selected workspace
3. On the modal, rename the workspace name and press enter

### Screenshots or Screencast
![Screenshot 2023-06-16 at 2 31 35 PM](https://github.com/ToolJet/ToolJet/assets/25361949/90001602-e11f-4614-a477-eae4784fe118)

similar to this issue: https://github.com/ToolJet/ToolJet/issues/6653

gokullan:
Hi, there! I would like to work on this. Can this be assigned to me?

omarPerez1987:
Hello! I would like to work on this. Can I be assigned this would be my first contribution to an open source project?

withshubh:
Hey @gokullan - Thanks for showing interest in contributing to ToolJet. Assinged the issue to you.

withshubh:
> Hello! I would like to work on this. Can I be assigned this would be my first contribution to an open source project?

Hey, @omarPerez1987 - Thanks for showing interest in contributing to ToolJet, unfortunately, we can't assign this issue to you as we assign on a first come first served basis. Please comment on the other open good first issues, I would be happy to assign them to you.

PR #6811, opened 2023-06-21
Added functionality to save edited workspace name on pressing 'Enter'

gokullan:
This PR addresses #6759.

**Description**:
An `onKeyPress` event is added to the input to check for an Enter key press. If yes, the [`editOrganization` function](https://github.com/ToolJet/ToolJet/blob/c55e83eec94c42a02524602a83c47c629435a183/frontend/src/_components/OrganizationManager/EditOrganization.jsx#L18) is invoked which makes the required update.

### DIFF
--- a/frontend/src/_components/OrganizationManager/EditOrganization.jsx
+++ b/frontend/src/_components/OrganizationManager/EditOrganization.jsx
@@ -48,6 +48,12 @@ export const EditOrganization = ({ showEditOrg, setShowEditOrg, currentValue }) 
     setNewOrgName(e.target.value);
   };
 
+  const handleKeyDown = (e) => {
+    if (e.keyCode === 13) {
+      editOrganization();
+    }
+  };
+
   const closeModal = () => {
     setShowEditOrg(false);
     setErrorText('');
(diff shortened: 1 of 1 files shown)


frontend/src/_components/OrganizationManager/EditOrganization.jsx
```diff
@@ -48,6 +48,12 @@ export const EditOrganization = ({ showEditOrg, setShowEditOrg, currentValue })
     setNewOrgName(e.target.value);
   };
 
+  const handleKeyDown = (e) => {
```
arpitnath, line 51:
```suggestion
  const handleKeyDown = (e) => {
    if (e.key === 'Enter') {
      e.preventDefault();
      editOrganization();
    }
  };
```
Instead of using the keyCode property, it's recommended to use the key property to handle key events. The keyCode property is deprecated, and the key property provides a more consistent and readable way to handle specific keys.

arpitnath (changes requested):
`e.preventDefault();` 
Prevent form submission if the input is part of a form element

arpitnath (dismissed):

arpitnath:
Please fix the error : https://github.com/ToolJet/ToolJet/actions/runs/5455411687/jobs/9926821554?pr=6811#step:5:291

gokullan:
Hey, sorry about that. I have fixed it now.

arpitnath (approved):

PR #6824, opened 2023-06-22
Improvement - adds ability to resolve workspace variables in GDS connection forms

### LINKED ISSUE #6352: Ability to use workspace variables in datasource configuration - Phase 1
At present, the usage of workspace variables is limited to the app builder within our system.

The proposed request aims to expand the capability of utilizing workspace variables in the connection forms of Global Datasources. This enhancement would provide the ability to leverage workspace variables within the context of Global Datasource configurations.

arpitnath:
Resolves #6352


server/src/services/data_queries.service.ts
```diff
@@ -127,6 +127,8 @@ export class DataQueriesService {
       organizationId
     );
 
+    // console.log('sourceOptions ===> ', sourceOptions);
```
gsmithun4, line 130: please remove comments

gsmithun4 (changes requested):

Issue #6894, opened 2023-07-03
Increase the user session time to 10 days


gsmithun4:
### Is your proposal related to a problem?
The session time out is set to 2 days now

### Describe the solution you'd like
Increase it to 10 days

### If the feature is approved, would you be willing to submit a PR?
No

PR #6895, opened 2023-07-03
Increased user session time to 10 days

AnantshreeChandola:
Increased user session time from 2 to 10 days.
Resolves https://github.com/ToolJet/ToolJet/issues/6894


gsmithun4 (changes requested):
please make the value empty in `.env.example`

gsmithun4 (approved):

PR #6903, opened 2023-07-04
Feature :: Restrict certain components from being added onto certain container-components

### LINKED ISSUE #6834: Restrict certain components from being added onto certain container-components

Right now, in all the components that has sub-container, any component can be added on to.
But in various circumstances, this can lead to unmanageably complicated scenarios. Eg: Calendar component inside Kanban component's cards.

We have to be able to define which all components are not allowed wherever a component contains a container.

The following table describes the restrictions:

| Container-component      | Restricted components |
| ----------- | ----------- |
| Container      |    Calendar, Kanban   |
| Listview   | Calendar, Kanban        |
| Modal   |  Calendar, Kanban      |
| Tabs   |    Calendar, Kanban     |
| Form   |     Kanban, calendar, modal, container, ListView, Tabs, Form    |
| Kanban - card   |      Calendar, Kanban, Form, Tabs, Modal, ListView, Container, Table   |
| Kanban - popout |  Calendar, Kanban |
| Calendar - popout   |     Calendar, Kanban    |

If user attempts to add a component that is not allowed to be added, an error message needs to be shown.
There is no need to add restrictions to existing apps, restriction should only be checked at component-dropping time.

stepinfwd:
fixes #6834  
`appbuilder-1.2`


kavinvenkatachalam (changes requested):
@stepinfwd Instead of passing `restrictedKey` to every component can we maintain the componentType as a key in the `restrictedWidgetsObj`?

kavinvenkatachalam (changes requested):
@stepinfwd Implementation wise it looks good! One suggestion in the UI - the error message is should be use `break-word` instead of `break-all` in this case

image

kavinvenkatachalam (approved):

sherfin94:
Hey @stepinfwd , adding table to kanban - card is freezing the app, please restrict table also in that.

cc: @kavinvenkatachalam

PR #6907, opened 2023-07-04
Fixed widget position calculation on resize

### LINKED ISSUE #6905: Components are getting slightly displaced from the grid when they are repositioned or resized
### Environment
Everywhere
When components are resized or repositioned, the borders of the component should always perfectly snap to the grid lines

There is slight displacement from the grid lines at times

1. Add two text input components one below the other, in alignment
2. Slight resize the component in the bottom a few times, width wise
3. Resize it back to the same width as the first one
4. Observe the slight difference in width

johnsoncherian:
Fixes #6905


frontend/src/Editor/SubContainer.jsx
```diff
@@ -383,17 +388,14 @@ export const SubContainer = ({
 
     let { left, top, width, height } = boxes[id]['layouts'][currentLayout] || defaultData;
 
-    const canvasBoundingRect = parentRef.current.getElementsByClassName('real-canvas')[0].getBoundingClientRect();
-    const subContainerWidth = canvasBoundingRect.width;
-
     top = y;
-    if (deltaWidth !== 0) {
+    if (deltaWidth !== 0 && direction === 'left') {
```
kavinvenkatachalam, line 392:
```suggestion
    if (deltaWidth !== 0 ) {
```

Seems `direction` check is not needed here. Tried without that and working for me

frontend/src/Editor/Container.jsx
```diff
@@ -296,11 +299,15 @@ export const Container = ({
...
+    width = Math.round(width + (deltaWidth * NO_OF_GRIDS) / canvasWidth); // convert the width delta to percentage
     height = height + deltaHeight;
 
     top = y;
-    left = (x * 100) / canvasWidth;
+    if (direction === 'left') {
+      // onResizeStop is triggered for a single click on the border, therefore this conditional logic
+      // should not be removed.
+      left = (x * 100) / canvasWidth;
+    }
```
kavinvenkatachalam, line 310:
```suggestion
    left = (x * 100) / canvasWidth;
```

Even here the check is not needed

kavinvenkatachalam (commented):
@johnsoncherian On the implementation wise it is looks good. Added few suggestions, please check and let me know if I'm wrong

frontend/src/Editor/Container.jsx
johnsoncherian: Hi Kavin, Yes feature works without the checks. Added these conditions checks since in these cases the left position of the widget is not expected to change. But we can remove them.

kavinvenkatachalam (approved):

PR #6954, opened 2023-07-07
Listview component with grid option

### LINKED ISSUE #6891: Grid mode for ListView component

Presently, the ListView component does not allow items to be displayed in a Grid

1. Add a property called `Mode` in ListView
  - It should have two options `List` and `Grid` (Dropdown)
  - Default should be `List`
2. When `Grid` mode is chosen
  - There should be another numerical property called `Columns` which defaults to `3`
  - The ListView should be displayed as a grid of 3 (or whichever number user supplied on `Columns`) columns
  - For example, if there are 6 items `x1`, `x2`, `x3`, `x4`, `x5`, `x6`, the first row will contain `x1`, `x2`, `x3` and the second row will contain `x4`, `x5`, `x6`

stepinfwd:
closes #6891


kavinvenkatachalam (changes requested):
@stepinfwd Most of the things seems to be working fine. Two issues

-  When I try to change the columns value, I'm not able to see any change in the UI. 
-  Debugger throws an error whenever I try to change the column value. 

 Please take a look.

kavinvenkatachalam (dismissed):

stepinfwd:
opened here 
https://github.com/ToolJet/ToolJet/pull/6975

PR #6961, opened 2023-07-07
Feature added  to introduce Link column type in table

### LINKED ISSUE #6848: Link column type in table

The table does not support hyperlink type

Add a new type of column called `Link`.

The display of its content should be similar to that of `Link` component.

Properties:

- `key` : Should describe the key against which the url to which the link points is available
- `Label key`: Should describe the kay against which the label of the link is available
  - This should be empty by default, and if it is empty, the URL itself should be taken as the link
- `Target` : This should be a dropdown (with Fx) with two options:
  - Same window
  - New window

manishkushare:
Resolves #6848

### DIFF
--- a/frontend/src/Editor/Components/Table/Link.jsx
+++ b/frontend/src/Editor/Components/Table/Link.jsx
@@ -1,0 +1,17 @@
+import React from 'react';
+
+export const Link = ({ linkKey = 'https://dev.to/', linkLabel, linkTarget }) => {
+  return (
+    <div>
+      <a
+        href={linkKey}
+        target={linkTarget}
+        onClick={(e) => {
+          e.stopPropagation();
+        }}
+      >
+        {linkLabel ?? linkKey}
+      </a>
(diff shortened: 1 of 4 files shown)


kavinvenkatachalam (changes requested):
@manishkushare Few issues/suggestions,

1. When I try to click on the link without changing the link target (default - `New window`) but it still opens in the same page
2. I'm not able to link the tableData with the `Link` type. Like every column, the key should be picked from the table data.
3. When I try to filter it with the link label, it seems to be not working
4. Downloading the table data with link type seems to be not working with label entered

cc: @sherfin94

kavinvenkatachalam (approved):

sherfin94 (changes requested):
Hey @manishkushare , let us remove the `Href` property also, it is not needed.

cc: @kavinvenkatachalam

kavinvenkatachalam (approved):

PR #7007, opened 2023-07-12
Workspace crashing on plugins load

vjaris42:
Added null checks and loaded state for plugins api


arpitnath (approved):

frontend/src/Editor/DataSourceManager/DataSourceManager.jsx
```diff
@@ -490,7 +494,7 @@ class DataSourceManagerComponent extends React.Component {
       const filteredDatasources = this.state.filteredDatasources.map((datasource) => {
         const src = datasource?.iconFile?.data
           ? `data:image/svg+xml;base64,${datasource.iconFile?.data}`
-          : datasource.kind.toLowerCase();
+          : datasource?.kind?.toLowerCase();
```
akshaysasidrn, line 497: @vjaris42 Do we need all the optional chaining if we have a state `pluginsLoaded` to see if datasources are present?

vjaris42: As discussed with @arpitnath for now we use both as a precaution. Later when we revamp the datasourcemodal we wont have the chaining.

akshaysasidrn (approved):

PR #7101, opened 2023-07-24
[localization] translate strings in portuguese

andreppedroza:
add Portuguese (Brazilian Portuguese) localization (pt.json).


andreppedroza:
I found some corrections to make. Closing for now.

PR #7106, opened 2023-07-24
Added horizontal alignment  property for each columns in the table to set the horizontal alignment individually 

### LINKED ISSUE #6797: `Alignment` as a general property for all table columns

Table columns does not support configuring horizontal alignment right now.

Add a general property for all table column types, that would align the field's contents on
- Left
- Center
- Right

manishkushare:
Resolves #6797


kavinvenkatachalam:
@manishkushare Please resolve the conflicts

frontend/src/Editor/Components/Table/columns/index.jsx
```diff
@@ -94,7 +94,20 @@ export default function generateColumnsData({
       customRule: column?.customRule,
       sortType,
       columnVisibility: column?.columnVisibility ?? true,
-      Cell: function ({ cell, isEditable, newRowsChangeSet = null }) {
+      horizontalAlignment: column?.horizontalAlignment ?? 'left',
+      Cell: function ({ cell, isEditable, newRowsChangeSet = null, horizontalAlignment }) {
+        const determineJustifyContentValue = (value) => {
```
arpitnath, line 99: The default case does not return anything explicitly, which means it implicitly returns undefined. It might be better to add a return null statement to the default case to ensure that any unexpected values return a defined value.

arpitnath (changes requested):
Rest LGTM

arpitnath (approved):

PR #7142, opened 2023-07-26
Feature :: Exposing CSA for child widgets of form and listview

### LINKED ISSUE #5819: Unable to control Text component inside Form component
### Version Information
v2.2.2-ee2.0.5-cloud2.0.1

### Environment
ToolJet Cloud 

While trying to control(Set Text/ Set visibility) the `text` component inside a `form` component, it is expected to update accordingly. 

The component text control actions are not reflected inside the form component. 

1. Add a form component
2. Add a text component inside the form component
3. Add a button component inside the form component
4. Set an event to button, On click: Control component; set an action Set text/ Set visibility
5. Try to trigger the action

### Screenshots or Screencast
image

No logs were found.

### LINKED ISSUE #6893: All container components has to expose the CSA of their child components

Presently, few container components, does not expose the component specific actions of their child components.

The following components has to expose the CSA of their child components:

* ListView
  - Child components are already exposed as `children`, we just need to ensure CSA are also exposed
* Form ( #5819 is the same issue )
  - Child components should be exposed as `children` , in order to expose their CSA

stepinfwd:
closes #6893


kavinvenkatachalam:
@stepinfwd Please fix the lint errors and remove the commented code

kavinvenkatachalam (changes requested):
@stepinfwd I'm not see the components inside the form in the event handler. Please take a look

image

kavinvenkatachalam (changes requested):
@stepinfwd Not able to control the children of form and listView. Please take a look

kavinvenkatachalam:
> @stepinfwd Not able to control the children of form and listView. Please take a look

This can be ignored

frontend/src/_helpers/appUtils.js
```diff
@@ -81,9 +81,33 @@ export function onComponentOptionChanged(_ref, component, option_name, value) {
...
+    isForm = false;
+  Object.keys(_ref.state.appDefinition.pages[_ref.state.currentPageId]?.components).forEach((item) => {
+    if (component.parent == item) {
+      try {
+        isListView =
+          _ref.state.appDefinition.pages[_ref.state.currentPageId]?.components[item]?.component?.component ===
+          'Listview';
+        isForm =
+          _ref.state.appDefinition.pages[_ref.state.currentPageId]?.components[item]?.component?.component === 'Form';
+      } catch {
```
kavinvenkatachalam, line 94: This will affect the performance if we have a lot of component and might be solved with different approach. 
We can check if the parent is `listView` or `Form` by using the parent ID in `Box.jsx` and the function call can be avoided

kavinvenkatachalam (changes requested):
@stepinfwd Every time we set exposed variable, we're causing a render to the whole appEditor. If I have 5 exposed variable on `Button.jsx`, I'm re-rendering the Editor 5 times. Can we pass together and avoid the unwanted re-renders?

kavinvenkatachalam (approved):

PR #7143, opened 2023-07-26
Setup sentry logging as per version and upload its source map

### LINKED ISSUE #6950: Sentry logging for component failures

Presently there is no data reported when components fail

When a component fails, the error should be logged into sentry system.

It should contain these metadata like:

1. Date and time
2. Error message
3. Component
4. Component property values
5. ToolJet version
6. App created time

nakulnagargade:
Resolves #6950.


arpitnath (approved):

emidhun:
@kavinvenkatachalam  Tested the frontend crashes on sentry.
Looking good to me.
image

PR #7150, opened 2023-07-27
Feature/restapi ssl options

ganesh8056:
1. Added Support for SSL Option in REST API Plugin.
2. Updated ES-Lint version in Tool-jet Root modules.


ganesh8056:
Originally there were 4 files changed. Since `release/marketplace_1.3` is not upto date with `release/marketplace_1.2` it shows Files changed to 46

plugins/packages/restapi/lib/index.ts
```diff
@@ -155,6 +156,45 @@ export default class RestapiQueryService implements QueryService {
...
+            key: [sourceOptions.client_key],
+            certificate: [sourceOptions.client_cert],
+          },
+        };
+        break;
+      default:
+        break;
+    }
+
+    if (process.env.NODE_EXTRA_CA_CERTS) {
```
akshaysasidrn, line 182: could be refactored as:

```
if (process.env.NODE_EXTRA_CA_CERTS) {
  const certificateAuthority = [
    ...tls.rootCertificates,
    readFileSync(process.env.NODE_EXTRA_CA_CERTS),
  ].join('\n');

  httpsParams.https = httpsParams.https || {};
  httpsParams.https.certificateAuthority = certificateAuthority;
}
```

akshaysasidrn (approved):

plugins/packages/restapi/lib/index.ts
ganesh8056: REST-API plugin, When **SSL** is enabled, user will be providing **ca_cert**, along with that when NODE_EXTRA_CA_CERTS was given in env file, then in that case we need to concat NODE_EXTRA_CA_CERTS along with **ca_cert** provided by the user. So one more condition is necessary. We can refactor like
`const certificateAuthority = [
        ...tls.rootCertificates,
        readFileSync(process.env.NODE_EXTRA_CA_CERTS)
      ]`
      
      `httpsParams.https = httpsParams.https || {}
      if (httpsParams.https.certificateAuthority) {
        httpsParams.https.certificateAuthority = httpsParams.https.certificateAuthority.concat(certificateAuthority)
      } else {
        httpsParams.https.certificateAuthority = certificateAuthority;
      }`
      
      But anyhow one more condition is need there

akshaysasidrn: got it

PR #7190, opened 2023-08-03
removing copilot from ce

arpitnath:


kavinvenkatachalam (approved):

frontend/src/Editor/QueryManager/Components/Transformation.jsx
```diff
@@ -250,7 +250,7 @@ return [row for row in data if row['amount'] > 1000]
           </OverlayTrigger>
         </div>
 
-        <EducativeLebel />
+        {/* <EducativeLebel /> */}
```
gsmithun4, line 253: shall we remove the codes if not used?

gsmithun4 (changes requested):

gsmithun4 (approved):

Issue #7218, opened 2023-08-08
Fix failing postgres and snowflake cypress specs


Mekhla-Asopa:
Fix cypress failure spec for Postgres

PR #7219, opened 2023-08-08
Fix cypress failure for postgres and snowflake

Mekhla-Asopa:
Fixes: https://github.com/ToolJet/ToolJet/issues/7218

Cypress status: 
1. Postgres
Screenshot 2023-08-08 at 12 43 09 PM

2. Snowflake
Screenshot 2023-08-08 at 1 18 48 PM

### DIFF
--- a/cypress-tests/cypress/e2e/editor/data-source/snowflakeHappyPath.cy.js
+++ b/cypress-tests/cypress/e2e/editor/data-source/snowflakeHappyPath.cy.js
@@ -118,7 +118,7 @@ describe("Data sources", () => {
     );
   });
 
-  it.skip("Should verify the functionality of PostgreSQL connection form.", () => {
+  it.skip("Should verify the functionality of Snowflake connection form.", () => {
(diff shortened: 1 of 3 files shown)


cypress-tests/cypress/e2e/editor/data-source/snowflakeHappyPath.cy.js
```diff
@@ -118,7 +118,7 @@ describe("Data sources", () => {
     );
   });
 
-  it("Should verify the functionality of PostgreSQL connection form.", () => {
+  it.skip("Should verify the functionality of PostgreSQL connection form.", () => {
```
akshaysasidrn, line 121: Is it fine to skip? Also the spec text should be for snowflake right?

Mekhla-Asopa: Yes, Due to credentials are expired we are going to skip this.
I have updated test case name.

PR #7266, opened 2023-08-16
App definition re-designed 

arpitnath:
The current system relies on a JSON-based schema for storing and managing app definitions.
The transition aims to provide a structured architecture for efficiently storing, retrieving, and managing app versions, pages, components, layouts, event handlers, while introducing component and page-specific permissions. This enhancement will enable better data organization, fine-grained access control, and improved management of permissions at the individual component and page levels.


arpitnath:
ref: #7448

PR #7273, opened 2023-08-17
Enhancement :: Reducing number of rerenders due to exposed variables

stepinfwd:


stepinfwd:
opening [here]( https://github.com/ToolJet/ToolJet/pull/7473/)

PR #7281, opened 2023-08-17
Fix/query rename delay

johnsoncherian:
When query is selected for creation, it does not immediately enables the rename query input. Instead it wait for the create api to finish. This given sub optimal UX in low speed.

### DIFF
--- a/frontend/src/_stores/dataQueriesStore.js
+++ b/frontend/src/_stores/dataQueriesStore.js
@@ -8,6 +8,7 @@ import { useAppVersionStore } from '@/_stores/appVersionStore';
 import { runQueries } from '@/_helpers/appUtils';
 import { v4 as uuidv4 } from 'uuid';
 import { toast } from 'react-hot-toast';
+import { isEmpty, throttle } from 'lodash';
 
 const initialState = {
   dataQueries: [],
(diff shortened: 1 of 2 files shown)


frontend/src/_stores/dataQueriesStore.js
```diff
@@ -117,25 +121,41 @@ export const useDataQueriesStore = create(
...
             .then((data) => {
               set((state) => ({
-                isCreatingQueryInProcess: false,
+                creatingQueryInProcessId: null,
                 dataQueries: state.dataQueries.map((query) => {
                   if (query.id === tempId) {
-                    return { ...query, ...data, data_source_id: dataSourceId };
+                    return {
+                      ...query,
+                      // ...data,
```
nakulnagargade, line 134: Please remove console.

frontend/src/_stores/dataQueriesStore.js
```diff
@@ -144,6 +164,11 @@ export const useDataQueriesStore = create(
             .finally(() => useAppDataStore.getState().actions.setIsSaving(false));
         },
         renameQuery: (id, newName) => {
+          /** If query creation in progress, skips call and pushes the update to queue */
+          if (get().creatingQueryInProcessId && get().creatingQueryInProcessId === id) {
```
nakulnagargade, line 168: I guess we can remove get().creatingQueryInProcessId check and check for only equality like below

if (get().creatingQueryInProcessId === id) {
.....
}

arpitnath (approved):
Agree with it https://github.com/ToolJet/ToolJet/pull/7281/files#r1299572514
LGTM!

Issue #7279, opened 2023-08-17
TreeSelect Component: Text colour not adapting when switching the light and dark modes


ShazanRizvi:
### Environment
All environments

### Description
When you place the tree select component onto the application's canvas, the text color inside the tree select component fails to adjust according to the black and white color scheme when transitioning between light and dark modes.

### Steps to Reproduce:

1. Go to ToolJet application and click on the **Create new App** button to create an application.
2. The application builder window is opened.
3. From the Components sidemenu on the righ side search for **tree select** component.
4. Drag and drop the tree select component from the sidemenu onto the canvas.
5. On the left bottom most of the screen is the dark mode/light mode toggle icon.
6. Click on the icon and observe the colour of the text present inside the tree select component.

### Expected outcome:
When the light mode and dark mode is toggled the tree select component should change the color of the text to white and black depending on the mode of the application.

**This image depicts the colour of text in light mode:**

<img width="399" alt="Screenshot 2023-08-17 at 8 07 46 PM" src="https://github.com/ToolJet/ToolJet/assets/85864291/9ba168d3-7ef7-4bf2-8688-7329dd9c972c">

**This image depicts the colour of the text in dark mode:** 

<img width="368" alt="Screenshot 2023-08-14 at 11 42 31 AM" src="https://github.com/ToolJet/ToolJet/assets/85864291/9041d96b-d76d-4ba5-bc0f-0b22893f84f3">

mazkaaa:
Hi! I'd to work on this issue.

Uzair4kf:
I would love to work on this as my first contribution

renji18:
Hi @ShazanRizvi, I'd like to work on this issue, to start off my journey in open source

tejakummarikuntla:
Hello @Hiperultimate,

I sincerely appreciate your contribution, and I would like to kindly request you to review the [contributor guidelines](https://github.com/ToolJet/ToolJet/blob/develop/CONTRIBUTING.md) and ensure that you raise pull requests (PRs) for the issues assigned to you.

P.S: Assigned the issue to you, as you are a first-time contributor to the project with a solution.

yetea:
I'd like to work on this issue

1-ankush-1:
hey is this issue still open ?

i would like to contribute to this issue.

tejakummarikuntla:
Linking the PR: https://github.com/ToolJet/ToolJet/pull/7288

Yash088:
[...]

PR #7288, opened 2023-08-18
Fix: TreeSelect Component: Text color now adapts when switching to dark mode

Hiperultimate:
Resolves issue [#7279](https://github.com/ToolJet/ToolJet/issues/7279)
image
image


kavinvenkatachalam (changes requested):
@Hiperultimate This will break the support of text colour in the styles tab of the widget. Revert the change and add default colour `#000` [here](https://github.com/ToolJet/ToolJet/blob/develop/frontend/src/Editor/WidgetManager/widgetConfig.js#L5399)

tejakummarikuntla:
Hey @Hiperultimate, could you please address the reviews? This is a blocker for us to get this merged.

PR #7341, opened 2023-08-28
Pagination changes as per new design

nakulnagargade:


manishkushare:
@nakulnagargade ,

1. I cannot turn off pagination, that is enable Pagination prop, either from toggle switch or programmatically as well, It seems to be always true.
2. fx for serverSidePagination and cleintSidePagination is removed, it may break the older app, as user may have programmatically set the values of those two props.

Issue #7380, opened 2023-09-05
[refactor]: Performace improvements (reducing renders) 


nakulnagargade:
### What area of the codebase is to be refactored?

Following four things to be done as part of this issue

1. In editor component, use selectionInProgress and selectedComponents as a value read from zustand store instead of passing it as prop to avoid re-renders.
2. On Box component, we are calculating properties and styles for every component present in the canvas even when user don't interact with canvas. We want to prevent this behaviour.
3. On every mouse click and release of it , we are setting one state called setSelectedComponents to empty array but then it is set as new instance hence resulting in change of reference (causing re-render) even though the value is same.
4. Club selected components and show widget manager together.

### What are the potential advantages of this proposed refactoring?

Solving above 4 problems will help in reducing unwanted rerenders and improve the performance of the app overall.

PR #7412, opened 2023-09-11
[Refactor]7380 - Improve performance by preventing states used as prop when not necessary

nakulnagargade:
Reolves #7380.


frontend/src/Editor/Header/index.js
```diff
@@ -44,6 +44,10 @@ export default function EditorHeader({
     shallow
   );
 
+  const { pageHandle } = useCurrentStateStore((state) => ({
+    pageHandle: state?.page?.handle,
+  }));
```
kavinvenkatachalam, line 49: We can add a `shallow` check here

frontend/src/Editor/Editor.jsx
```diff
@@ -1752,41 +1669,36 @@ class EditorComponent extends React.Component {
...
+                        </center>
+                      )}
+                    </div>
+                  }
+                  widgetManagerTab={
+                    <WidgetManager
+                      componentTypes={componentTypes}
+                      zoomLevel={zoomLevel}
+                      darkMode={this.props.darkMode}
+                    />
```
kavinvenkatachalam, line 1699: This is okay for now. Once app def change is merged, we can separate all the logic out of the components

frontend/src/Editor/Container.jsx
```diff
@@ -313,7 +310,14 @@ export const Container = ({
       setBoxes(newBoxes);
       updateCanvasHeight(newBoxes);
     },
-    [isVersionReleased, enableReleasedVersionPopupState, boxes, setBoxes, selectedComponents, updateCanvasHeight]
+    [
+      isVersionReleased,
+      enableReleasedVersionPopupState,
+      boxes,
+      setBoxes,
+      useEditorStore.getState().selectedComponents,
```
kavinvenkatachalam, line 318: We're not re-rendering the components when the value is changed. So, this will not work what I feel. We can remove `useEditorStore.getState().selectedComponents` from the dependency as we're fetching the latest data inside the function. Check whether it is working after removing it and if so, we can proceed with that

frontend/src/Editor/Container.jsx
```diff
@@ -563,7 +567,7 @@ export const Container = ({
     removeComponent,
     currentLayout,
     deviceWindowWidth,
-    selectedComponents,
+    useEditorStore.getState().selectedComponents,
```
kavinvenkatachalam, line 570: Here as well

frontend/src/Editor/Container.jsx
```diff
@@ -541,7 +545,7 @@ export const Container = ({
       removeComponent,
       currentLayout,
       deviceWindowWidth,
-      selectedComponents,
+      selectedComponents: useEditorStore.getState().selectedComponents,
```
kavinvenkatachalam, line 548: Check whether it is being used in any of the components. If not we can remove this from `containerProps`

kavinvenkatachalam (changes requested):
@nakulnagargade Requested few changes. Please take a look

frontend/src/Editor/Header/index.js
nakulnagargade: Done

frontend/src/Editor/Container.jsx
nakulnagargade: Yes, not needed,  removed it.

frontend/src/Editor/Container.jsx
nakulnagargade: selectedComponents not being used from containerProps, hence removed it from containerProps.

kavinvenkatachalam (approved):

frontend/src/Editor/SubContainer.jsx
```diff
@@ -70,6 +70,10 @@ export const SubContainer = ({
     shallow
   );
 
+  const { selectedComponents } = useEditorStore((state) => ({
+    selectedComponents: state.selectedComponents,
+  }));
+
```
kavinvenkatachalam, line 76: Do we need this? It's used on `onDragStop` function and we can use the `getState` here

kavinvenkatachalam (changes requested):
@nakulnagargade Added a comment

frontend/src/Editor/SubContainer.jsx
nakulnagargade: Resolved. Used getState here.

kavinvenkatachalam (approved):

arpitnath (changes requested):
Does this file serve any purpose? If not, we can safely remove it.

arpitnath (approved):

PR #7447, opened 2023-09-15
[docs] revoved comments from schema and added tables and example

withshubh:
#### changes
- updated form doc with an example form that uses the custom schema of components
- removed comments from custom schema code blocks
- Added separate properties table for each component custom schema


karan-rathod-316 (approved):
LGTM!

PR #7448, opened 2023-09-15
Release : Appbuilder - appdefinition architecture revamp

arpitnath:
The current system relies on a JSON-based schema for storing and managing app definitions.
The transition aims to provide a structured architecture for efficiently storing, retrieving, and managing app versions, pages, components, layouts, event handlers, while introducing component and page-specific permissions. This enhancement will enable better data organization, fine-grained access control, and improved management of permissions at the individual component and page levels.

- Migrations for the existing JSON-based app definition schema to a structured table-based architecture. Also introduces component and page-specific permissions, improves data organization, and enables fine-grained access control. Additionally, it adds the 'globalSettings' column to the 'appVersion' table.


gsmithun4:
@arpitnath Suggestion: shall we move the global settings to different table? since we are querying versions for other operations like listing version, we don't need these settings. separating these data will improve certain APIs

server/src/services/app_import_export.service.ts
```diff
@@ -145,6 +179,7 @@ export class AppImportExportService {
         multiPages: true,
         multiEnv: true,
         globalDataSources: true,
+        normalizedAppDefinitionSchema: true,
```
akshaysasidrn, line 182: what is this key used for?

server/src/services/app_import_export.service.ts
```diff
@@ -166,23 +201,22 @@ export class AppImportExportService {
...
-      await this.setupImportedAppAssociations(
-        manager,
-        importedApp,
-        schemaUnifiedAppParams,
-        user,
-        externalResourceMappings
-      );
-      await this.createAdminGroupPermissions(manager, importedApp);
-    });
+    const isNormalizedAppDefinitionSchema = appParams?.schemaDetails?.normalizedAppDefinitionSchema;
```
akshaysasidrn, line 208: Can we make this decision on the basis of releasing tooljet version? Idea being when we want to decouple import flows based on version it will be easier

kavinvenkatachalam (approved):

PR #7451, opened 2023-09-18
Fixed undo app name issue

### LINKED ISSUE #7192: App name should not be part of undo-redo history

Right now, undo/redo will undo or redo any changes to the app's name

Undo and redo should not affect app name at all

S-Abdul-Rahman:
Fixed #7192


johnsoncherian (approved):

frontend/src/Editor/Header/EditAppName.jsx
```diff
@@ -31,6 +31,19 @@ function EditAppName({ appId, appName = '', onNameChanged }) {
       });
   };
 
+  React.useEffect(() => {
+    const handleKeyDown = (e) => {
```
Ckbhatia, line 35: You can move this function outside of this component

kavinvenkatachalam (approved):

kavinvenkatachalam (changes requested):
@S-Abdul-Rahman We can call the function inside the useEffect and we need to remove the event listener when the component is unmounted

kavinvenkatachalam:
@arpitnath Will it be fixed in your PR?

S-Abdul-Rahman:
> @S-Abdul-Rahman We can call the function inside the useEffect and we need to remove the event listener when the component is unmounted

@kavinvenkatachalam  So, are you suggesting that these changes need to be reverted back to the previous ones? In the previous code, I added this function inside useEffect and I removed the event listener when the component unmounted.

kavinvenkatachalam (approved):

arpitnath:
https://github.com/ToolJet/ToolJet/pull/7451#issuecomment-1735297159
@kavinvenkatachalam Yes

PR #7464, opened 2023-09-20
Fix query panel not returning falsey values on clicking preview

nakulnagargade:
on clicking preview, if the output is falsey, we were not returning it. This PR fixes it. (we now return even the falsey values).


frontend/src/Editor/QueryManager/Components/Preview.jsx
```diff
@@ -39,6 +39,9 @@ const Preview = ({ darkMode }) => {
   }, [queryPreviewData]);
 
   const renderRawData = () => {
+    if (queryPreviewData === null || queryPreviewData === undefined || queryPreviewData === false) {
```
johnsoncherian, line 42:
```suggestion
    if (!queryPreviewData) {
      return `${queryPreviewData}`;
    }

    return isJson ? JSON.stringify(queryPreviewData).toString() : queryPreviewData.toString();
```

johnsoncherian: Can you try this. I think this should handle all.

kavinvenkatachalam (approved):

johnsoncherian (approved):

johnsoncherian (dismissed):

kavinvenkatachalam (approved):

PR #7473, opened 2023-09-21
Feat ::  Reducing no of rerenders due to exposed variables

### LINKED ISSUE #7493: Reducing number of rerenders caused due to setting exposed variables
Optimize the app for improved performance by reducing unnecessary re-renders.

 Every time we set exposed variable, we're causing a render to the whole appEditor. If I have 5 exposed variable on Button.jsx,we are re-rendering the Editor 5 times.
 
yes

stepinfwd:
closes #7493


frontend/src/Editor/Box.jsx
```diff
@@ -316,8 +316,12 @@ export const Box = memo(
...
                 exposedVariables={exposedVariables}
                 styles={{ ...validatedStyles, boxShadow: validatedGeneralStyles?.boxShadow }}
-                setExposedVariable={(variable, value) => onComponentOptionChanged(component, variable, value, id)}
-                setExposedVariables={(variableSet) => onComponentOptionsChanged(component, Object.entries(variableSet))}
+                setExposedVariable={(variable, value) => {
+                  onComponentOptionChanged(component, variable, value, id);
+                }}
+                setExposedVariables={(variableSet) =>
+                  onComponentOptionsChanged(component, Object.entries(variableSet), id)
+                }
```
kavinvenkatachalam, line 324: We can revert this

kavinvenkatachalam (changes requested):
@stepinfwd Placing the table inside the sub container is breaking. You can add a new tab widget and switch to the last tab. 

image

kavinvenkatachalam (approved):

PR #7488, opened 2023-09-25
Update selector design as per new design specs

nakulnagargade:
This PR update widget selector design as per new design specs.


nakulnagargade:
Closing this PR as this is as per older design and newer design is already been implemented with grid changes.

Issue #7326, opened 2023-08-24
Component Sidebar: Position of the Forms component should be the first item in the Forms section


ShazanRizvi:
### Environment
All Environments

### Description:
In the sidebar's list of components, the forms component is currently located within the **other** section. However, it should be positioned at the top of the **Forms** section instead of being placed at the bottom of the other section.

### Steps to Reproduce:

1. Go to the ToolJet application and click on the Create New App button to create an application.
2. The application builder window is opened.
3. There exists a sidebar on the right side of the app builder which displays a list of all the components which can be dragged and dropped on the canvas.
4. There are some sections dedicated to types of components such as Layouts, Forms, etc.
5. Scroll through this sidebar find the Form component and look for the section it resides in.

### Expected Output:
The forms component should be placed as the first option in the **Forms** section.

Current position of forms component:
<img width="194" alt="Screenshot 2023-08-24 at 12 32 44 PM" src="https://github.com/ToolJet/ToolJet/assets/85864291/49196133-5119-4321-8611-31a0a3c404f2">
Should be on top of this section:
<img width="192" alt="Screenshot 2023-08-24 at 12 33 17 PM" src="https://github.com/ToolJet/ToolJet/assets/85864291/f337ba1b-c2c1-4826-8f81-bcbb2f35924f">

SWASTI-SUNDAR:
hey can you assign this issue to me i want to contribute in it

Prathamesh2805:
If nobody is working on it can someone, please assign it to me, I would love to work on it!

tejakummarikuntla:
Hey @SWASTI-SUNDAR , thanks for your interest; assigned the issue to you.

Brag2gr8:
Hello, First timer
Assign this to me

tejakummarikuntla:
Unassigning @SWASTI-SUNDAR due to inactivity.

JunedKhan101:
@tejakummarikuntla I'd like to pick this up. Can you assign this to me?

tejakummarikuntla:
We want to express our sincere appreciation for your interest and involvement in this issue. Regrettably, we have decided to address this internally for specific reasons. We apologize for any inconvenience and greatly appreciate your understanding.

PR #7490, opened 2023-09-25
Fixed form component position

S-Abdul-Rahman:
Fixed Form component position #7326


kavinvenkatachalam (changes requested):
@S-Abdul-Rahman The change in `frontend/src/Editor/WidgetManager/widgetConfig.js` is not needed. You can revert it

kavinvenkatachalam:
@S-Abdul-Rahman We can keep the changes of `widgetConfig.js`

kavinvenkatachalam (approved):

PR #7633, opened 2023-10-05
Fixed query builder filter for ToolJet db

### LINKED ISSUE #6236: In the query builder, the filter does not work for empty strings
The data should be updated based on a filter condition if an empty string is provided

It updates the entire column's data

1. Login with the valid credentials 
2. Create a table with some valid data and make an empty string for any row
3. Create an application and add a tooljet query
4. Add a created table in the query
5. Select the operation as an update row
6. Add a filter condition for an empty row

### Screenshots or Screencast
Screenshot 2023-05-03 at 4 13 50 PM
Screenshot 2023-05-03 at 4 31 30 PM

S-Abdul-Rahman:
Fixed #6236


frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/util.js
```diff
@@ -28,7 +28,7 @@ export const hasEqualWithNull = (queryOptions, operation) => {
     const filterKeys = Object.keys(filters);
     for (let i = 0; i < filterKeys.length; i++) {
       const filter = filters[filterKeys[i]];
-      if (filter.operator !== 'is' && filter.value === null) {
+      if (filter.operator !== 'is' && filter.value === '') {
```
akshaysasidrn, line 31: Think this should be `(filter.operator !== 'is' && (filter.value === '' || filter.value === null) )`

User can still pass null with `'{{null}}'` at the query editory

S-Abdul-Rahman: > Think this should be `(filter.operator !== 'is' && (filter.value === '' || filter.value === null) )`
> 
> User can still pass null with `'{{null}}'` at the query editor. Can rename the function to something like`hasEmptyStringOrNullValue`

Yeah Sure, will make changes like this as well as rename the function.

akshaysasidrn (dismissed):

frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/util.js
```diff
@@ -20,15 +20,15 @@ import { get } from 'lodash';
...
  */
-export const hasEqualWithNull = (queryOptions, operation) => {
+export const hasEmptyStringOrNullValue = (queryOptions, operation) => {
   const filters = get(queryOptions, `${operation}.where_filters`);
   if (filters) {
     const filterKeys = Object.keys(filters);
     for (let i = 0; i < filterKeys.length; i++) {
       const filter = filters[filterKeys[i]];
-      if (filter.operator !== 'is' && filter.value === null) {
+      if (filter.operator !== 'is' && (filter.value === '' || filter.value === null)) {
```
ganesh8056, line 31: Empty string can be input to the column. The fix doesn't allow to match a empty string in the column. For `is operator` the input can be empty either null or not null. @akshaysasidrn Could you correct me if the explanation is wrong.

akshaysasidrn: @ganesh8056 yeah you are right here. We need to validate for empty filters before executing the query.

akshaysasidrn (approved):

Issue #7711, opened 2023-10-09
Pressing enter/return key should delete the component


withshubh:
### What is the expected behaviour?

When the confirmation modal for deleting the component pops up, the user should be able to provide input `Yes` by pressing the `Enter/return` key or `Cancel` using the `Esc` key. 

### What is the current behaviour?

Currently, the user will have to use the mouse click to select the option.

### How to reproduce the issue?

1. Drag a component onto the canvas
2. Use `delete` button (windows/linux) or `back` (mac) to delete the component
3. The confirmation modal will popup
4. Try pressing `Enter` or `return` 

### Screenshots or Screencast
![Screenshot 2023-10-09 at 4 57 35 PM](https://github.com/ToolJet/ToolJet/assets/25361949/df52b9a5-aab4-4e0e-b540-a5129b6fd059)

NotTheRightGuy:
I can help with this. Can you assign this issue to me?

joanjeremiah:
I would like to work on this issue, please assign this to me
Thanks

ShrutiAngane:
Hello @withshubh the issue looks like something that I would love to work on. Can you please assign this to me ?

NithinSravan:
Hi @withshubh I would love take up this task if it's still available. Can you please assign it to me?

tejakummarikuntla:
Hey everyone, thanks for your interest. Assigned the issue to @HeetVekariya.
Feel free to join our community on [Slack](https://tooljet.com/slack). If you need any help related to issues/contributions, please post in the #contributors channel.

NotTheRightGuy:
Thanks, @tejakummarikuntla I will start working on it.

PR #7726, opened 2023-10-10
Fix/widget deletion confirmation keys

Yash-rai-29:
PR Summary
This pull request addresses issue #7711 , where the confirmation modal for deleting a component should allow users to confirm deletion using the Enter/return key and cancel using the Esc key.

Changes
Added event listeners for the Enter/return and Esc keys to trigger deletion confirmation and cancellation, respectively.
Updated the useEffect hook to handle these key events and manage the confirmation accordingly.


tejakummarikuntla:
Hello @Yash-rai-29, please refrain from raising Pull Requests for issues that have not been assigned to you. Thank you.

PR #7765, opened 2023-10-10
Add new templates and update an existing template

abhinabaadhikari:
Added the following templates:
- Sales analytics dashboard
- Lead management system
- Inventory management

Updates description of the following template:
- Supply chain management


abhinabaadhikari:
Added two more templates:
- Customer support admin
- Customer ticketing form

akshaysasidrn (approved):

PR #7812, opened 2023-10-11
Fixes CurrentUser & Mode not present in globals in inspector

nakulnagargade:
Fixes :

1. CurrentUser & Mode not present in globals in inspector (Already tested).
2. Creating page not changing the slug

### DIFF
--- a/frontend/src/Editor/EditorFunc.jsx
+++ b/frontend/src/Editor/EditorFunc.jsx
@@ -141,6 +141,7 @@ const EditorComponent = (props) => {
     undo: [],
     redo: [],
   });
+  const [newPageId, setNewPageId] = useState('');
 
   // refs
   const canvasContainerRef = useRef(null);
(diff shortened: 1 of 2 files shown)


arpitnath:
@nakulnagargade 
revert the commit for page switch fix on creating a new change.
I dont think so we should introduce one more state and an useEffect to update the app slug.
Have fixed it here : https://github.com/ToolJet/ToolJet/pull/7873
cc: @emidhun

frontend/src/Editor/EditorFunc.jsx
```diff
@@ -1265,8 +1260,17 @@ const EditorComponent = (props) => {
       switchPage: true,
       pageId: newPageId,
     });
+    setNewPageId(newPageId);
   };
 
+  // This UseEffect is used when new page is created
+  useEffect(() => {
+    if (newPageId) {
```
arpitnath, line 1268: Lets not introduce a useEffect if we can avoid it

arpitnath (changes requested):

nakulnagargade:
> @nakulnagargade Could you please revert the commit related to the page switch fix when creating a new change? I believe there's no need to introduce an additional state and a useEffect to update the app slug. The issue has been resolved here : #7873 cc: @emidhun

Commit is reverted.

PR #7831, opened 2023-10-12
Fixed current version app export issue

### LINKED ISSUE #7708: Export tooljet tables that are only used within version
When exporting specific versions, we should only export tooljet tables that are only used within version that is exported.
ex:v1 has two queries with table_a  and table_bv2 has queries only with table_b
When we export only v2 with table schema, it should only export table_b .
Currently we export all the tables across the versions even if only a specific version is exported.

S-Abdul-Rahman:
Fixed #7708


frontend/src/HomePage/ExportAppModal.jsx
```diff
@@ -81,13 +109,12 @@ export default function ExportAppModal({ title, show, closeModal, customClassNam
         closeModal();
       });
   };
-
+  var isDownload = ''
```
akshaysasidrn, line 112: Think you can skip this declaration

frontend/src/HomePage/ExportAppModal.jsx
```diff
@@ -164,14 +191,14 @@ export default function ExportAppModal({ title, show, closeModal, customClassNam
               className="import-export-footer-btns"
               variant="tertiary"
               data-cy="export-all-button"
-              onClick={() => exportApp(app, null, exportTjDb, tables)}
+              onClick={() => exportApp(app, null, exportTjDb, tables, isDownload = 'true')}
```
akshaysasidrn, line 194: Can't we simply pass `tables` or `allTables` here itself instead of a new parameter?

akshaysasidrn (changes requested):

frontend/src/HomePage/ExportAppModal.jsx
S-Abdul-Rahman: @akshaysasidrn yes, this makes sense, will change that.

frontend/src/HomePage/ExportAppModal.jsx
```diff
@@ -40,9 +66,10 @@ export default function ExportAppModal({ title, show, closeModal, customClassNam
     fetchAppVersions();
     fetchAppTables();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [versionId]);
 
-  const exportApp = (app, versionId, exportTjDb, tables) => {
+  const exportApp = (app, versionId, exportTjDb, exportTtables) => {
+    // const finalTable = isDownload === 'true' ? allTables : tables
```
akshaysasidrn, line 72: please remove comment

frontend/src/HomePage/ExportAppModal.jsx
```diff
@@ -40,9 +66,10 @@ export default function ExportAppModal({ title, show, closeModal, customClassNam
     fetchAppVersions();
     fetchAppTables();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [versionId]);
 
-  const exportApp = (app, versionId, exportTjDb, tables) => {
+  const exportApp = (app, versionId, exportTjDb, exportTtables) => {
```
akshaysasidrn, line 71: `exportTables`*

akshaysasidrn (approved):

PR #7858, opened 2023-10-13
fixes: fixes on on app load switch page action via run queires

arpitnath:


emidhun:
Not working on preview.
Please have a look.

Issue #7802, opened 2023-10-11
Fix the `Edit email address` button after signup


withshubh:
### What is the expected behaviour?

The width and height of the `Edit email address` button should be the same as the `Resent verification email` button above it. 

### What is the current behaviour?

Currently, the height and width are not the same as the button above it.

### How to reproduce the issue?

1. Enter the details on the [sign-up page](https://www.tooljet.com/signup). Note: You should sign up for the first time using the email.
2.  Click on `Get started for free` button
3. On the next page, observe the `Edit email address` button.

### Screenshots or Screencast

![ToolJet-Dashboard](https://github.com/ToolJet/ToolJet/assets/25361949/935df27d-d86e-4b5d-8b8c-634772307a43)

ronisarkarexe:
@withshubh can you please assign me this issue? I want to work on this issue.

Virtual4087:
you can assign me this issue

Muhammad-Shahid-Jamal:
Hi i would like to work on this issue plz assign it to me

aryankshl:
hey @withshubh can you please assign this to me?

Devvrat1010:
hello @withshubh can you please assign this task to me ?

GeeekySouro003:
Hello sir i will be very grateful if i get the oppurtunity to work on this.Please assign me on this.

aastha44ma:
Hello Sir, Can you please assign this issue to me ? I will give my best to provide quality work.

AryanSaxenaa:
Hi @withshubh ,Please assign this issue to me

akshayw1:
Hello @withshubh, Please assign me the issue, My way to deal with the task, will be, Initially I will go through contributing guide, and will understand the require stuff, Remaining the task, Need to Just apply the Styled CSS as of above button, Which can be easily done by me. Looking forward to work on these issue.

Rahulkottak:
![Screenshot 2023-10-12 150923](https://github.com/ToolJet/ToolJet/assets/114240026/684d9f9b-a642-4c6d-9587-a850b4d94711)

I'm getting this type of interface.. can u plz confirm the issue once again. @withshubh

MrLEGENDx10:
assign me

sohalkumar:
@withshubh The issue only occurred when we tried to access a protected route (I tried accessing `/dashboard`) without signing in, which redirects to the login/signup page.
I have created a PR that should solve this issue.

[...]

PR #7875, opened 2023-10-14
Fixed the Edit email address button

Tomkndn:
fixes: #7802 
screenshot for the fixes:
image


frontend/src/_styles/theme.scss
```diff
@@ -10610,7 +10610,6 @@ tbody {
 .tj-btn-tertiary {
   padding: 10px 20px;
   gap: 8px;
-  width: 112px;
```
gsmithun4, line 10613: .tj-btn-tertiary is used in some other places, we can't set width for it

gsmithun4 (changes requested):

PR #7912, opened 2023-10-16
Update tooltips for QR scanner, rating, divider, and file picker comp…

ShivamJha2436:
…onents


ShivamJha2436:
I solved this issue

PR #7939, opened 2023-10-17
Revamps related to app import and app clone resources

AnantshreeChandola:
Revamps related to app import and app clone resources
#7940

### DIFF
--- a/frontend/src/HomePage/HomePage.jsx
+++ b/frontend/src/HomePage/HomePage.jsx
@@ -133,14 +133,14 @@ class HomePageComponent extends React.Component {
     _self.setState({ creatingApp: true });
     try {
       const data = await appService.createApp({ icon: sample(iconList), name: appName });
-
       const workspaceId = getWorkspaceId();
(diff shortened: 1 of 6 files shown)


frontend/src/HomePage/HomePage.jsx
```diff
@@ -171,13 +171,16 @@ class HomePageComponent extends React.Component {
     this.setState({ showAppDeletionConfirmation: true, appToBeDeleted: app });
   };
 
-  cloneApp = async (appId, appName) => {
+  cloneApp = async (appName, appId) => {
```
gsmithun4, line 174: AppCard.jsx line 139, this function is called with 
`cloneApp(app)` app object as only attribute

gsmithun4 (changes requested):

frontend/src/HomePage/HomePage.jsx
AnantshreeChandola: Removed that cloneApp func call, it was not needed.

gsmithun4 (approved):

PR #8020, opened 2023-10-25
Make use of paramterized queries on join

akshaysasidrn:


server/src/services/tooljet_db.service.ts
```diff
@@ -1,8 +1,9 @@
 import { BadRequestException, HttpException, Injectable, NotFoundException, Optional } from '@nestjs/common';
-import { EntityManager, In, QueryFailedError } from 'typeorm';
+import { EntityManager, In, ObjectLiteral, QueryFailedError, SelectQueryBuilder, TypeORMError } from 'typeorm';
 import { InjectEntityManager } from '@nestjs/typeorm';
 import { InternalTable } from 'src/entities/internal_table.entity';
 import { isString, isEmpty } from 'lodash';
+import { camelCase } from 'lodash';
```
ganesh8056, line 6: could we add **camelCase** import to line 5. While auto import it has added import to new line.

ganesh8056 (changes requested):

Issue #7792, opened 2023-10-11
Change labels to sentence case for PDF component


withshubh:
### Problem

There is an inconsistency in the text casing, some labels under `Properties` and `Styles` tab in the component manager are not in [Sentence case](https://www.k-state.edu/grad/academics/etdr/write/sentence-case.html).

### What is the expected behaviour?

All the labels under the `Properties` and `Styles` tab in the component managers should be in the Sentence case.

### What is the current behaviour?

The below labels are not in Sentence case(changes are only needed in the below labels).

| Current Values       | Expected Values       |
| :------------------- | :-------------------- |
| Show the Download        | Show the download         |

### How to reproduce the issue?

1. Open ToolJet.
2. Create a new app.
3. Search for the component and drag it onto the Canvas.
4. Click on the component to open the Component Manager in the right sidebar.
5. Examine the labels under both the `Properties` and `Styles` tabs. Some labels are not in sentence case and should be updated accordingly.

### Additional Information

1. Changes are only required in the `widgetConfig.js` file, locate the object which has the related component name in the name key, and make changes only in that particular object.  
**For e.g. for table component, you will make changes in the below object of the widgetConfig.js file:**
```
{
    name: 'Table',
    displayName: 'Table',
    description: 'Display paginated tabular data',
    component: 'Table',
.....
}
```
2. Do not make changes to any other labels, files or components.

### Screenshots or Screencast

**Example** using the table component, items that are not in Sentence case are highlighted in red, these are inconsistent with other labels which are in Sentence case

![image](https://github.com/ToolJet/ToolJet/assets/25361949/a7eca8e0-54b0-4400-a680-12d8d0ff9893)

imsoumik03:
hey @withshubh can you please assign me this issue?

adarshmalviya12:
I can work on this

MrLEGENDx10:
assign me

tejakummarikuntla:
Hey everyone, thank you for your interest. Assigned the issue to @imsoumik03.
Feel free to join our community on [Slack](https://tooljet.com/slack). If you need any help related to issues/contributions, please post in the #contributors channel.

Damon248:
Hello @tejakummarikuntla! Can I work on this task or any other similar issue like this?

SujalGupta244:
hey @tejakummarikuntla would you please assign me this issue

karan-rathod-316:
Hi @Damon248, assigning this to you.

[...]

PR #8062, opened 2023-10-30
Changed the values in widgetConfig file into sentance case. 

Damon248:
Solved issue number #7792 and #7800


karan-rathod-316 (approved):
LGTM

PR #8083, opened 2023-10-31
removed feature images below how it works in readme

withshubh:


Navaneeth-pk:
I think we can keep the screenshot but remove other images below "how it works" diagram.

withshubh:
Made the change

Screenshot 2023-11-01 at 4 39 12 PM

gsmithun4 (dismissed):

PR #8097, opened 2023-11-01
feat: Text component revamp

### LINKED ISSUE #7661: Text component: Enhancement
### Required enhancements:

[Design file](https://www.figma.com/file/mnyECAMd6Wc7qRIDyYmYfI/Components-redesign-01?type=design&node-id=1502-665929&mode=design&t=iaeG56fwLCqUdNZE-0)
(Ignore elements in the design that are not described in the following checklist)

- [ ] Add new property that determines how the text is interpreted. This property can have three possible values: `plain text`, `markdown` and `HTML`. Check the design before implementation
- [ ] Next section should be events
- [ ] Next section should be titled `Additional options`, which contains
  - [ ] Loading state
  - [ ] Visibility
  - [ ] Disabled
  - [ ] Tooltip
- [ ] In the styles section, we need a new style called `Vertical alignment`, this should determine how the text content is vertically aligned within the component's area. Check the design.
- [ ] There is a new section called 'Container options' in the style tab, which has the following styles:
  - [ ] Background
  - [ ] Border color
  - [ ] Border radius (New style)
  - [ ] Box shadow
  - [ ] Padding (New style, check the design)
- [ ]  The order in which styles should be populated is the same as in the design
- [ ] New CSA called `clear` which sets the text of the component to empty string
- [ ] Three new CSAs:
  - [ ] `setVisibility`: same as the existing CSA called `visibility`, lets not display `visibility` in the inspector anymore, even though it should work
  - [ ] `setLoadingState`: allows setting loading state to either `true` or `false`
  - [ ] `setDisabled`: Allow disabling or enabling the component
- [ ] Three new exposed variables:
  - [ ] `isLoading`: returns the loading state
  - [ ] `isDisabled` : returns the disabled state
  - [ ] `isVisible` : returns the visibility state

nakulnagargade:
Closes #7661


sherfin94 (changes requested):
Hey @nakulnagargade , nice work, but few items needs to be looked into:

1. When text is dropped in dark mode, it looks like this:
image

2. The text component is horizontally resizeable from viewer mode, this should not be allowed
image

3. Lets call this `Background color` instead: 
image

4. Tooltip is `undefined` by default, lets make it an empty string, like other components, so that its not red when we drop: 
image

5. Scroll seems to be working regardless of whether we have disabled it or not

nakulnagargade:
@sherfin94 All the points are taken care of. Regarding background color, we have renamed it to "Background" as per the discussion with Jaseem

> 3. Lets call this `Background color` instead:
> 
> image

sherfin94 (dismissed):
Hi @nakulnagargade , the tooltip does not seem to be working, please look into that as well.

nakulnagargade:
> Hi @nakulnagargade , the tooltip does not seem to be working, please look into that as well.

@sherfin94 This is resolved.

PR #8145, opened 2023-11-14
Fix ::App breaking when using string functions in codehinter

### LINKED ISSUE #8146: App breaks when using string functions in codehinter
App shouldn't break in any case

App breaks when typing something like `{{ "hello".toUpperCase()}}`

1.Try typing `{{ "hello".toUpperCase()}}`
2.App breaks when we reach`{{ "hello".toUpperCase}}`

Yes

stepinfwd:
close #8146


frontend/src/Editor/Components/Text.jsx
```diff
@@ -32,12 +32,12 @@ export const Text = function Text({ height, properties, styles, darkMode, setExp
   // eslint-disable-next-line react-hooks/exhaustive-deps
   useEffect(() => {
     const text = computeText();
-    setText(text);
-    setExposedVariable('text', text);
+    setText(typeof text == 'string' ? text : '');
+    setExposedVariable('text', typeof text == 'string' ? text : '');
```
kavinvenkatachalam, line 36: This would break the existing components if it is a number

kavinvenkatachalam (changes requested):
@stepinfwd Here are the few findings,

- Instead of success preview, can we show it as an error
image

- Instead of checking if it is text in each component, can we check whether it is a function and set it as an empty value? If possible we can check it using validate function

PR #8153, opened 2023-11-14
Bugfix :: Resolving Issue with comments not displaying during walkthrough

### LINKED ISSUE #8154: Issue with Comments Not Displaying - During Walkthrough
comments was not appearing on comments option in step 6 during walktrough.

In the app builder tour guide - (step 5) would need to be updated.  Break it into 2 steps instead -
1)preview + share
2)promote - and also talk about release in this.
Also the next step (step 6) related to comments was not appearing on comments option.

1.Go to walthrough creating a new app in incognito or clearing cache

Yes

stepinfwd:
closes #8154


kavinvenkatachalam (changes requested):
@stepinfwd After first step, it is going directly to the third instead of second. Can you check it out? Seems like second is datasource which is removed in the editor. If so, we can remove that step

kavinvenkatachalam (approved):

PR #8163, opened 2023-11-17
fix: Only one of the events trigger when multiple events with debounce is created

### LINKED ISSUE #8164: When multiple events with debounce is created only one of them trigger
All the events shoukd trigger with proper debounce

When every event has debounce only of them trigger

1.Create multiple event with debounce values
2.Observe

### Screenshots or Screencast

https://github.com/ToolJet/ToolJet/assets/31006028/4e9d35e5-2733-40e9-89e3-4b6e5e767f87

Yes

stepinfwd:
closes #8164 8164


frontend/src/_helpers/appUtils.js
```diff
@@ -384,16 +384,24 @@ function logoutAction() {
 }
 
 function debounce(func) {
-  let timer;
+  const timers = new Map();
+
   return (...args) => {
     const event = args[1] || {};
+    const eventId = JSON.stringify(event) + '_' + Date.now(); // Append timestamp
```
kavinvenkatachalam, line 391: can we use `uuid()` instead of stringify and combine with date?

kavinvenkatachalam (changes requested):
@stepinfwd Requested a small change. Rest looks good to me

kavinvenkatachalam (approved):

PR #8169, opened 2023-11-17
Fix for handling 'go to app' event for previously added unreleased apps and new apps

AnantshreeChandola:
1. We will only show released apps on the drop down for 'Go to App' action event on app editors.
2. For previously added unreleased apps, we are giving an option to editors to open the app from the error modal and release it.
Screenshot 2023-11-17 at 9 33 18 PM
#8254

### DIFF
--- a/frontend/src/_helpers/handleAppAccess.js
+++ b/frontend/src/_helpers/handleAppAccess.js
@@ -28,14 +28,9 @@ export const handleAppAccess = async (componentType, slug, version_id) => {
     /* Released app link [launch/sharable link] */
     try {
       return await appsService.validateReleasedApp(slug);
-    } catch (error) {
-      let editPermission = true;
-      try {
-        await appsService.validatePrivateApp(slug, queryParams);
-      } catch (err) {
-        editPermission = false;
-      }
-      handleError(componentType, error, redirectPath, editPermission);
+    } catch (errorResponse) {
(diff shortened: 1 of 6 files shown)


frontend/src/_helpers/handleAppAccess.js
```diff
@@ -26,9 +26,17 @@ export const handleAppAccess = (componentType, slug, version_id) => {
...
+      return await appsService.validateReleasedApp(slug);
+    } catch (error) {
+      let editPermission = true;
+      try {
+        await appsService.validatePrivateApp(slug, queryParams);
+      } catch (err) {
+        editPermission = false;
+      }
+      handleError(componentType, error, redirectPath, editPermission);
+    }
```
shah21, line 39: Instead of doing two APIs to find if the user has edit access to the app or not. we can add an edit access check from the API level and pass some flag or data with the `validateReleasedApp` API error object. 
for reference: https://github.com/ToolJet/tj-ee/blob/f6d3df49042aa7e34b0e23159604856b59198be1/server/src/controllers/apps.controller.ts#L107

shah21 (changes requested):
I have requested some changes
@AnantshreeChandola Can please take a look.

shah21 (dismissed):

gsmithun4 (approved):

PR #8234, opened 2023-12-01
Added code coverage for cypress

emidhun:
Fixes: #8246


frontend/babel.config.js
```diff
@@ -1,18 +1,24 @@
...
+      segments: 1, // NOT REQUIRED
+      // 0 = full file path (Default)
+      // 1 = file name ONLY
+      // 2 = file name and last segment
+      splitSegment: "/", // How to split the path - NOT REQUIRED
+      // Default is / for Linux and OSX
+      // Windows users can use "\\" here if needed
+    },
   ],
+  "@babel/transform-runtime",
```
kavinvenkatachalam, line 14: It was an array previously. Is it removed intentionally?

kavinvenkatachalam (changes requested):
@emidhun Can you check the following comment?

frontend/babel.config.js
emidhun: I have accidentally removed, since there is no other properties with this.
I will revert this and keep the plugin as it is.

kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

PR #8241, opened 2023-12-05
feat: Getter functions for query data and variables

### LINKED ISSUE #8242: Stale data while accessing query response or variables
### Environment
All envs

If I run a query inside of RunJS and access the data of the query in the next line, the data that is returned should be the latest returned data.

If I set a variable inside of RunJS and access the value of the variable in the next line, the value that is returned should be the value that I set the variable with.

If I run a query inside of RunJS and access the data of the query in the next line, the data that is returned is stale/previous data.

If I set a variable inside of RunJS and access the value of the variable in the next line, the value that is returned is state/previous data.

1. Create a RunJS
2. Write: `actions.setVariable('test' 2)`
3. Write `return variables.test`
4. `2` will not be returned on the first execution of the query

Yes

sherfin94:
Resolves #8242

This PR adds the following actions:

* `actions.getVariable('variable')`: Used to get the latest value of `variable` within RunJS and RunPy
* `actions.getPageVariable('pageVariable')`: Used to get the latest value of `pageVariable` within RunJS and RunPy

and the following functions on queries:

* `queries.<queryName>.getData()`: Used to get the latest value of query's data within RunJS and RunPy
* `queries.<queryName>.getRawData()`: Used to get the latest value of query's raw data within RunJS and RunPy
* `queries.<queryName>.getLoadingState()`: Used to get the latest value of query's loading state within RunJS and RunPy

All these functions are synchronous functions and `await` is not needed while calling them

### DIFF
--- a/frontend/src/Editor/CodeBuilder/utils.js
+++ b/frontend/src/Editor/CodeBuilder/utils.js
@@ -72,6 +72,7 @@ export function getSuggestionKeys(refState, refSource) {
   const actions = [
     'runQuery',
     'setVariable',
+    'getVariable',
     'unSetVariable',
     'showAlert',
     'logout',
@@ -82,6 +83,7 @@ export function getSuggestionKeys(refState, refSource) {
     'goToApp',
     'generateFile',
     'setPageVariable',
+    'getPageVariable',
     'unsetPageVariable',
     'switchPage',
   ];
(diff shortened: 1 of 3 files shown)


kavinvenkatachalam (approved):

sherfin94:
Hey @karan-rathod-316 , when you're documenting this, please mention that the `get..` functions  are supposed to be used inside RunJS and RunPy, and to use the regular `variable.myVariable` or `queries.myQuery.data` inside component property/style fields or any other fields.

PR #8247, opened 2023-12-06
[hotfix] Public app is not loading on subpath deployment. 

### LINKED ISSUE #8248: Public app is not loading on subpath deployment. #8247
### Version Information
Latest
### Environment

All

It should load the public app even if the user is not authenticated.

it stuck on loading page. I can't see the app

1. Create an app & release
2. Make app public
3. Open the private tab and access the app.

shah21:
Resolves #8248


gsmithun4 (approved):

vmatt:
Thanks @shah21 , @gsmithun4  for the quick fix!

PR #8264, opened 2023-12-07
docs: Add one click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:

https://tooljet-5796.dome.tools/

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.


jbgh:
@Navaneeth-pk sorry to hear that feel free to give the link a try.  Dome is truly a one click deploy experience for your users.  Let me know if you change your mind or have any feedback.  Thanks!

PR #8265, opened 2023-12-08
aws-lambda plugin

antraxa:
To execute aws lambdas from Tooljet


akshaysasidrn (approved):
Thank you for your contribution @antraxa 🎉 
We'll add docs for these and publish our upcoming marketplace release.

Issue #7050, opened 2023-07-18
Airtable API keys will be removed


C4rlit0:
### Replace deprecated Airtable API key with Airtable Personal Access Token

Airtable API keys will be deprecate on January 2024 
See here : https://support.airtable.com/docs/airtable-api-key-deprecation-notice

### Describe the solution you'd like

You should user Personal Access Token instead
See here : https://airtable.com/developers/web/guides/personal-access-tokens

PR #8295, opened 2023-12-12
bugfix: airtable plugin support for personal access token

ganesh8056:
Support for **Personal Access token** in Airtable plugin.
 #7050

### DIFF
--- a/plugins/packages/airtable/lib/index.ts
+++ b/plugins/packages/airtable/lib/index.ts
@@ -9,11 +9,14 @@ export default class AirtableQueryService implements QueryService {
 
   async run(sourceOptions: SourceOptions, queryOptions: QueryOptions): Promise<QueryResult> {
     let result = {};
+    let apiToken = '';
     let response = null;
     const operation = queryOptions.operation;
     const baseId = queryOptions.base_id;
     const tableName = queryOptions.table_name;
-    const apiKey = sourceOptions.api_key;
+
+    if (sourceOptions.api_key) apiToken = sourceOptions.api_key;
+    if (sourceOptions.personal_access_token) apiToken = sourceOptions.personal_access_token;
 
     try {
       switch (operation) {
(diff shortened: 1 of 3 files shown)


plugins/packages/airtable/lib/manifest.json
```diff
@@ -15,13 +15,20 @@
...
   },
   "defaults": {
     "api_key": {
       "value": ""
+    },
+    "personal_access_token": {
+      "value": ""
     }
   },
   "properties": {
```
akshaysasidrn, line 34: we can remove `api_key` such that it will still work until jan for those who don't update
but at connection options with this change we only use PAT

plugins/packages/airtable/lib/manifest.json
```diff
@@ -31,9 +38,15 @@
       "type": "password",
       "description": "Api key for airtable",
       "helpText": "For generating API key, visit: <a href='https://airtable.com/account' target='_blank' rel='noreferrer'>Airtable account page</a>"
+    },
+    "personal_access_token": {
+      "label": "Personal access token",
+      "key": "personal_access_token",
+      "type": "password",
+      "description": "Personal access token for airtable"
```
akshaysasidrn, line 46: do add helpText for PAT

plugins/packages/airtable/lib/index.ts
```diff
@@ -9,11 +9,14 @@ export default class AirtableQueryService implements QueryService {
...
   async run(sourceOptions: SourceOptions, queryOptions: QueryOptions): Promise<QueryResult> {
     let result = {};
+    let apiToken = '';
     let response = null;
     const operation = queryOptions.operation;
     const baseId = queryOptions.base_id;
     const tableName = queryOptions.table_name;
-    const apiKey = sourceOptions.api_key;
+
+    if (sourceOptions.api_key) apiToken = sourceOptions.api_key;
```
akshaysasidrn, line 18: please add a comment that this is for backward compatibility and needs migration to be removed later on

akshaysasidrn (changes requested):

akshaysasidrn (approved):

Issue #8345, opened 2023-12-19
Add automation for component deletion from inspector.


emidhun:
### Specify the kind of test

AppBuilder Cypress E2E

### Describe the test

- Modify the `inspectorHappypath.cy.jc` spec
- Add a method `deleteComponentFromInspecor()` inside `utils/inspector.js`

PR #8346, opened 2023-12-19
Automation: Deletion of component from inspector, right side panel & via keyboard actions

Nandinisaha13:
fixes: #8345

### DIFF
--- a/cypress-tests/cypress/e2e/editor/inspectorHappypath.cy.js
+++ b/cypress-tests/cypress/e2e/editor/inspectorHappypath.cy.js
@@ -12,8 +12,16 @@ import {
   addSupportCSAData,
 } from "Support/utils/events";
 import { multipageSelector } from "Selectors/multipage";
+import {
+  navigateToCreateNewVersionModal
+} from "Support/utils/version";
+import { createNewVersion } from "Support/utils/exportImport";
+
 
 describe("Editor- Inspector", () => {
+  let currentVersion = "";
+  let newVersion = [];
+  let versionFrom = "";
   beforeEach(() => {
     cy.apiLogin();
     cy.apiCreateApp(`${fake.companyName}-App`);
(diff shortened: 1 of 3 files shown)


cypress-tests/cypress/e2e/editor/inspectorHappypath.cy.js
```diff
@@ -170,4 +170,10 @@ describe("Editor- Inspector", () => {
     cy.notVisible(commonWidgetSelector.draggableWidget("button1"));
     cy.apiDeleteApp();
   });
+  it("should verify deletion of component from inspector", () => {
+    cy.dragAndDropWidget("button", 500, 500);
+    cy.get(commonWidgetSelector.sidebarinspector).click();
+    deleteComponentFromInspector("button1");
+
```
emidhun, line 177: - I think we need to add assertion to verify the deletion of component and the toast message, after the deletion.
- Let's create one more version and verify that the components is not exist on canvas.

cypress-tests/cypress/support/utils/inspector.js
```diff
@@ -32,3 +32,8 @@ export const verifyValue = (node, type, children, index = 0) => {
     .eq(index)
     .verifyVisibleElement("have.text", type);
 };
+export const deleteComponentFromInspector = (node) => {
+  cy.get('[data-cy="inspector-node-components"] > .node-key').click();
+  cy.get(`[data-cy="inspector-node-${node}"] > .node-key`).click();
+  cy.get(cy.get('[style="height: 13px; width: 13px;"] > img').click());
```
emidhun, line 38: I think we can only use `cy.get('[style="height: 13px; width: 13px;"] > img').click()` here.

emidhun (changes requested):
@Nandinisaha13 I have added few changes, Please add the cases and let me know if you need any help.

Nandinisaha13:
@emidhun please take a look have fixed the comments.

emidhun:
@Nandinisaha13  Could you please add the delete functionality from the right side panel as well?
I think Only this case is pending on deletion.
Please let me know if it is possible.
Rest everything is looking good.

emidhun (approved):
LGTM

PR #8379, opened 2023-12-21
Feat/grid revamp

johnsoncherian:


server/src/services/components.service.ts
```diff
@@ -164,6 +167,9 @@ export class ComponentsService {
 
             await manager.update(Layout, { id: componentLayout.id }, layout);
           }
+          if (component) {
```
arpitnath, line 170: add a comment on working of the logic

kavinvenkatachalam (changes requested):
@johnsoncherian Added some suggestions. Please take a look.

Can you remove the unwanted files as well?

server/src/services/components.service.ts
```diff
@@ -259,3 +280,9 @@ export class ComponentsService {
     return componentWithLayout;
   }
 }
+
+function resolveGridPositionForComponent(dimension: number, type: string) {
```
gsmithun4, line 284: Sane function is declared here
https://github.com/ToolJet/ToolJet/pull/8379/files#diff-b5e1de6ff3c19c9d79fa87a3f5ddf3045631a1d3a7c6c375213175e06a6f02afR66
Shall we move it to some util file and reuse?

server/migrations/1702196735455-AddAutoLayoutFlagToPagesTable.ts
```diff
@@ -0,0 +1,30 @@
...
+      })
+    );
+
+    await queryRunner.addColumn(
+      'layouts',
+      new TableColumn({
+        name: 'dimension_unit',
+        type: 'varchar',
+        isNullable: false,
+        default: `'count'`,
```
gsmithun4, line 22: Shall we create an enum for this?

server/src/helpers/utils.helper.ts
```diff
@@ -236,6 +236,13 @@ export function isTooljetVersionWithNormalizedAppDefinitionSchem(version) {
   return semver.satisfies(semver.coerce(version), '>= 2.24.0');
 }
 
+/**
```
gsmithun4, line 239: Shall we create a new util file for component related functions?
`components.helper.ts`

server/src/services/app_import_export.service.ts
```diff
@@ -450,7 +463,12 @@ export class AppImportExportService {
                     const newLayout = new Layout();
                     newLayout.type = type;
                     newLayout.top = layout.top;
-                    newLayout.left = layout.left;
+                    newLayout.left =
+                      layout.dimensionUnit !== 'count'
```
gsmithun4, line 467: Create enum for `dimensionUnit`

server/src/services/apps.service.ts
```diff
@@ -528,6 +528,20 @@ export class AppsService {
 
         oldComponentToNewComponentMapping[component.id] = newComponent.id;
 
+        let parentId = component.parent ? component.parent : null;
+
+        const isParentTabOrCalendar = isChildOfTabsOrCalendar(component, page.components, parentId);
+
+        if (isParentTabOrCalendar) {
+          const childTabId = component.parent.split('-')[component.parent.split('-').length - 1];
```
gsmithun4, line 536: Shall we add Optional chaining here to avoid unexpected errors?

gsmithun4 (changes requested):

server/src/helpers/utils.helper.ts
johnsoncherian: This functions is no longer required so removed it.

server/src/entities/layout.entity.ts
```diff
@@ -24,6 +24,14 @@ export class Layout {
   @Column({ name: 'component_id' })
   componentId: string;
 
+  @Column({
+    type: 'enum',
+    enumName: 'dimension_unit',
+    name: 'dimension_unit',
+    enum: ['signup', 'invite', 'google', 'git'],
```
arpitnath, line 31: The enumeration appears to be incorrect.

arpitnath:
```suggestion
    enum: ['count', 'percent'],
```

arpitnath (changes requested):

server/src/services/app_import_export.service.ts
gsmithun4: Can create an enum like this
https://github.com/ToolJet/tj-ee/blob/develop/server/src/helpers/user_lifecycle.ts#L3

gsmithun4: Best if we can move component related utils to different file, other than utils.helper.ts
can create a directory - `components` under /src/helper/ and add types and utils files there

gsmithun4 (approved):

PR #8387, opened 2023-12-22
Feature/tjdb null contraint support

ganesh8056:
Added Support for **Not Null**  constraint in the following operations in the Backend
1. Create table
2. Add new column

**New API :** 
1. Edit Column - In the Edit column API only the mentioned values can be Modified ( column name, default value, Not null constraint )

### DIFF
--- a/server/src/services/tooljet_db.service.ts
+++ b/server/src/services/tooljet_db.service.ts
@@ -459,9 +459,21 @@ export class TooljetDbService {
 
     if (column?.column_name && column?.new_column_name)
       query += `ALTER TABLE "${internalTable.id}" RENAME COLUMN ${column.column_name} TO ${column.new_column_name};`;
-
-    const result = await this.tooljetDbManager.query(query);
-    await this.tooljetDbManager.query("NOTIFY pgrst, 'reload schema'");
-    return result;
+    const internalTableInfo = {};
+    try {
+      const result = await this.tooljetDbManager.query(query);
+      await this.tooljetDbManager.query("NOTIFY pgrst, 'reload schema'");
+      return result;
+    } catch (error) {
+      internalTableInfo[internalTable.id] = tableName;
+      if (error instanceof QueryFailedError || error instanceof TypeORMError) {
+        let customErrorMessage: string = error.message;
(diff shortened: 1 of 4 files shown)


server/src/services/tooljet_db.service.ts
```diff
@@ -163,12 +165,14 @@ export class TooljetDbService {
       let query = `${column['column_name']} ${column['data_type']}`;
       if (column['column_default']) query += ` DEFAULT ${this.addQuotesIfString(column['column_default'])}`;
       if (column['constraint_type']) query += ` ${column['constraint_type']}`;
+      if (column['isNotNull']) query += ` NOT NULL`;
```
akshaysasidrn, line 168: Does it make sense to change the request body so there can be multiple constraints for creating a table?
We can reuse the constraint_type key's value to accommodate this. It will adhere to SQL syntax itself ie. primary key, not null, uniqueness are all constraints that can be added for a column at hand.

server/src/services/tooljet_db.service.ts
```diff
@@ -430,4 +435,30 @@ export class TooljetDbService {
...
+      } SET DEFAULT ${this.addQuotesIfString(column['column_default'])};`;
+    if ('isNotNull' in column) {
+      column.isNotNull
+        ? (query += `ALTER TABLE "${internalTable.id}" ALTER COLUMN ${column.column_name} SET NOT NULL;`)
+        : (query += `ALTER TABLE "${internalTable.id}" ALTER COLUMN ${column.column_name} DROP NOT NULL;`);
+    }
+
+    if (column?.column_name && column?.new_column_name)
+      query += `ALTER TABLE "${internalTable.id}" RENAME COLUMN ${column.column_name} TO ${column.new_column_name};`;
+    const result = await this.tooljetDbManager.query(query);
```
akshaysasidrn, line 460: Do we need to explicitly handle the case wherein there maybe existing column values which has null values and we are then editing the column to have not null constraint?

akshaysasidrn (changes requested):

akshaysasidrn (approved):

PR #8438, opened 2023-12-28
Feature/create table with not null constraint

ganesh8056:
This PR Consists of 

1. Support for Not Null Constraint in create new table flow in tooljetDB


akshaysasidrn:
image
We need to handle replacing the table messages on error toast on adding rows

akshaysasidrn:
image
Text is not visible in dark mode, we can take the CSS changes together separately if needed.

akshaysasidrn:
image
Create button is enabled without selecting datatype, can we add that as well?

akshaysasidrn (approved):

akshaysasidrn (approved):

PR #8477, opened 2024-01-02
8432 changes in widget config.js

16achint:
Retrieve changes and make new modifications since there is no issue with the background color that needs to be fixed.


karan-rathod-316:
This has been handled in a different PR.

Issue #8331, opened 2023-12-18
TJDB: Ability to modify constraint when editing a row in existing table


akshaysasidrn:
### Is your proposal related to a problem?
Once we add ability to add constraint, we would need means to remove them as well.

PRD: https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-14476

### Describe the solution you'd like

<img width="980" alt="image" src="https://github.com/ToolJet/ToolJet/assets/11629675/c339f38c-4c5e-4c0c-84dd-060fe70aa35b">
- Toggle for columns will have values shown null, default & custom based on the constraints present 
- Disable Record/Row creation if value of any field having NOT NULL constraint is not supplied
- Id column will be disabled for edit

PR #8493, opened 2024-01-04
New design for editng a existing row in ToolJet database

S-Abdul-Rahman:
Completed the edit row new design as well as fixed the id selection issue in edit row (When we select an ID from the dropdown, the values for that particular ID are not changing)  #8331


akshaysasidrn (approved):
Looks good 👍🏾

PR #8540, opened 2024-01-10
Rearranges crucial entry points for navigation inside editor and dasbhoard

vjaris42:
Fixes mobile view of Dashboard


frontend/src/HomePage/FolderFilter/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  const updateFolderQuery = (name, id) => {
+    const path = `${id ? `?folder=${name}` : ''}`;
+    navigate({ pathname: location.pathname, search: path }, { replace: true });
+  };
+
+  useEffect(() => {
+    if (isEmpty(value)) {
+      setCurrentFolder({ name: 'All apps', label: 'All apps', value: '', id: '' });
```
shah21, line 23: Saw this obj is using in 2 places. shall we convert it to a variable and reuse it?
```suggestion
     const defaultFolder = { name: 'All apps', label: 'All apps', value: '', id: '' }
      setCurrentFolder(defaultFolder);
```

shah21 (changes requested):
I have added a minor code improvement. @vjaris42

shah21 (approved):

PR #8552, opened 2024-01-11
Coerce invalid values to empty state values for properties/styles

### LINKED ISSUE #8494: Reduce/Eliminate component crashes ('Something went wrong' error)

Presently, if the properties or styles passed into a component is of an invalid type, component has a tendency to crash and go into 'Something went wrong' error.

Coerce the values supplied onto properties and styles into either a valid type or their default values.

sherfin94:
Helps #8494 

This PR does the following things:

1. Whenever an invalid or undefined value is passed to a property/style, the component will actually receive only an empty state of the expected data type, however, an error will still be shown. For example, if a string value like `"hello"` is passed into Table component's 'data' property, the component will actually receive `[]`, and an error will be shown nonetheless.
2. For a property/style that expects 'string' data type, if a value of 'number' data type is received, then that value is converted into a string
3. Vice versa of 2
4. Changes many property validations that accept both string and number, to just accept string or number, since values are coerced between string and number anyways.

### DIFF
--- a/frontend/src/Editor/Box.jsx
+++ b/frontend/src/Editor/Box.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useMemo, useContext, useRef, memo } from 'react';
+import React, { useEffect, useState, useMemo, useContext, memo } from 'react';
 import { Button } from './Components/Button';
 import { Image } from './Components/Image';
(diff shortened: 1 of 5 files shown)


sherfin94:
Closing this PR as it is no longer relevant

PR #8553, opened 2024-01-11
Added window titles according to pages

### LINKED ISSUE #8114: Changing title of tab as per the name of page
The objective of this feature is to enhance the user experience by dynamically updating the browser tab title based on the currently active page within ToolJet. This improvement will provide users with a clear and intuitive representation of the page they are currently viewing.

1. Page Title Management
- The tab title of the browser should dynamically change to match the name of the currently active ToolJet page.
- The tab title should reflect the name of the page in a user-friendly and concise manner.
- The tab title should revert to a default title (e.g., "ToolJet") when the user logs out or navigates to an unsupported URL.

2. Supported Pages
The tab title should dynamically change for the following pages in ToolJet:

- Marketplace
- Dashboard
- Data sources
- Workspace settings
- Settings
- Audit Logs
- Profile

3. Default Tab Title
The default tab title, when no specific page is active, should be set to "Dashboard."

AnantshreeChandola:
Resolves #8114

### DIFF
--- a/frontend/src/Editor/Viewer.jsx
+++ b/frontend/src/Editor/Viewer.jsx
@@ -24,7 +24,7 @@ import {
 import queryString from 'query-string';
 import ViewerLogoIcon from './Icons/viewer-logo.svg';
 import { DataSourceTypes } from './DataSourceManager/SourceComponents';
-import { resolveReferences, isQueryRunnable } from '@/_helpers/utils';
+import { resolveReferences, isQueryRunnable, setWindowTitle } from '@/_helpers/utils';
 import { withTranslation } from 'react-i18next';
 import _ from 'lodash';
 import { Navigate } from 'react-router-dom';
@@ -285,12 +285,13 @@ class ViewerComponent extends React.Component {
       .fetchAppBySlug(slug)
       .then((data) => {
         const isAppPublic = data?.is_public;
+        const released = data.current_version_id ? true : false;
         if (authentication_failed && !isAppPublic) {
           return redirectToErrorPage(ERROR_TYPES.URL_UNAVAILABLE, {});
         }
         this.setStateForApp(data, true);
         this.setStateForContainer(data);
-        this.setWindowTitle(data.name);
+        setWindowTitle({ page: 'viewer', appName: data.name, released: released });
       })
       .catch((error) => {
         this.setState({
@@ -467,10 +468,6 @@ class ViewerComponent extends React.Component {
     const canvasBoundingRect = document.getElementsByClassName('canvas-area')[0]?.getBoundingClientRect();
     return canvasBoundingRect?.width;
   };
-
-  setWindowTitle(name) {
-    document.title = name ?? 'My App';
-  }
 
   computeCanvasBackgroundColor = () => {
     const bgColor =
--- a/frontend/src/Editor/Viewer.jsx
+++ b/frontend/src/Editor/Viewer.jsx
@@ -24,7 +24,7 @@ import {
 import queryString from 'query-string';
 import ViewerLogoIcon from './Icons/viewer-logo.svg';
 import { DataSourceTypes } from './DataSourceManager/SourceComponents';
-import { resolveReferences, isQueryRunnable, setWindowTitle } from '@/_helpers/utils';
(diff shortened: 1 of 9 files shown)


frontend/src/_helpers/utils.js
```diff
@@ -1060,3 +1060,48 @@ export const determineJustifyContentValue = (value) => {
...
+    'data-sources': 'Data sources',
+    'audit-logs': 'Audit logs',
+    'account-settings': 'Profile settings',
+    settings: 'Profile settings',
+  };
+  const whiteLabelText = defaultWhiteLabellingSettings.WHITE_LABEL_TEXT;
+  let pageTitleKey = pageDetails?.page || '';
+  let pageTitle = '';
+  if (!pageTitleKey) {
+    pageTitleKey = Object.keys(pathToTitle).find((path) => location?.pathname?.includes(path)) || '';
```
shah21, line 1086: I think it is better to be more specific when we check a path against an URL. i think this flashing issue might be the cause of this change

https://github.com/ToolJet/ToolJet/assets/18279004/3ba075e6-7344-462e-ba7a-ca0a750069dc

Adding more details about the above video.
If the app slug contains any of these paths https://arc.net/l/quote/tfxzywll it will flash the path first then change to the app-name

frontend/src/GlobalDatasources/GlobalDataSourcesPage/index.jsx
```diff
@@ -63,6 +64,7 @@ export const GlobalDataSourcesPage = ({ darkMode = false, updateSelectedDatasour
   }, []);
 
   useEffect(() => {
+    setWindowTitle({ page: `${selectedDataSource?.name || 'Data sources'}` });
```
shah21, line 67: **Dev-guidelines Suggestion**
Shall we move these hard coded texts ("Page names") to constant in future.

shah21 (changes requested):
Can you check that flashing issue?. @AnantshreeChandola

frontend/src/_helpers/utils.js
AnantshreeChandola: Resolved the issue, For handling apps and applications, I have added a separate check which will resolve the falshing issue. We have limited main pages on the main screen. I do not think they'll be sharing any of the keywords added in the object in the future as well. So, this should be full proof.

vjaris42 (approved):

frontend/src/_helpers/utils.js
shah21: FYI, Usually when we do paths check we specifically take that subpath and run against the paths array. Thats is more secure way to do.

frontend/src/Editor/Viewer.jsx
```diff
@@ -285,12 +285,13 @@ class ViewerComponent extends React.Component {
       .fetchAppBySlug(slug)
       .then((data) => {
         const isAppPublic = data?.is_public;
+        const released = data.current_version_id ? true : false;
```
shah21, line 288: When we do this check. Lets say a version is release but the user needs to see the preview not the released app. so the both window title will be same. we have to show `Preview - <app-name> | Tooljet` . right ?

Can you please check those behaviours ?

frontend/src/Editor/Viewer.jsx
```diff
@@ -285,12 +285,13 @@ class ViewerComponent extends React.Component {
...
         if (authentication_failed && !isAppPublic) {
           return redirectToErrorPage(ERROR_TYPES.URL_UNAVAILABLE, {});
         }
         this.setStateForApp(data, true);
         this.setStateForContainer(data);
-        this.setWindowTitle(data.name);
+        setWindowTitle({ page: pageTitles.VIEWER, appName: data.name, released: released });
```
shah21, line 294:
```suggestion
        setWindowTitle({ page: pageTitles.VIEWER, appName: data.name, released });
```

shah21 (changes requested):

frontend/src/Editor/Viewer.jsx
AnantshreeChandola: the current behaviour is correct.

shah21 (approved):

PR #8555, opened 2024-01-11
Added group duplication feature

kriks7iitk:
1. Added group duplication feature


gsmithun4 (changes requested):
Please remove all non related file changes,
Vs code settings and package-locks changes are there
And wrap styles inside parent class, don't change any existing shared styles 
@shah21 @vjaris42 please take a look

kriks7iitk:
> Please remove all non related file changes, Vs code settings and package-locks changes are there And wrap styles inside parent class, don't change any existing shared styles @shah21 @vjaris42 please take a look

@gsmithun4 All the theme are wrapped, and seeting.json is file is reverted

Issue #8334, opened 2023-12-18
TJDB: Ability to navigate and edit cells with keyboard


akshaysasidrn:
### Is your proposal related to a problem?

Since we are enabling on click cell edits for ToolJet Database, to make the user experience more intuitive, we will extend keyboard support to navigate between the cells.

 PRD: https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-14356

### Describe the solution you'd like
- Key based navigation once clicked on a cell
⬆️ To move up in the column
⬇️ To move down in the column
⬅️ To move left in the row
➡️ To move right in the row

**Scope of Navigation**
The navigation is only limited to current page of the total records.
Say user id on page 1 of 20 having 20 records each and 4 columns, Upon reach last row if users presses ⬇️ the cursor will not move neither to 21st record on next page nor on record 1 on page 1.

**Limitations of ⬇️ ⬆️ navigation**
The cursor will not move up while pressing ⬆️ on 1st record/row
The cursor will not move down while pressing ⬇️ on last record/row

**Limitations of ⬅️ ➡️ navigation**
The cursor will not move left while pressing ⬅️ on 1st column
The cursor will not move right while pressing ➡️ on last column

PR #8560, opened 2024-01-12
Feature cell navigation in ToolJet database

S-Abdul-Rahman:
completed : #8334


akshaysasidrn (changes requested):
cell-nav
The borderlines seems to be broken with the merge. Also state b/w cells is flickering. Please do check

akshaysasidrn (approved):

PR #8567, opened 2024-01-12
Adding an option to self-hosted list

cosark:
Gives users access one-click deploy template at repocloud.io


Navaneeth-pk:
We are not planning to officially support more platforms at the moment.

PR #8581, opened 2024-01-16
feat: enable or disable row selection on click of cell using property

### LINKED ISSUE #8580: Feat requrest : Avoid event propagating to row when click on table cell
When we click on the input cell of the table, the event is propagated to the row and thus, row selection is toggled

Avoid event propagating on click of an editable cell

Yes

manishkushare:
Resolves #8580


kavinvenkatachalam (changes requested):
@manishkushare As discussed, please make the required changes

kavinvenkatachalam (dismissed):

arpitnath (approved):

Issue #8595, opened 2024-01-18
[docs]: Fix text formatting issue for Marketplace Plugins > Github page


karan-rathod-316:
### Summary

For the text, **GitHub Personal Access Toke**n in the first paragraph - the last letter is not formatted in bold. Please update this. Ensure that the entire text (**GitHub Personal Access Token**) is bolded.

### Issue Type

Documentation bug

### The entire URL of the documentation with the issue

https://docs.tooljet.com/docs/marketplace/plugins/marketplace-plugin-github/

### Steps to reproduce the issue

1. Go to https://docs.tooljet.com/docs/marketplace/plugins/marketplace-plugin-github/
2. Check the first paragraph of the document

![image](https://github.com/ToolJet/ToolJet/assets/58169101/de4a3702-c896-4889-aa5d-644f91db86ae)

### If the issue is confirmed, would you be willing to submit a pull request?

No

### Code of Conduct

- [X] I agree to follow the ToolJet Code of Conduct

ayushjha119:
@karan-rathod-316 can i do this

karan-rathod-316:
Hi @ayushjha119,

I really appreciate your enthusiasm in contributing to ToolJet. Since this issue is marked as a `good-first-issue,` we will try to assign it to a first-time contributor. In case there are no updates, I'll consider your PR and merge it.

ajAYB-code:
Hi @karan-rathod-316 can you assign this issue to me.

manauwarcode:
🆗

PR #8597, opened 2024-01-18
Fix:#8595 Text formatted for marketplacce plugins

ayushjha119:
Fix: #8595 
For the text, GitHub Personal Access Token in the first paragraph - the last letter is not formatted in bold. the letter n was outside ** . made it inside ** so it can be bold


ayushjha119:
Fixes: #8595

Issue #8260, opened 2023-12-07
Map component is crashing on v2.23.0


emidhun:
### Environment
CE, EE

### What is the expected behaviour?
If there is no active google map api key in the .env, 
the map should show `This page didn't load Google Maps correctly. See the JavaScript console for technical details.`
<img width="400" alt="image" src="https://github.com/ToolJet/ToolJet/assets/20367955/5bc976ef-da18-4378-9a27-ead5aa637145">

### What is the current behaviour?
The map component is crashing.
<img width="457" alt="image" src="https://github.com/ToolJet/ToolJet/assets/20367955/51df36d0-64d9-49ea-9d53-963ba13c7be1">

### How to reproduce the issue?

1. Create an app.
2. Drop a map component.

### Please provide any traces or logs that could help here.
<img width="1424" alt="image" src="https://github.com/ToolJet/ToolJet/assets/20367955/03d023ea-53a7-49b2-9277-b57f02c6a9c4">

blank0537:
@emidhun @shubh22 My current version is v2.27.x and was able to replicate it. After dragging the map component in Canvas and when I click outside then only the bug shows up. It happens only if the API key is not available and a user clicks somewhere else.
I Fixed this issue and raised a PR: #8605  for this

DevangAgarwal93:
Issue is fixed in later versions.

PR #8605, opened 2024-01-20
Fix: #8260 Map component crash fix when no google map api key

blank0537:
Fix: #8260 Map component crashes when no Google Map API key is available.


emidhun:
@kavinvenkatachalam please have a look.

Issue #6816, opened 2023-06-22
App Crashes When Passing Invalid Data to Alert Action in runJS


emidhun:
### Version Information
    v2.7.0
### Environment
    ToolJet Cloud / Docker / Linux / MacOS ...

### What is the expected behaviour?
 The app should not crash and  should show an alert with invalid data.

### What is the current behaviour?
 App is crashing on running runjs action for showAlert with an invalid data, instead of text.

### How to reproduce the issue?

1. Create an app.
2. Create a query `actions.showAlert('error',{})`
3. Run the query.

### Please provide any traces or logs that could help here.
<img width="1440" alt="image" src="https://github.com/ToolJet/ToolJet/assets/20367955/a6e792f2-196a-491b-a234-dd51cd9729e8">

### Any possible solutions?
Event should be failed with an error on debugger, invalid data.

blank0537:
@emidhun @shubh22 On current version 2.27.x, I was able to replicate this issue. It also happens for warning and info action alerts. Also, I tried the below cases and in these cases app also crashes 

actions.showAlert('error', () => {});
actions.showAlert('error', []);

I have raised PR #8606 . It has fix for error, warning and info action alerts. 
And incase if i pass null or undefined then it shows toaster but the app doesn't crash but If we consider as invalid data then let me know I'll make the change

PR #8606, opened 2024-01-20
Fix: #6816 App crashes in runjs for alert actions

blank0537:
Fix: #6816 

App crashes when passing invalid data to alert action. In the issue, it's mentioned for error action alert but it also happens for warning and info alerts. So this fix will work for them also.


blank0537:
@emidhun @shubh22 I have made some changes in this commit. Now It will create log in debugger if actions.showAlert has invalid data.

johnsoncherian (approved):

PR #8637, opened 2024-01-24
feat: Add setFilters and clearFilters CSA in table

### LINKED ISSUE #8199: Ability to add filter in table component programatically
As a user using the application's table component, I want to be able to apply and clear filters to the table data using CSA.

Acceptance Criteria:

1. Given the table component in the application, When I call the setFilters function with an array of objects containing keys i.e "column","condition", "value" then the table should update to display only the rows that meet the specified filter criteria.

2. Given the setFilters function, When I provide an array of filter objects with keys that do not correspond to valid ("column","condition", "value") in the table, then those invalid filters should be ignored, and the table should not be affected by them.
 
3. Given the table with existing filters applied, When I call the setFilters function, then all existing filters should be cleared, and the new filters should be applied.
 
4. Given the table with existing filters applied, When I call the clearFilters function, then all existing filters should be cleared.

nakulnagargade:
Resolves #8199.

Usage - `components.table1.setFilters([{column:'name',condition:'contains',value: 'ah'}])`

### DIFF
--- a/frontend/src/Editor/Components/Table/Filter.jsx
+++ b/frontend/src/Editor/Components/Table/Filter.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import Select from '@/_ui/Select';
 import defaultStyles from '@/_ui/Select/styles';
 import { useTranslation } from 'react-i18next';
@@ -96,7 +96,7 @@ export function Filter(props) {
     setTimeout(() => fireEvent('onFilterChanged'), 0);
   }
 
-  React.useEffect(() => {
+  useEffect(() => {
     if (filters.length > 0) {
       const tableFilters = JSON.parse(JSON.stringify(filters));
       const shouldFire = findFilterDiff(activeFilters, tableFilters);
@@ -107,7 +107,8 @@ export function Filter(props) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [JSON.stringify(filters)]);
 
-  React.useEffect(() => {
+  useEffect(() => {
+    // Add CSA to set filters
     setExposedVariable('setFilters', async function (_filters) {
       if (!isArray(_filters)) return;
       const filterArr = [];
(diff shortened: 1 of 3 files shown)


frontend/src/Editor/Components/Table/Filter.jsx
```diff
@@ -93,6 +107,46 @@ export function Filter(props) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [JSON.stringify(filters)]);
 
+  React.useEffect(() => {
```
kavinvenkatachalam, line 110:
```suggestion
  useEffect(() => {
```

frontend/src/Editor/Components/Table/Filter.jsx
```diff
@@ -93,6 +107,46 @@ export function Filter(props) {
...
+        setAllFilters(filterArr.filter((filter) => filter.id !== ''));
+        mergeToFilterDetails({
+          filters: filterArr,
+        });
+      }
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [JSON.stringify(props.columns)]);
+
+  React.useEffect(() => {
```
kavinvenkatachalam, line 140:
```suggestion
  useEffect(() => {
```

frontend/src/Editor/Components/Table/Filter.jsx
```diff
@@ -93,6 +107,46 @@ export function Filter(props) {
...
+    setExposedVariable('clearFilters', async function (_filters) {
+      setAllFilters([]);
+      mergeToFilterDetails({
+        filters: [],
+      });
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [JSON.stringify(props.columns)]);
```
kavinvenkatachalam, line 148: Can we combine both in a single `useEffect` as the dependency is the same? And can we some comments about what's the CSA - [ref](https://github.com/ToolJet/ToolJet/pull/8674/files#diff-6303cb67e509da2bc5279261ca27dc50be29c03ab7c2ba1dcaf6b57a42cb8122R887)?

kavinvenkatachalam (changes requested):
@nakulnagargade Added few suggestions. Please take a look

kavinvenkatachalam (approved):

PR #8638, opened 2024-01-24
migration: Move visibility and disabled states from style to properties

nakulnagargade:
This adds migration to currently revamped components to move visibility and disabled states from style to properties.


arpitnath (changes requested):
1. since there are large number of components (cloud and ee), processing them all at once might not be efficient and could lead to memory issues. Consider batch processing the components in smaller groups to avoid these problems. ref: https://github.com/ToolJet/ToolJet/blob/develop/server/src/helpers/utils.helper.ts#L163 
2. If an error occurs during the migration of a specific component, you should log it and continue with the next one rather than stopping the entire migration.
3. Consider wrapping the entire migration in a database transaction. This way, if an error occurs during the migration, you can roll back the changes made to the components

arpitnath (approved):

PR #8664, opened 2024-01-30
Fix : All fx code blocks are getting active when fxActive is true for a table column

### LINKED ISSUE #8663: Bug : All fx code blocks are getting active when fxActive is true for a table column
### Version Information
v2.27.7

### Environment
Development

We can programmatically set the value for editable and column visibility in the table column manager. To do so, we have to set the fxActive property to true, which in return will show the code box instead of the toggle switch for a respective property.

Upon setting the fxActive property true for a particular property, which in return shows the code box instead of the toggle switch for all other properties as well, instead for a  particular property. This means, let's say, we are setting the fx button to active for editable property only, but in return, the code blocks are rendered not just for editable property, but also for visibility property as well.

1. Drag a table
2. Click on the table to open the table widget inspector
3. Now in the columns section, click on any one of the columns to open the column manager
4. In the column manager, at the bottom, we can two properties which are editable and column visibility.
5. Click on the fx of any one of the above properties to render code block instead of toggle switch.
6. Now click outside of column manager to close it, and again open that same column manager and you can see code block are being rendered , not just for that particular property but for the both.

### Screenshots or Screencast
[fxactive - ToolJet - 30 January 2024 - Watch Video](https://www.loom.com/share/ce9d4ee2f6a8442684699835e1c8fab4)
Yes

manishkushare:
Resolves #8663


arpitnath (changes requested):
Considering the increased complexity and multiple new conditions in the onFxPress callback, it might be beneficial to separate this logic into a distinct function

arpitnath (changes requested):
The code does not clearly indicate how it handles scenarios where `fxActiveFields` might be null or an unexpected data type. Implementing error-handling mechanisms, such as try-catch blocks or setting fallback values, would enhance the robustness of the code in such situations.

arpitnath (approved):

PR #8703, opened 2024-02-05
[docs] custom parameters for queries

withshubh:
#### Changes
- Updated screenshot in queries - concepts
- Revamped the Query Panel doc under App builder
- Added how-to guide on using Custom Query Parameters


karan-rathod-316 (approved):
Looks good to me!

PR #8750, opened 2024-02-08
fix: Move visibility, disabled and tooltip from styles to properties in import export

nakulnagargade:
This moves visibility, disabled from styles and tooltip from general to properties.


server/src/services/app_import_export.service.ts
```diff
@@ -565,6 +566,31 @@ export class AppImportExportService {
     return appResourceMappings;
   }
 
+  // This is to migrate few properties from styles to properties as per new designs
+  migrateVisibilityDisabledStylesToProperties(component: Component, componentTypes: string[]) {
+    const properties = component.properties;
+    const styles = component.styles;
+    const general = component.general;
+
+    if (componentTypes.includes(component.type)) {
```
arpitnath, line 575: Can we consider to handle unexpected scenarios gracefully?

```
if (!componentTypes.includes(component.type)) {
    return { properties: component.properties, styles: component.styles, general: component.general };
  }
```

server/src/services/app_import_export.service.ts
```diff
@@ -565,6 +566,31 @@ export class AppImportExportService {
     return appResourceMappings;
   }
 
+  // This is to migrate few properties from styles to properties as per new designs
```
arpitnath, line 569: Lets provide clear documentation for the function, including its purpose, input parameters, and examples.
```suggestion
  /**
 * Migrates visibility, disabledState, and tooltip styles to properties based on component type.
 * @param {Component} component - The component object to process.
 * @param {string[]} componentTypes - An array of component types that should undergo migration.
 * @returns {Object} An object containing updated properties, styles, and general objects.
 */
```

arpitnath (changes requested):

arpitnath (commented):
Avoid unnecessary operations if component type is not included

`if (!isComponentTypeIncluded(component, componentTypes)) {
    return component;
  }`

server/src/services/app_import_export.service.ts
```diff
@@ -57,6 +57,7 @@ const DefaultDataSourceNames: DefaultDataSourceName[] = [
   'workflowsdefault',
 ];
 const DefaultDataSourceKinds: DefaultDataSourceKind[] = ['restapi', 'runjs', 'runpy', 'tooljetdb', 'workflows'];
+const NewRevampedComponents = ['Text', 'TextInput', 'PasswordInput', 'NumberInput', 'DropDown'];
```
arpitnath, line 60: Using an enum for `NewRevampedComponents` improves code safety, readability, and maintenance

arpitnath (approved):

PR #8812, opened 2024-02-16
feat: Show more specific error messages for query failures in editor

### LINKED ISSUE #8492: Query errors must show exact reason for error instead of 'Internal Server error' or 'Query could not be completed' or 'Something went wrong' in the debugger
**Enhanced Solution for placement of error messages**
Developer: Kartik Gupta
1. Push both the preview and run query error messages to the debugger.
2. Remove the display of pop-up messages in the editor for both preview and run queries.
3. Ensure there is no UX difference for preview and run query error messages inside the debugger.
4. Show only user-generated messages (e.g., "Query success") in pop-ups, while all system error messages should be pushed to the debugger.

**Solution for detailed error messages**
Developer: Johnson

Presently, whenever a query fails, either of the two messages are shown in a toast:

`Internal Server error`
`Query could not be completed`
`Something went wrong` (When RunJS/RunPy fails)

This is not a well specified error message.

Instead of those three static messages, the exact error message should be shown in the toast.

The exact error message should be shown in the debugger entry as well.

johnsoncherian:
Fixes #8492


frontend/src/_helpers/appUtils.js
```diff
@@ -973,6 +973,22 @@ export function previewQuery(_ref, query, calledFromQuery = false, userSuppliedP
             query,
             'edit'
           );
+          if (finalData.status === 'failed') {
+            useSuperStore
+              .getState()
+              .modules[_ref.moduleName].useCurrentStateStore.getState()
```
johnsoncherian, line 979: We need to remove super store reference and access store directly.

Issue #8882, opened 2024-02-21
ToolJet database : active tab issue for default value in ToolJet database when editing a row


S-Abdul-Rahman:
### What is the expected behaviour?

When we click edit row if any of the columns contain the default value in the cell, the active tab should be in the "Default Value" state

### What is the current behaviour?

When we click edit row if any of the columns contain the default value in the cell, currently the active tab is pointing to "Custom" state

### How to reproduce the issue?

1.Login with valid credentials
2.Create a table with valid name and data
3.Add new row and add default value for any one column
4.now try to edit the row which have default value column, you can see its pointing to "Custom" state

PR #8883, opened 2024-02-21
fix : Active tab issue for default value in ToolJet database when editing a row

S-Abdul-Rahman:
ToolJet database : active tab issue for default value in ToolJet database when editing a row #8882


akshaysasidrn:
The default comparison doesn't seem to be working on testing out locally. Please check.
Default value I had kept is `int` type with value _212_ for that column
image
image

akshaysasidrn (dismissed):

akshaysasidrn (approved):

PR #8885, opened 2024-02-21
feat: Updating the format of the name of exported file in table component

### LINKED ISSUE #8725: Updating the format of the name of exported file in table component
Updating the title of the exported file. A simple solution would be to append `“Tablename_DD-MM-YYYY_HH-mm.filetype”` as the file name.

Example: `Customers_25-03-2022_16-10.csv`

nakulnagargade:
Resolves #8725.


frontend/src/Editor/Components/Table/Table.jsx
```diff
@@ -53,6 +53,7 @@ import Skeleton, { SkeletonTheme } from 'react-loading-skeleton';
 import { OverlayTriggerComponent } from './OverlayTriggerComponent';
 // eslint-disable-next-line import/no-unresolved
 import { diff } from 'deep-object-diff';
+import { format } from 'date-fns';
```
kavinvenkatachalam, line 56: Can we use `moment` instead? Looks like this package is not used anywhere in FE and we can remove it

kavinvenkatachalam (changes requested):
@nakulnagargade Added a suggestion

frontend/src/Editor/Components/Table/Table.jsx
nakulnagargade: Done

kavinvenkatachalam (approved):

Issue #8429, opened 2023-12-28
Change labels to sentence case for Checkbox Component


withshubh:
### Problem

There is an inconsistency in the text casing, some labels under `Properties` and `Styles` tab in the component manager are not in [Sentence case](https://www.k-state.edu/grad/academics/etdr/write/sentence-case.html).

### What is the expected behaviour?

All the labels under the `Properties` and `Styles` tab in the component managers should be in the Sentence case.

### What is the current behaviour?

⚠️ The below labels are not in Sentence case(changes are only needed in the below labels).

| Current Values   | Expected Values  |
| :--------------- | :--------------- |
| Default Status      | Default status      |
| Checkbox Color | Checkbox color |

### How to reproduce the issue?

1. Open ToolJet.
2. Create a new app.
3. Search for the component and drag it onto the Canvas.
4. Click on the component to open the Component Manager in the right sidebar.
5. Examine the labels under both the `Properties` and `Styles` tabs. Some labels are not in sentence case and should be updated accordingly.

### Additional Information

1. Changes are only required in the `widgetConfig.js` file, locate the object that has the **Checkbox** component name in the name key, and make changes only in that particular object.  
   **For e.g. for table component, you will make changes in the below object of the widgetConfig.js file:**

```
{
    name: 'Table',
    displayName: 'Table',
    description: 'Display paginated tabular data',
    component: 'Table',
.....
}
```

2. Do not make changes to any other labels, files or components.

### Screenshots or Screencast

**Example** using the table component, items that are not in Sentence case are highlighted in red, these are inconsistent with other labels which are in Sentence case

![image](https://github.com/ToolJet/ToolJet/assets/25361949/a7eca8e0-54b0-4400-a680-12d8d0ff9893)

ayushjha119:
@withshubh Hey i can start working on this , can you assign this to me

karan-rathod-316:
Hi @ayushjha119, let's wait for other first-time contributors to pick this up. We appreciate your interest in contributing to ToolJet.

sakshamg1304:
Hii @withshubh, I can start working on this, can you assign this issue to me ?

karan-rathod-316:
Hi @sakshamg1304, we have assigned a different issue to you. Since this is labelled as a good-first-issue, let's wait for another first-time contributor to pick this one up. If we don't find anyone, we can assign it to you:-).

dineshmickey:
[...]

PR #8951, opened 2024-02-29
fix: Change labels to sentence case for Checkbox Component

prathameshgaikwad:
Solves issue #8429 Change labels to sentence case for Checkbox Component.

Made changes at 4 places, 2 each for "Display Status" and "Checkbox Color", in the widgetConfig.js file.


karan-rathod-316 (approved):
Looks good to me!

PR #8998, opened 2024-03-05
fix: ToolJet database column now allows special characters in the default value.

ganesh8056:
Special characters can now be used in the default value for the tool jet database column.

### DIFF
--- a/server/src/services/tooljet_db.service.ts
+++ b/server/src/services/tooljet_db.service.ts
@@ -1,5 +1,14 @@
 import { BadRequestException, HttpException, Injectable, NotFoundException, Optional } from '@nestjs/common';
-import { EntityManager, In, ObjectLiteral, QueryFailedError, SelectQueryBuilder, Table, TypeORMError } from 'typeorm';
+import {
+  EntityManager,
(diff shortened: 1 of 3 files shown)


server/src/services/tooljet_db.service.ts
```diff
@@ -259,14 +264,29 @@ export class TooljetDbService {
...
-    const result = await this.tooljetDbManager.query(query);
-    await this.tooljetDbManager.query("NOTIFY pgrst, 'reload schema'");
-    return result;
+    try {
+      await tjdbQueryRunnner.addColumn(
+        internalTable.id,
+        new TableColumn({
+          name: column['column_name'],
+          type: column['data_type'],
+          ...(column['column_default'] && { default: this.addQuotesIfString(column['column_default']) }),
```
akshaysasidrn, line 277: I think `this.addQuotesIfString` won't be necessary as TypeORM will handle it and we are not generating the SQL query explicitly

akshaysasidrn (changes requested):

akshaysasidrn (approved):

Issue #9000, opened 2024-03-06
ToolJet database : default value isn't necessary when NOT NULL constraint is added and no row is present in the table 


S-Abdul-Rahman:
### What is the expected behaviour?

while creating a new column default value isn't necessary when NOT NULL constraint is added and no row is present in the table

### What is the current behaviour?

while adding a new column, now the default value is required field even NOT NULL constraint is added for that column

### How to reproduce the issue?

1.Login with valid credentials
2.Create a table with valid name and data
3.click on create new column button then the drawer will open give a column name and select datatype and add NOT NULL constraint, the default value field shows as its required field

PR #9005, opened 2024-03-06
Fix : remove default value when not null constraint is added and table is empty in ToolJet database

S-Abdul-Rahman:
Fix : default value isn't necessary when NOT NULL constraint is added and no row is present in the table #9000


ganesh8056:
LGTM 👍🏻

akshaysasidrn (approved):

akshaysasidrn (approved):

PR #9019, opened 2024-03-06
Chore: ToolJet database on the horizontal scrolling left end has extra space

ganesh8056:
The ToolJet database has more space at the left end when scrolling horizontally.


S-Abdul-Rahman:
Looks good  👍

Issue #8067, opened 2023-10-30
Tooljet database: In filter, unable to pass "not null" value for the "is" operation


Mekhla-Asopa:
### What is the expected behaviour?
In the filter condition, users can input a "not null" value as part of the "is" operation.

### What is the current behaviour?
Getting an error message, while entering the "not null" value for the "is" operation in the filter condition.

### How to reproduce the issue?
1. Login with the valid credentials 
2. Navigate to the tooljet database page
3. Create a table with more rows and columns 
4. Click on the Filter button 
5. Add a filter condition with the "is" operation and enter the "not null" value 

### Screenshots or Screencast

<img width="1440" alt="Screenshot 2023-10-30 at 2 27 39 PM" src="https://github.com/ToolJet/ToolJet/assets/59684099/353c63d7-f448-4b5e-9be2-3d9d9ec5857c">

alpha951:
Please assign this issue to me

Mekhla-Asopa:
@mansukh-tj  Please check, the screencast.

https://github.com/ToolJet/ToolJet/assets/59684099/7a213c84-2a8c-4fa7-86fe-97f443819657

PR #9055, opened 2024-03-11
fix TJDB not null value fail for is operation

blank0537:
### Fix: #8067 
PR contains solution for Issue of TJDB where "is" filter option not working when passing "not null" value in value input field.

Solution video:-

https://github.com/ToolJet/ToolJet/assets/111295371/407e320f-90f3-418a-8d04-a186cf42c7a2

### DIFF
--- a/frontend/src/_helpers/postgrestQueryBuilder.js
+++ b/frontend/src/_helpers/postgrestQueryBuilder.js
@@ -302,7 +302,11 @@ export default class PostgrestQueryBuilder {
    * @param value  The value to filter with.
    */
   filter(column, operator, value) {
-    this.url.append(`${column}`, `${operator}.${value}`);
+    if (operator === 'is' && value === 'not null') {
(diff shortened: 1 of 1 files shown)


mansukh-tj:
If we make "not null" case insensitive, then this feature will become good to merge

Filter should also parse 
1. Not Null
2. NOT NULL
3. Not NULL and relative values

blank0537:
@mansukh-tj @akshaysasidrn Updated PR with this change. Now it will parse "not null" case insensitive value. 
Also handled it for "null" as well. Now it will parse "null" case insensitive values as well. For e.g. nuLL, NULL, Null etc.

akshaysasidrn (approved):

akshaysasidrn (approved):

akshaysasidrn:
Thank you for your contribution @blank0537 🎉

Issue #8556, opened 2024-01-11
Submit Form Issue creating a new version


renegarci:
### Version Information
ToolJet v2.27.3 over AWS ECS

### Environment

    develop (non paid version)

### What is the expected behaviour?
Creating a new version of app we expected that nothing is modified and create the  new app exactly as the previous one, without problems.

### What is the current behaviour?
Creating a new version of app that contains a Form with the "Button to submit form" already configured with some button resource to be used as submit,  this configuration is lost and we need to setting up again. 

### How to reproduce the issue?

1. Inside a modal, create a form.
2. Setting the property "Button to submit form" to use some button introduced in the form. And verify that the button is triggering the submit action.
3. Create a new version of this app and validate that "Button to submit form" is not configured, it is lost. 

### Screenshots or Screencast
A video is provided

### Please provide any traces or logs that could help here.
NA

### Any possible solutions?
No idea

### Can you identify the location in the driver source code where the problem exists?
No idea

### If the bug is confirmed, would you be willing to submit a PR?
Unfortunately no

https://github.com/ToolJet/ToolJet/assets/65568700/4000a675-6d8c-4a2c-adbe-e6470d3f85bc

a-mart:
Any update on this? Happens with any button with event handlers. Makes how we use the product almost impossible

renegarci:
im using the 2.50.20 LTS version and it is solved, i dont know over which version was solved before.

DevangAgarwal93:
Issue does not exist anymore

PR #9065, opened 2024-03-12
Fix form submit button value persistency issue in modal

blank0537:
### FIX: #8556 
PR contains the solution of an issue where when we add form component in modal component and select the option in "Button to submit form" property. After selection when we create new version the selected option of "Button to submit form" is not maintained.

### **Solution video**

https://github.com/ToolJet/ToolJet/assets/111295371/a7ffce84-0c58-4e01-91b7-a3af3c7342ae

### DIFF
--- a/server/src/services/apps.service.ts
+++ b/server/src/services/apps.service.ts
@@ -520,6 +520,8 @@ export class AppsService {
         await manager.save(newEvent);
       });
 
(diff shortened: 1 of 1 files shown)


johnsoncherian:
Closing this since it is fixed in #9449

Issue #8483, opened 2024-01-03
Textarea content gets blur when textarea is placed inside the container


Nandinisaha13:
### What is the expected behaviour?
Textarea should not get blurred when textarea is placed inside container.

### What is the current behaviour?
Textarea content gets blur when textarea is placed inside the container

### How to reproduce the issue?

1. Drag & drop a container
2. Drag & drop textarea inside container
3. Observe
![Screenshot 2024-01-03 at 5 26 07 PM](https://github.com/ToolJet/ToolJet/assets/41302917/4d8bf714-1331-4a80-a5de-174438c98cee)

iamabhshk:
Can I work on this issue?

DevangAgarwal93:
[Fix textarea value blurr issue when inside container #9072](https://github.com/ToolJet/ToolJet/pull/9072)

PR #9072, opened 2024-03-13
Fix textarea value blurr issue when inside container

blank0537:
### FIX: #8483 
PR contains solution of textarea value getting blurred when placed inside container component.
It was happening because textarea had border-radius style. In some browsers or browser versions, this issue appears. A good approach is to wrap it with wrapper div and apply border related styles to it.

### Solution Screenshot
8483-issue


johnsoncherian (approved):

mniami:
What's the reason that this PR was not merged?

Issue #8291, opened 2023-12-12
Templates with ToolJet Database fails to create when ToolJet Database is not enabled


akshaysasidrn:
### Version Information
v2.25.0 and prior

### Environment
all

### What is the expected behaviour?
ToolJet app should show a warning if the data source or TJDB is not connected but should not prevent the creation of such a template. Users should, at a minimum, be able to visualize the UI components of the template, enabling them to customize and build upon them according to their needs.

### What is the current behaviour?

When the user tries to create an application from a template that uses ToolJet Database nothing happens.
Error gets thrown from the server preventing the creation of the app.

### How to reproduce the issue?

1. Set .env variable `ENABLE_TOOLJET_DB` to false
2. Start the application
3. Try creating from template which uses ToolJet Database

### Screenshots or Screencast

Uploading Screen Recording 2023-12-12 at 11.09.35 AM.mov…

### Please provide any traces or logs that could help here.
```
ERROR [12/12/2023, 5:47:04 AM UTC] (6527 on MacBook-Pro-3.local): Cannot read properties of undefined (reading 'query')
    TypeError: Cannot read properties of undefined (reading 'query')
        at TooljetDbService.createTable (/Users/arpitnath/Desktop/_work/ToolJet-cloud/server/src/services/tooljet_db.service.ts:169:35)
```

### Any possible solutions?
Need to scope out.

### Can you identify the location in the driver source code where the problem exists?
https://github.com/ToolJet/ToolJet/blob/develop/server/src/services/import_export_resources.service.ts#L45

adrenalineBot:
From the provided code and error message, it seems like the issue is related to the TooljetDbService.createTable function, which is trying to read the 'query' property of an undefined object. Unfortunately, the provided code snippets do not include this function, so I cannot provide a precise solution.

However, the error message suggests that there might be an issue with how the TooljetDbService is initialized or how it's trying to access the 'query' property. It's possible that the object it's trying to access is not properly initialized or is out of scope at the time createTable is called.

The error occurs when the ENABLE_TOOLJET_DB environment variable is set to false and you try to create an application from a template that uses the ToolJet Database. This suggests that the application might not be properly handling the case when the ToolJet Database is not enabled.

---
This answer was generated by [Adrenaline,](https://useadrenaline.com/) an AI-powered programming assistant. It's not guaranteed to be accurate, but is hopefully useful enough to serve as a starting point for your own research into the issue.

[...]

PR #9082, opened 2024-03-13
Fix create app failure issue because TJDB is disabled

blank0537:
### FIX: #8291 
PR contains solution of an issue where when we create new app using template which has requirement of **Tooljet Database** then it fails to create the app if **Tooljet Database is disabled**. 
As per **expected behaviour** mentioned in Issue, it should create the app by giving warning message.

### Whats In the Solution
I added a check if **Tooljet Database** is disabled then it will trigger a toast message.
Also, one more check I added that if user in query manager selects the tooljet database query and tries to add table name by selecting option **Add new table** from dropdown then also it will trigger toast message saying that you need to connect tooljet database. (Let me know if message needs to be changed)

And one thing is that(my pov) if we let the user to create app even if data source is required because of UI aspect so that he can customize(as per described in issue) then I think we should also remove predefined queries as well which are currently created by templated. If we totally looking from UI perspective then preconfigured queries should not be there. (Open to suggestions).

### Solution Video

https://github.com/ToolJet/ToolJet/assets/111295371/21cba169-8207-43da-88a2-e5622952716d


mansukh-tj:
In this "Tooljet Database is not connected" alert, can you please add "Learn more" with this link https://docs.tooljet.com/docs/tooljet-database/#enabling-the-tooljet-database-for-your-instance

akshaysasidrn (approved):

PR #9128, opened 2024-03-19
[fix] Multiple `ws` binding issues & release button disabled if the proxy don't support `ws` connections

shah21:
**More info about the problem**
1. `yjs` was making multiple `ws` binding connections for every render on `Editor` page.
2. For proxy deployments if the clients don't add the `upgrade` web-socket headers to the configurations list. then the app-builder won't show the release button due the socket enabled check


arpitnath (approved):
LGTM

kavinvenkatachalam (approved):

Issue #9067, opened 2024-03-12
TJDB: As a user, I should be able to use TJDB queries for public applications


akshaysasidrn:
### Is your proposal related to a problem?
Users are unable to use TJDB queries in apps built. Queries when run will get an authentication error.

### Describe the solution you'd like

TJDB queries should be run like any other data source plugin. 
Authentication and authorization required should be handled by the server when querying the tables used in the app.

PR #9134, opened 2024-03-19
Add ability to run TJDB queries from public apps

akshaysasidrn:
closes: #9067


ganesh8056 (approved):

server/src/services/postgrest_proxy.service.ts
```diff
@@ -6,63 +6,61 @@ import * as proxy from 'express-http-proxy';
...
+        method,
+        headers: reqHeaders,
+        responseType: 'json',
+        throwHttpErrors: false, // don't throw for non-2xx status codes
+        ...(!isEmpty(body) && { body: JSON.stringify(body) }),
+      });
+
+      return response.body;
+    } catch (error) {
+      return error;
```
gsmithun4, line 57: Shouldn't we throw error, to understand/display error message on front end? or do we have any other ways to capture error

server/src/modules/auth/auth.module.ts
```diff
@@ -65,6 +66,7 @@ import { SessionScheduler } from 'src/schedulers/session.scheduler';
       },
       inject: [ConfigService],
     }),
+    TooljetDbModule,
```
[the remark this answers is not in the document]
akshaysasidrn: It is due to the dependency of `TooljetDbOperationsService` required in `PluginsHelper`
If I remove it I get the error:
```sh
Error: Nest can't resolve dependencies of the PluginsHelper (PluginRepository, ?). Please make sure that the argument TooljetDbOperationsService at index [1] is available in the AuthModule context.

Potential solutions:
- If TooljetDbOperationsService is a provider, is it part of the current AuthModule?
- If TooljetDbOperationsService is exported from a separate @Module, is that module imported within AuthModule?
```

Ideally, we should remove plugins helper from auth module. But since such refactors will go beyond scope I rather started adding `TooljetDbModule` as a dependency everywhere such that when a refactor happens we can simply delete it from all the irrelevant modules.

server/src/services/postgrest_proxy.service.ts
akshaysasidrn: You are right, we should throw `QueryError` here instead as this will be invoked from query execution.

gsmithun4 (approved):

Issue #9186, opened 2024-03-25
In app builder navigation right click not working


nechalmaggon:
Right click on 'Back to apps' and selecting 'open in new window' is not showing the dashboard

PR #9211, opened 2024-03-26
fix: 9186 app builder navigation right click not working

sid0-0:
Fixes https://github.com/ToolJet/ToolJet/issues/9186


gsmithun4:
Fixed in different PR

PR #9213, opened 2024-03-26
Feature supabase plugin integration

### LINKED ISSUE #9138: Supabase Integration
**Marketplace Inputs**
1. Project URL
2. Service Role Secret
image

**Query Builder Inputs**

1. Data source
2. Table
3. Operations - Count Rows, Create Row, Update Row, Delete Row, Get Rows

image

Filter list:

- Equals
- Not Equals
- Less Than
- Less Than Equal to
- Greater than
- Great than equal to
- is
- is not
- in

API documentation: https://supabase.com/docs/reference/javascript/introduction

blank0537:
Closes #9138 
PR contains an implementation of the Supabase plugin. A few changes in operations/fields that are not as per the given design.

## **MARKETPLACE INPUTS**
**Project Url:** Implemented as per design. Taking supabase project url from user in this field.
**Service Role Secret:** Implemented as per design. Taking supabase service role key from user in this encrypted field.

## **QUERY BUILDER INPUTS**
**Note:** For the Table field, it will appear after the user selects the operation from the dropdown(Also took references from other datasources) and it will be a text field instead of a dropdown because there is no supabase method to fetch tables list.

**Where option:** As per design, Usage of this to filter data. It has three fields. 
         First is text field. It needs column name of table to filter by, 
         Second is dropdown. It has filter list operators. (Provided in issue),
         Third is text field. It needs value to filter.

**Sort:** As per design, Usage of this to sort data. It has two fields.
        First is text field. It needs column name of table which will be used to sort data
        Second is dropdown. It has two options to sort data by ascending and descending.

### **Operations**
All operations are implemented as mentioned in the issue.

### **Count Rows**
**Table:** Pass table name here.(refer note for table)
**Where:** As per design, (refer Where option above)

### **Get Rows**
**Table:** Pass table name here.(refer note for table)
**Where:** As per design, (refer Where filter above)
**Sort:** As per design, (refer Sort above)
**Limit:** As per design, Pass numeric value to get number of rows from table.

### **Create Row(s)**
This does not have **Values** field(as per design). I added **Body** field instead because it needs json object to create row and also supabase has bulk row creation capability so I implemented that instead. Now it can handle single or bulk row creation.
**Table:** Pass table name here.(refer note for table)
**Body:** It takes Array of objects to create row(s) in table.

### **Update Row(s)**
As per my understanding, I guess the expectation here is to update the column by giving value in the table for multiple rows based on filtration. 
This does not have **Values** field(as per design). Instead It has two fields one for **Column Name** and one for **Value**
**Table:** Pass table name here.(refer note for table)
**Where:** As per design, (refer Where option above)
**Column Name:** Pass column name of table. It will update values to given column name.
**Value:** Pass value which needs to be updated to that column

### **Delete Row(s)**
**Table:** Pass table name here.(refer note for table)
**Where:** As per design, (refer Where option above)
**Limit:** As per design, Pass numeric value to get number of rows from table.

## SOLUTION VIDEO (E2E)

https://github.com/ToolJet/ToolJet/assets/111295371/a41cefa2-1e6b-4b6e-bdcd-89c9453876fe

@karan-rathod-316 @2001asjad Let me know if any change required.


mansukh-tj:
We have tested the PR. There are a few issues that we'd like you to fix before we merge this.

1. Get rows - is not working for tables with more than 0 rows
2. Update Row(s) needs a revamp as user should be able to update more than 1 column at once - you may use columns component used in Update row operation of TJDB
3. According to our understanding, the In operator in Where clause isn't working ideally. It should take inputs like this (2,6,7) or ("Jack","John")
4. For count operation, can you just return the { count: n } , n being number of records?

blank0537:
@mansukh-tj 
1. I didn't get this. My table has more than 0 rows and I'm getting data. Please provide an example that didn't work for you.
3. Ok. I thought requirement was accepting single value only. One more thing, is it going to be round brackets? because as per supabase documentation it accepts an array(with square brackets). Let me know the format then I'll make change accordingly.

mansukh-tj:
1. Sharing screen recording

https://github.com/ToolJet/ToolJet/assets/144771579/db686cf3-43ef-4726-be0d-ce5f52c8c4b1



3. Can do as per Supabase's documentation.

blank0537:
1-  Just want to confirm are you using "anon public" key or "service_role" secret key? If its "anon public" key you have to define RLS policy for that table in supabase. if its "service_role" secret key then it will bypass that automatically. 
Also was not able to replicate this error message on my setup with same data on supabase. 
I wonder if network calls are getting resolved correctly on deployed server. Because this error message is not related to supabase. 
From my findings in codebase, Its probably getting triggered from appUtils.js file from catch block of runQuery function.
Once I push new changes (adding few checks to validate) then we can try again to see if it works.
3- ok will move forward with this

blank0537:
@mansukh-tj 
> We have tested the PR. There are a few issues that we'd like you to fix before we merge this.
> 
> 1. Get rows - is not working for tables with more than 0 rows
> 2. Update Row(s) needs a revamp as user should be able to update more than 1 column at once - you may use columns component used in Update row operation of TJDB
> 3. According to our understanding, the In operator in Where clause isn't working ideally. It should take inputs like this (2,6,7) or ("Jack","John")
> 4. For count operation, can you just return the { count: n } , n being number of records?

These are resolved.

Also added some validation error messages after runquery execution
- Select one operation -> when operation from dropdown is not selected
- Table name is required -> if table name not provided in field (for all operations)
- Body required to create rows in table -> if rows data not provided (create rows)
- No column(s) provided to update -> if no column provided (update rows)
- Provide column(s) with valid data -> when column name not provided (update rows)
- Duplicate column keys are not allowed -> when adding duplicate column names (update rows)

server/src/assets/marketplace/plugins.json
```diff
@@ -1,67 +1,75 @@
...
+	{
+		"name": "aws-lambda",
+		"description": "Plugin for aws-lambda",
+		"version": "1.0.0",
+		"id": "aws-lambda",
+		"author": "Tooljet",
+		"timestamp": "Tue, 05 Dec 2023 00:27:16 GMT"
+	},
+	{
+		"name": "supabase",
```
akshaysasidrn, line 60: Let's capitalise here and in the description

frontend/src/_components/DynamicForm.jsx
```diff
@@ -153,6 +154,12 @@ const DynamicForm = ({
         return OpenApi;
       case 'react-component-zendesk':
         return Zendesk;
+      case 'react-component-supabase-filter':
+        return SupabaseFilter;
+      case 'react-component-supabase-sort':
```
akshaysasidrn, line 159: Can we reuse the components added in the PocketBase plugin?

akshaysasidrn (changes requested):
Put a comment, please check @blank0537

server/src/assets/marketplace/plugins.json
blank0537: Done

frontend/src/_components/DynamicForm.jsx
blank0537: Done. Converted SupabaseColumn component to MultiColumn component which can be reused in future by passing value to type as "columns" in operations.json file

blank0537:
@akshaysasidrn mentioned comments resolved.

marketplace/plugins/supabase/lib/index.ts
```diff
@@ -0,0 +1,207 @@
...
+    const supabaseClient = await this.getConnection(sourceOptions);
+
+    if (!supabaseClient) {
+      throw new Error('Invalid credentials');
+    }
+
+    return {
+      status: 'ok',
+    };
+  }
```
akshaysasidrn, line 206: @blank0537 while testing we found that client auth is only done when we run queries. There isn't no officially supported method which we can use for testing session with anon keys.

From fiddling with it, I found we can do something like this instead which would force auth first.

```js
  async testConnection(sourceOptions: SourceOptions): Promise<ConnectionTestResult> {
    const supabaseClient = await this.getConnection(sourceOptions);

    try {
      const res = await supabaseClient.from('').select('1');

      if (res.error) {
        throw new QueryError(`Connection test failed`, res.error, {});
      }

      return {
        status: 'ok',
      };
    } catch (error) {
      throw new QueryError('Connection test failed', error.message, {});
    }
  }
}
```

Could you help make the change?

marketplace/plugins/supabase/lib/index.ts
```diff
@@ -0,0 +1,207 @@
...
+      throw new QueryError('Query could not be completed', error, {});
+    }
+
+    return {
+      status: 'ok',
+      data: result,
+    };
+  }
+
+  async getRows(queryOptions: QueryOptions, supabaseClient: any) {
```
akshaysasidrn, line 70: Also let's get rid of `any` usages, supabase has typescript classes defined. We can use them here.

Point to note here being, 
- let typescript infer class wherever possible
- if we have to declare types lean away from using `any`
- if you have no types available try to use `unknown` instead
- `any` should be used as the last resort

Let typescript make our dev lives easier and sane when type inferences are available :grinning:

akshaysasidrn (changes requested):
Found some issue again on testing, please check @blank0537

marketplace/plugins/supabase/lib/index.ts
blank0537: @akshaysasidrn this is done

marketplace/plugins/supabase/lib/index.ts
blank0537: @akshaysasidrn This is done. Updated Types. I use `any` for third party packages (had some bad experiences in past with some packages 🙂). Let me know if any change required.

blank0537:
@akshaysasidrn mentioned comments resolved.

akshaysasidrn (approved):

akshaysasidrn:
Thank you for the contribution @blank0537!

Issue #9034, opened 2024-03-08
TJDB: As a user I should be able to create primary key on table creation


akshaysasidrn:
### Is your proposal related to a problem?
- The user is unable to change the data type of ID field
- The user is unable to add a custom Primary Key (single and composite)

### Describe the solution you'd like
PRD: [link](https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-13736)
ADR: [link](https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-17976)

Changes:
- New interface for Table schema (repositioning and pinning of columns are skipped)
- Primary key column is always created with `Id` for column name of type serial on create table drawer open
- Primary key selection from single to composite is done with checkboxes
- Primary key selection will enable not null constraint for the specific column
- New icon for primary key

#### Constraints to note:
- **Not-Null**: Primary Key Fields by default become NOT NULL
- **Uniqueness**: Upon data entry, Primary Key Fields have to be checked for uniqueness
- **Data Type Compatibility**: Primary Key Fields must have compatible data types
- **Immutable Data Types**: Once set, primary key values can not be edited
- **Composite Key Limitation**: Data length for composite key will depend upon underlying data length of fields selected to be primary key

![image](https://github.com/ToolJet/ToolJet/assets/11629675/9a43d2c9-bc30-4843-9dd8-5fa7a2789ab0)

Issue #9043, opened 2024-03-08
TJDB: As a user I want to be able to add unique constraint on column to ensure uniquness of data


akshaysasidrn:
### Is your proposal related to a problem?
There is no means for a user to ensure data uniqueness within a column

### Describe the solution you'd like
PRD: [link](https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-13736?block=block-32bcdfeb-e859-43c3-bb3c-f5e9c57d5837)
Design: [link](https://www.figma.com/file/tYa0dxFKAObnxC8Jx0ACzn/TJ-DB?type=design&node-id=4678-139374&mode=design&t=XKXcNaihT6Evfrc9-0)

This constraint can be added to a particular column. This constraint makes sure that all values are unique in this column.
It has to be updated in:
- Create table
- Edit table
- Create column
- Edit column

Note:
1. A unique column can accept NULL values unless an explicitly NULL constraint is added
2. For PKs, this constraint cannot be removed, turning off toggle is disabled

![image](https://github.com/ToolJet/ToolJet/assets/11629675/a53ddd6b-2029-48a1-ba3a-fc3000d61447)
![image](https://github.com/ToolJet/ToolJet/assets/11629675/81f43d46-c776-4e20-8ace-f4299877d8c4)

Issue #9040, opened 2024-03-08
TJDB: As a user I should be ensured with refrential integrity b/w related tables


akshaysasidrn:
### Is your proposal related to a problem?
The user should not be able to directly modify data of PK once set. Therefore the UI should reflect this

### Describe the solution you'd like
PRD: [link](https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-13736?block=block-d74b1b95-ea33-40c9-9436-266a8ed4edcf)

Editing for columns with PK should be disabled in
- Cell navigation on PK column
- Edit row

mansukh-tj:
This issue will also cover
1. Selecting Foreign Key value from a dropdown (having referenced column values of target table) in cell navigation, edit row, create a row, default value in edit column and create column
2. If the value does not exist in target table, throw an error and fail the edit row/create row query

akshaysasidrn:
> This issue will also cover
> 
>     1. Selecting Foreign Key value from a dropdown (having referenced column values of target table) in cell navigation, edit row, create a row, default value in edit column and create column
> 
>     2. If the value does not exist in target table, throw an error and fail the edit row/create row query

handled in https://github.com/ToolJet/ToolJet/issues/9073

Issue #9036, opened 2024-03-08
TJDB: As a user I should be able to edit the Primary key for a table


akshaysasidrn:
### Is your proposal related to a problem?
The user should be able to update the Primary key for a table.

### Describe the solution you'd like

PRD: [link](https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-13736)
ADR:  [link](https://app.clickup.com/37484951/v/dc/13qycq-2704/13qycq-17976)
Design: [link](https://www.figma.com/file/tYa0dxFKAObnxC8Jx0ACzn/TJ-DB?type=design&node-id=4678-145525&mode=design&t=XKXcNaihT6Evfrc9-0)

- From the edit table drawer user should be able to change the PK checkboxes and save changes to the table.
- Upon clicking save, a modal should open which indicate the PK changes
- On save the success/error message should be shown as a toast

Note: Editing PK can be from single to composite or vice-versa

Constraints to note:
#### Constraints to note:
- **Not-Null**: Primary Key Fields by default become NOT NULL
- **Uniqueness**: Upon data entry, Primary Key Fields have to be checked for uniqueness
- **Data Type Compatibility**: Primary Key Fields must have compatible data types
- **Immutable Data Types**: Once set, primary key values can not be edited
- **Composite Key Limitation**: Data length for composite key will depend upon underlying data length of fields selected to be primary key

![image](https://github.com/ToolJet/ToolJet/assets/11629675/ddae0b14-61ba-4129-a03f-64e8e756fb65)

![image](https://github.com/ToolJet/ToolJet/assets/11629675/9ecb1638-249e-4d4a-bb38-21904b216446)

![image](https://github.com/ToolJet/ToolJet/assets/11629675/0bc7be9e-3021-409d-a136-b9e02deedd4f)

PR #9279, opened 2024-04-03
feat : Primary key feature in ToolJet database

S-Abdul-Rahman:
1. The ToolJet database now has the primary key feature. 
2. It is possible to add a unique constraint to a column. 
3. A new data type of serial has been introduced.

Closes : 
#9034 
#9036
#9040
#9043
#9035


frontend/src/TooljetDatabase/Forms/ColumnForm.jsx
```diff
@@ -156,16 +167,18 @@ const ColumnForm = ({ onCreate, onClose, rows }) => {
           <input
             value={defaultValue}
             type="text"
-            placeholder="Enter default value"
+            placeholder={dataType?.value === 'serial' ? 'Auto-generated' : 'Enter default value'}
             className={
-              isNotNull === true && defaultValue.length <= 0 && rows.length > 0 ? 'form-error' : 'form-control'
+              dataType?.value !== 'serial' && isNotNull === true && defaultValue.length <= 0 && rows.length > 0
```
akshaysasidrn, line 172: can we define these conditions in a const variable and reuse them?

```js
const defaultValueRequired = dataType.value !== 'serial' && isNotNull && isEmpty(defaultValue) && !isEmpty(rows) 
```

Also let us always start to stick with to use `cx` from classnames when we are handling condition for CSS classes

frontend/src/TooljetDatabase/Forms/TableSchema.jsx
```diff
@@ -0,0 +1,332 @@
...
+  const lightDisabledBorder = '#dadcde';
+  const lightFocussedBorder = '#90B5E2 !important';
+  const darkFocussedBorder = '#90b5e2 !important';
+  const lightBorder = '#dadcde';
+  const darkBorder = '#dadcde';
+  const dropdownContainerWidth = '360px';
+
+  const CustomSelectOption = (props) => {
+    const isCheckDataType =
+      columns[columnSelection.index]?.column_default?.includes('nextval(') &&
```
akshaysasidrn, line 37: can we also add `::regclass` being present in the default value as well?

We can use a regex like: `/^nextval\(".*"::regclass\)$/`

frontend/src/_components/ConfirmDialog.jsx
```diff
@@ -73,9 +77,12 @@ export function ConfirmDialog({
         </Modal.Header>
       )}
       <Modal.Body className="confirm-dialogue-body" data-cy="modal-message">
+        {isEditToolJetDbTable && (
+          <ChangesComponent currentPrimaryKeyIcons={currentPrimaryKeyIcons} newPrimaryKeyIcons={newPrimaryKeyIcons} />
+        )}
         {message}
       </Modal.Body>
-      <Modal.Footer className="mt-3" style={footerStyle}>
+      <Modal.Footer className={`${footerStyle?.marginTop ? '' : 'mt-3'}`} style={footerStyle}>
```
akshaysasidrn, line 85: use `cx`

akshaysasidrn (changes requested):

akshaysasidrn:
Also don't forget that we would have to add `is_unique` key for [handling imports](https://github.com/ToolJet/ToolJet/blob/feature/tjdb-create-primarykey/server/src/helpers/tjdb_dto_transforms.ts#L12) from older versions.

server/src/services/tooljet_db.service.ts
```diff
@@ -239,22 +248,129 @@ export class TooljetDbService {
...
+        // New columns to be inserted
+        if (!Object.keys(oldColumn).length && Object.keys(newColumn).length) {
+          columnsToBeInserted.push(
+            new TableColumn({
+              name: newColumn?.column_name,
+              type: newColumn?.data_type,
+              ...(newColumn?.column_default && {
+                default:
+                  newColumn?.data_type === 'character varying'
+                    ? this.addQuotesIfString(newColumn?.column_default)
```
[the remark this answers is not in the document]
ganesh8056: If we remove **addQuotesIfString** method ORM is considering string as expression. So it would be needed

PR #9294, opened 2024-04-04
updated the runjs guide along with the formating

2001asjad:


karan-rathod-316 (changes requested):
@2001asjad Looks good overall. Just two minor changes, 

1. Run Multiple Actions From runjs Query should be changed to `Run Multiple Actions From RunJS Query`

2. `Go To App` and `Run Multiple Actions From runjs Query` headers are bigger than the other headers. Add triple-hash before them to fix it (###Go to App).

karan-rathod-316 (approved):
Looks good to me!

johnsoncherian (approved):

Issue #9296, opened 2024-04-04
Dark mode loading state 


nechalmaggon:
The loader animation color should be #3A4251 throughout the product. Find attached below a screen recording of an example of the current loader animation. 

Loading state color in the following screens are to be updated- 
1. App dashboard 
2. TJ Database 
3. Data sources 
4. Settings  (all sub-tabs)
5. Workspace settings (all sub-tabs)
6. Audit logs 

https://github.com/ToolJet/ToolJet/assets/127945292/5405f7f4-572b-4a9f-83f7-476e30ce1afd

Sadaf-A:
Hey @nechalmaggon , I would like to work on this!

karan-rathod-316:
Hi @Sadaf-A, assigning the issue to you. Happy coding :-).

karan-rathod-316:
@Sadaf-A I just want to highlight that this is a critical issue with a deadline of 3 days. Do let me know if you need an extension :-).

Sadaf-A:
> @Sadaf-A I just want to highlight that this is a critical issue with a deadline of 3 days. Do let me know if you need an extension :-).

Sure thing, I'll try my best to complete within the given timeframe

Sadaf-A:
> @Sadaf-A I just want to highlight that this is a critical issue with a deadline of 3 days. Do let me know if you need an extension :-).

Hey, I will need today's time to complete this issue I hope that is alright

karan-rathod-316:
Hi @Sadaf-A,

Just wanted to know if there's an update on this PR. This is a priority issue and needs to be closed soon.

Sadaf-A:
> Hi @Sadaf-A,
> 
> Just wanted to know if there's an update on this PR. This is a priority issue and needs to be closed soon. 

Hey, while I'm trying to solve this issue actively I'm not able to solve it yet. The codebase is new and it may take longer than I expected.

Sadaf-A:
Hey @karan-rathod-316, I believe this should solve the issue please have a look. I have tested it out on my local setup would it be a good idea to attach a demo video?

Thank You

karan-rathod-316:
Hi @Sadaf-A, 

No, video is not needed. Thank you so much for your contribution, will move this to Review.

PR #9329, opened 2024-04-08
Setting "#3A4251" as the highlight color for loader animation in dark mode

Sadaf-A:
Resolves: #9296 

Currently, `react-loading-skeleton` is being used for animated loading skeletons. For dark mode we have overridden the base color and set it to "#2F3C4C" but we did not override the highlight color hence it was coming to be the same as that of light mode.
I have overridden the highlight color to "#3A4251" as requested in the issue.


Sadaf-A:
> Resolves: #9296 
> 
> Currently, react-loading-skeleton is being used for animated loading skeletons. For dark mode we have overridden the base color and set it to "#2F3C4C" but we did not override the highlight color hence it was coming same as in the light mode which is defined in the library itself.
> I have overridden the highlight color to "#3A4251" as requested in the issue.

Issue #9332, opened 2024-04-09
TJDB: As a user, I should have default values preselected on create new row form in the dashboard


akshaysasidrn:
### Is your proposal related to a problem?
For a table that has columns with default values, the user has to go to each column in the form and explicitly have default values selected.

![row-default](https://github.com/ToolJet/ToolJet/assets/11629675/cb5fbd1c-300b-43d3-93f1-c179a3e30ff0)

### Describe the solution you'd like

Have the columns with the default values preselected when the user tries to add a new row and create a drawer opens.

PR #9341, opened 2024-04-10
Feature: Preselect Default Values in Create New Row Form on TJDB Dashboard (Issue: #9332)

parthy007:
Closes: [#9332](https://github.com/ToolJet/ToolJet/issues/9332)

### DIFF
--- a/frontend/src/TooljetDatabase/Forms/RowForm.jsx
+++ b/frontend/src/TooljetDatabase/Forms/RowForm.jsx
@@ -15,7 +15,19 @@ const RowForm = ({ onCreate, onClose }) => {
   const darkMode = localStorage.getItem('darkMode') === 'true';
   const { organizationId, selectedTable, columns } = useContext(TooljetDatabaseContext);
   const [fetching, setFetching] = useState(false);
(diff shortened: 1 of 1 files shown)


frontend/src/TooljetDatabase/Forms/RowForm.jsx
```diff
@@ -15,12 +15,16 @@ const RowForm = ({ onCreate, onClose }) => {
...
   const [inputValues, setInputValues] = useState(
     Array.isArray(columns)
-      ? columns.map((item) => {
-          if (item.column_default !== null) {
+      ? columns.map((item, index) => {
+          if (item.accessor === 'id') {
             return { value: '', checkboxValue: false, disabled: false };
+          }
+          if (item.column_default !== null) {
+            return { value: item.column_default || '', checkboxValue: item.column_default, disabled: true };
```
akshaysasidrn, line 27: Will fail when a user creates a row with bool column which has false as default value.

Based on boolean datatype you would have to set the checkbox value, string will always evaluate as true for bool otherwise

akshaysasidrn:
image
Columns without default values don't have custom label selected. Please check.

akshaysasidrn (approved):

PR #9393, opened 2024-04-16
Fix: duplicated datasource in export of app

kriks7iitk:
1. Fix: Global data source duplicated for exported app.

### DIFF
--- a/server/src/services/app_import_export.service.ts
+++ b/server/src/services/app_import_export.service.ts
@@ -102,6 +102,7 @@ export class AppImportExportService {
           .where('data_sources.appVersionId IN(:...versionId)', {
             versionId: appVersions.map((v) => v.id),
           })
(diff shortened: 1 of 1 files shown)


server/src/services/app_import_export.service.ts
```diff
@@ -102,6 +102,7 @@ export class AppImportExportService {
           .where('data_sources.appVersionId IN(:...versionId)', {
             versionId: appVersions.map((v) => v.id),
           })
+          .andWhere('data_sources.scope != :scope', { scope: DataSourceScopes.GLOBAL })
```
akshaysasidrn, line 105: Can we rename the variable `datasources` assignment here with rather `legacyLocalDataSources`?

gsmithun4 (approved):

Issue #9064, opened 2024-03-12
TJDB: When user bulk uploads CSV, server should verify data for all primary keys are made mandatory except serial data type


akshaysasidrn:
### Is your proposal related to a problem?
The current behaviour supports skipping primary keys, which are always of serial data type. These are auto-generated by the database. Going forward, serial data type is optional and a user can chose whatever type and name they prefer for the primary key.

### Describe the solution you'd like

We should make csv upload behaviour such that:
1. Throw error 'Primary key required for column' if data is missing and the column is not of serial data type
2. For serial data type column, regardless of whether data exists or not within the csv the value should always be autogenerated on the insert

This is currently handled here: https://github.com/ToolJet/ToolJet/blob/develop/server/src/services/tooljet_db_bulk_upload.service.ts

PR #9403, opened 2024-04-17
Feature: Validate Primary Key Presence in Bulk CSV Upload & Autogenerate value of serial data type column

parthy007:
Closes [#9064](https://github.com/ToolJet/ToolJet/issues/9064)


server/src/services/tooljet_db_bulk_upload.service.ts
```diff
@@ -157,6 +168,19 @@ export class TooljetDbBulkUploadService {
       const columnsInCsv = Object.keys(row);
       const transformedRow = columnsInCsv.reduce((result, columnInCsv) => {
         const columnDetails = internalTableColumnSchema.find((colDetails) => colDetails.column_name === columnInCsv);
+
+        const { keytype, data_type } = columnDetails;
+
+        if (keytype === 'PRIMARY KEY' && data_type !== 'serial') {
```
akshaysasidrn, line 174: I don't think we get `data_type` as serial explicitly. We might have to look at the default value and infer that it is of serial type. [[related](https://github.com/ToolJet/ToolJet/pull/9279#discussion_r1552132732)]

server/src/services/tooljet_db_bulk_upload.service.ts
```diff
@@ -157,6 +168,24 @@ export class TooljetDbBulkUploadService {
...
       const transformedRow = columnsInCsv.reduce((result, columnInCsv) => {
         const columnDetails = internalTableColumnSchema.find((colDetails) => colDetails.column_name === columnInCsv);
+
+        const { keytype, data_type, column_default } = columnDetails;
+
+        if (
+          keytype === 'PRIMARY KEY' &&
+          data_type !== 'integer' &&
+          column_default &&
+          !column_default.includes('nextval(')
```
akshaysasidrn, line 178:
```js
const primaryKeyHasDefaultValue = (keytype === 'PRIMARY KEY' && !isEmpty(column_default)

if (primaryKeyHasDefaultValue) {
....
```
Naming the check as such would make better sense, wdyt?

akshaysasidrn: corrected snippet

PR #9431, opened 2024-04-18
Fix : String column types bug fixes

manishkushare:
Resolves 
1. Remove the border bottom when the striped form of the table is selected
2. Hiding max row height field when content wrap is disabled


frontend/src/Editor/Inspector/Utils.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { Code } from './Elements/Code';
 import { QuerySelector } from './QuerySelector';
 import { resolveReferences } from '@/_helpers/utils';
+import { cond } from 'lodash';
```
johnsoncherian, line 5:
```suggestion
```

johnsoncherian (approved):

Issue #8959, opened 2024-03-01
Form Submit Button Resets in New Application Versions


AllenAJ:
### What is the expected behaviour?

  Whenever I create a new version, the form submit button should not reset.

### What is the current behaviour?

  Each time I create a new version, the form submit button gets reset.

### How to reproduce the issue?
1. In the app builder, drag the form component
2. In properties, choose 'Button to submit' option.
3. Now, create a new version(v1 to v2)
4. Go to the form properties again and you'll see that 'Button to submit' option is reset. 

### Screenshots or Screencast
![form v1]( https://img001.prntscr.com/file/img001/mTx10q61StqhtferR-S_gQ.png)
![form v2]( https://img001.prntscr.com/file/img001/3cVbp8LOT4K4Cz69eCNH-w.png)

### If the bug is confirmed, would you be willing to submit a PR?
No

maheshkeswani:
Hey @karan-rathod-316  I m willing to work on this issue .

2001asjad:
Sure @maheshkeswani , go ahead.

PrashantIndurkar:
@maheshkeswani Hello are you still working on this issue?

JunedKhan101:
Hi @2001asjad I'd like to work on this. Kindly assign this to me.

2001asjad:
Hey @maheshkeswani , let us know if you're still working on this issue.

maheshkeswani:
Hey @2001asjad ,I have unassigned myself U Can assing it to someone else

2001asjad:
Hey, @PrashantIndurkar, I am assigning this issue to you. Do let us know if you're moving forward with it.

2001asjad:
Hello @PrashantIndurkar , are you still working on this issue?

2001asjad:
Since there has been no confirmation from the previous assignee, @JunedKhan101 I am assigning this issue to you.

Sadaf-A:
Hey @AllenAJ , 
I'd like to have a go at this if you could please assign it to me.

2001asjad:
@JunedKhan101 , let us know if you're going forward with the issue.

JunedKhan101:
I'm facing some issues to setup the project, you can assign this to someone else.

2001asjad:
@Sadaf-A, I am assigning this issue to you. Thanks for contributing to ToolJet.

Sadaf-A:
@2001asjad 
[...]

PR #9449, opened 2024-04-20
Form Submit Button does not get Reset in New Application Versions anymore

Sadaf-A:
Resolves: #8959 

This PR introduces changes that will make sure that the Form Submit Button does not get Reset after a new version is created. It maps the new ID of the submit button to the new Form in the `createNewPagesAndComponentsForVersion()` function


johnsoncherian (approved):

prajval-tooljet:
Tested and Verified

PR #9661, opened 2024-05-09
Fix for missing foreign key constraint on app import

shaurya-sharma064:
closes: #9695 

Problem:
Importing the app breaks the foreign key constraints in the table.

Fix:
Created foreign keys separately after the creation of tables as keys were not forming due to the sequence of table creation. Had to use a bunch of inner transactions in the process.


akshaysasidrn (changes requested):

server/src/services/tooljet_db.service.ts
```diff
@@ -249,25 +261,27 @@ export class TooljetDbService {
...
+      connectionManagers
     );
 
     if (isFKfromCompositePK)
       throw new ConflictException(
         'Foreign key cannot be created as the referenced column is in the composite primary key.'
       );
 
-    const queryRunner = this.manager.connection.createQueryRunner();
+    const queryRunner = appManager?.queryRunner || appManager.connection.createQueryRunner();
```
akshaysasidrn, line 280: Can't we always call `.connection.createQueryRunner();` ? 
Such that we can then have that specific query released without having to check `transactionDepth`? 🤔

server/src/services/tooljet_db_import_export_service.ts
```diff
@@ -27,8 +37,74 @@ export class TooljetDbImportExportService {
...
+          {
+            table_name: tableName,
+            foreign_keys: foreignKeys,
+          },
+          connectionManagers
+        );
+      }
+
+      await tjdbQueryRunner.commitTransaction();
+      await queryRunner.commitTransaction();
```
akshaysasidrn, line 88: Can we run a posgrest reload schema query after this?

akshaysasidrn (changes requested):

akshaysasidrn (dismissed):

server/src/helpers/tjdb_dto_transforms.ts
```diff
@@ -32,7 +32,7 @@ const transformationsByVersion = {
     });
     dto.schema = {
       columns: transformedColumns,
-      foreign_keys: [],
+      foreign_keys: dto.schema.foreign_keys || [],
```
akshaysasidrn, line 35: For versions which was released before 2.42.0 will not have foreign_keys right?
Why do we have to add it? 🤔

akshaysasidrn (approved):

Issue #9505, opened 2024-04-25
Unable to select text in non-editable table cell


GrowwithNaveen:
**Description:**
This issue occurs when attempting to select text within a cell of a table component that has been configured as non-editable. While the user can click on the cell, unable to highlighting it and drag the cursor to select the text content.

**Current behavior:**
Text within non-editable table cells cannot be selected.

**Expected behavior:**
Users should be able to select text within non-editable table cells, even if editing the content is disabled. This allows for easy copying of the data.

**Steps to reproduce:**
1. Create a table component with cells set to non-editable.
2. Click on a cell containing text.
3. Attempt to drag the cursor across the text to select it.

PR #9725, opened 2024-05-15
fix: selection issue in table row while editing

Kartik18g:
Fixes [9505](https://github.com/ToolJet/ToolJet/issues/9505)


Kartik18g:
Selecto hijacks the selection event of browser by overriding it, which was preventing selection within a table cell.
` dragCondition={(e) => {
        // clear browser selection on drag
        const selection = window.getSelection();
        if (selection) {
          selection.removeAllRanges();
        }
        const target = e.inputEvent.target;
        if (target.getAttribute('id') === 'real-canvas') {
          return true;
        }
        // if clicked on a component, select it and return false to prevent drag
        if (target.closest('.react-draggable')) {
          const closest = target.closest('.react-draggable');
          const id = closest.getAttribute('widgetid');
          const component = appDefinition.pages[currentPageId].components[id].component;
          const isMultiSelect = e.inputEvent.shiftKey;
          setSelectedComponent(id, component, isMultiSelect);
        }
        return false;
      }}`

the drag condition prop allows us to control selecto's behaviour, now selecto only works if a user starts selecting from outside of a container and not within it(Which allows us to select table cell). In order to allow selection of container when it is clicked, we manually select the container by setSelectedComponent instead of letting selecto select it.

Kartik18g:
1.8 merge https://github.com/ToolJet/ToolJet/pull/9944

PR #9760, opened 2024-05-16
Loader added for foreign key dropdown

### LINKED ISSUE #9697: TJDB: As a user, I should know that FKR dropdown values are being loaded

Currently, the values in the dropdown has no loading state and it flickers b/w empty state and the loaded state. The change in information is misleading and can confuse the user.

https://github.com/ToolJet/ToolJet/assets/11629675/2bba2c41-ce05-4efa-a735-069f65bb0dd7

Loading state example:
image

Add a loading state for the dropdown FKR dropdowns used in the dashboard.

shaurya-sharma064:
Loader added for foreign key dropdown

closes #9697

### DIFF
--- a/frontend/src/TooljetDatabase/Forms/TableSchema.jsx
+++ b/frontend/src/TooljetDatabase/Forms/TableSchema.jsx
@@ -13,6 +13,7 @@ import Information from '@/_ui/Icon/solidIcons/Information';
 import DropDownSelect from '../../Editor/QueryManager/QueryEditors/TooljetDatabase/DropDownSelect';
 import tjdbDropdownStyles, { dataTypes, formatOptionLabel, serialDataType, checkDefaultValue } from '../constants';
 import Select, { components } from 'react-select';
+import Skeleton from 'react-loading-skeleton';
 
 function TableSchema({
   columns,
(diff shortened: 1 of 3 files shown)


server/src/dto/tooljet-db.dto.ts
```diff
@@ -119,7 +119,7 @@ export class CreatePostgrestTableDto {
   @MaxLength(31, { message: 'Table name must be less than 32 characters' })
   @MinLength(1, { message: 'Table name must be at least 1 character' })
   @Matches(/^[a-zA-Z0-9_]*$/, {
-    message: 'Table name can only contain letters, numbers and underscores',
+    message: 'Unexpected character found in table name. Table name can only contain alphabets, numbers and underscores',
```
akshaysasidrn, line 122: if we are using fullstops, then end the error message with one as well?

frontend/src/TooljetDatabase/Forms/TableSchema.jsx
```diff
@@ -293,9 +294,17 @@ function TableSchema({
                 emptyError={
                   <div className="dd-select-alert-error m-2 d-flex align-items-center">
                     <Information />
-                    No table selected
+                    No data available
                   </div>
                 }
+                loader={
```
akshaysasidrn, line 300: We should also handle on the foreign key dropdown in the main Dashboard too

akshaysasidrn (changes requested):

akshaysasidrn (approved):

PR #9785, opened 2024-05-20
Feature: Infinite scrolling for searching foreign key values in the ToolJet database.

ganesh8056:
Previously we don't support infinite scroll on searching the Foreign key values in ToolJet database. Now the ToolJet database has a feature that allows for infinite scrolling to search for foreign key values.


frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/SelectBox.jsx
```diff
@@ -71,149 +73,147 @@ function DataSourceSelect({
...
+    setSearchResults([]);
+  }
+
+  function fetchForeignKeyDetails(page, totalRecords, isFirstPageLoaded, searchValue, foreignKeys, organizationId) {
+    const limit = 15;
+    const offset = (page - 1) * limit;
+
+    if (isFirstPageLoaded && offset >= totalRecords) return;
+    setIsLoadingFKDetails(true);
+    const referencedColumns = foreignKeys?.find((item) => item.column_names[0] === cellColumnName);
```
akshaysasidrn, line 90: can we handle foreign keys missing instead of optional chaining?

akshaysasidrn: Try to handle null and undefined cases explicitly instead of relying on optional chaining as these null/undefined values will trickle down and make it hard to reason about code IMHO. Only use them if we can't predict the optional part at all.

akshaysasidrn (changes requested):
Have put some small comments, please take a look.

Also, can we solve this behaviour? It is tedious to scroll to the bottom value every time a re-fetch happens.
infinite-scroll

akshaysasidrn (approved):

akshaysasidrn (dismissed):

akshaysasidrn (approved):

Mapuppy09 (commented):
Good , it needed more about foreign key access please let me know.

Issue #9609, opened 2024-05-03
TJDB: Adding both FILTER and SORT conditions in the query throws error


mansukh-tj:
### Version Information
Version 2.38.0-ee2.15.28-cloud2.3.11

### Environment
ToolJet Cloud 

### What is the expected behaviour?

The query should run smoothly with both conditions

### What is the current behaviour?

If either of SORT or FILTER conditions are used, query works fine.
As soon as both of them are added it throws in error.

### Screenshots or Screencast

**Case 1**
![image](https://github.com/ToolJet/ToolJet/assets/144771579/28f1d3c8-e64c-4243-98f7-2b9bc489dd51)
![image](https://github.com/ToolJet/ToolJet/assets/144771579/73066c8a-8bf5-4e5e-a29b-fa425dd4385c)

**Case 2**
![image](https://github.com/ToolJet/ToolJet/assets/144771579/7f198aa6-5559-469c-9532-9e5e6bdf56fb)
![image](https://github.com/ToolJet/ToolJet/assets/144771579/b2e2128d-88f5-44c7-8e93-31987b819719)

Sadaf-A:
Hey @mansukh-tj !
I would like to work on this issue

mansukh-tj:
@Sadaf-A Assigned!

akshaysasidrn:
@Sadaf-A Kindly let us know if you have already started working on it. This is in priority and we are planning to have it fixed sooner. If you haven't started with it, we'll pick this up internally.

Sadaf-A:
> @Sadaf-A Kindly let us know if you have already started working on it. This is in priority and we are planning to have it fixed sooner. If you haven't started with it, we'll pick this up internally.

Hey, I have started working on this and made some progress I will notify you if I'm not able to solve it on time
Thank You

mansukh-tj:
@Sadaf-A Please let us know by today EOD. If not completed by today, we will be internally assigning it.

Sadaf-A:
> @Sadaf-A Please let us know by today EOD. If not completed by today, we will be internally assigning it.

Hey, I haven't been able to solve it yet

akshaysasidrn:
@Sadaf-A Hope you understand the urgency. This is breaking and we need to resolve it promptly. Therefore taking this up internally.

Sadaf-A:
> @Sadaf-A Hope you understand the urgency. This is breaking and we need to resolve it promptly. Therefore taking this up internally.

[...]

PR #9794, opened 2024-05-21
fix : Adding both filter and sort for tjdb query is not working and it throws error

S-Abdul-Rahman:
Resolves : Adding both filter and sort for tjdb query is not working and it throws error

#9609

### DIFF
--- a/server/src/services/tooljet_db_operations.service.ts
+++ b/server/src/services/tooljet_db_operations.service.ts
@@ -74,7 +74,7 @@ export class TooljetDbOperationsService implements QueryService {
       !isEmpty(offset) && query.push(`offset=${offset}`);
     }
     const headers = { 'data-query-id': queryOptions.id, 'tj-workspace-id': queryOptions.organization_id };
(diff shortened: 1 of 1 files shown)


server/src/services/tooljet_db_operations.service.ts
```diff
@@ -74,7 +74,7 @@ export class TooljetDbOperationsService implements QueryService {
       !isEmpty(offset) && query.push(`offset=${offset}`);
     }
     const headers = { 'data-query-id': queryOptions.id, 'tj-workspace-id': queryOptions.organization_id };
-    const url = `/api/tooljet-db/proxy/${tableId}` + `?${query}`;
+    const url = `/api/tooljet-db/proxy/${tableId}` + `?${query.join('&')}`;
```
akshaysasidrn, line 77: Nit: can we only add this string `?...` when query array has something? The URL which it hits looks unpolished otherwise without any query params

akshaysasidrn (dismissed):

akshaysasidrn (approved):

Issue #9767, opened 2024-05-17
TJDB: Unable to read large text while editing cell with FKR


Mekhla-Asopa:
### What is the expected behaviour?
The large text should be visible in the drop-down

### What is the current behaviour?
Unable to read large text while editing cell with FKR

### How to reproduce the issue?

1. Login with valid credentials
2. Navigate to the database
3. Create two table with required data
4. Create FKR between two tables
5. Edit cell in the source table

### Screenshots or Screencast

https://github.com/ToolJet/ToolJet/assets/59684099/16b3ce10-6c5c-47bd-a7d1-49e5b7d4d48d

PR #9799, opened 2024-05-21
fix : Foreign key dropdown width oversize issue in create and edit row

### LINKED ISSUE #9766: TJDB: The add a row drawer size became changed while adding large text with FKR data
The drawer size should not changed as large text was entered into the field

The drawer size changed as large text was entered into the field

2. Navigate to the database
3. Create two tables
4.  Enter a large test in the cell
5. Create a FKR between two columns with VARCHAR data-type
6. Add a large text in the target table
7. Add a new row in the source table and add large text for the FKR column

### Screenshots or Screencast

https://github.com/ToolJet/ToolJet/assets/59684099/5994379a-7438-4819-9bbd-07e9a341dba5

S-Abdul-Rahman:
Resolves

#9766 , #9767


frontend/src/TooljetDatabase/Menu/CellEditMenu/styles.scss
```diff
@@ -51,4 +51,18 @@
             text-align: center;
         }
     }
+}
+
+.foreignKeyAcces-container {
```
akshaysasidrn, line 56: Access*


since it is a root css class, can we prefix `tjdb`?

akshaysasidrn (changes requested):

akshaysasidrn (dismissed):

akshaysasidrn (dismissed):

akshaysasidrn (approved):

PR #9813, opened 2024-05-21
Fix: Perform the substring search operation in the foreign key values

ganesh8056:
The feature allows for the closest search operation for applicable datatypes on the entire dataset of foreign key values.


akshaysasidrn (changes requested):
Change looks good. Can we also prevent client search happening?
The behaviour seems flaky otherwise.

fk-search

akshaysasidrn (approved):

akshaysasidrn (approved):

Issue #9750, opened 2024-05-16
TJDB: Unable to drag and drop option in the bulk upload


Mekhla-Asopa:
### What is the expected behaviour?
The file can be dragged and dropped while uploading a file using bulk upload

### What is the current behaviour?
Unable to drag and drop a file in the bulk upload

### How to reproduce the issue?
1. Create a table with required column name
2. Click on the bulk upload button
3. Drag and drop file 

### Screenshots or Screencast
https://github.com/ToolJet/tj-ee/assets/100783212/116c7bbd-b339-461e-920e-f25d62080dce

PR #9827, opened 2024-05-23
Fix: Drag & drop functionality in TJDB dropzone while bulkuploading 

parthy007:
Closes: [#9750](https://github.com/ToolJet/ToolJet/issues/9750)


frontend/src/TooljetDatabase/Drawers/BulkUploadDrawer/FileDropzone.jsx
```diff
@@ -23,7 +23,13 @@ export function FileDropzone({
 
   const { getRootProps, getInputProps, isDragActive, acceptedFiles } = useDropzone({
     accept: { parsedFileType: ['text/csv'] },
-    onDrop,
```
akshaysasidrn, line 26: If the prop is no longer used, can we remove it from the props list?

frontend/src/TooljetDatabase/Drawers/BulkUploadDrawer/FileDropzone.jsx
```diff
@@ -23,7 +23,13 @@ export function FileDropzone({
...
+    onDrop: (acceptedFiles) => {
+      if (acceptedFiles.length > 0) {
```
akshaysasidrn, line 27: We will only deal with one file as of now right? I think we can make the code imply that for now?

```
acceptedFiles.length === 1
```

akshaysasidrn (changes requested):

akshaysasidrn (approved):

akshaysasidrn (approved):

PR #9828, opened 2024-05-23
Fix : Caching the foreign key column options 

### LINKED ISSUE #9745: TJ DB: Foreign Key column options are loaded every time an FK cell is clicked
FK column options should not be loaded every time an FK cell is clicked. Instead, first 10 values should be cached so that user has all the options shown as he opens FK cell options. 
Also, for the red banner inside dropdown need to change copy from "No table selected" to "No values found"

### Screenshots or Screencast

https://github.com/ToolJet/ToolJet/assets/144771579/38c96845-1926-4689-8aa4-248ff00b8316

manishkushare:
Resolves #9745


frontend/src/TooljetDatabase/Table/index.jsx
```diff
@@ -89,6 +89,80 @@ const Table = ({ collapseSidebar }) => {
   const [isCellUpdateInProgress, setIsCellUpdateInProgress] = useState(false);
   const [referencedColumnDetails, setReferencedColumnDetails] = useState([]);
 
+  const [cachedOptions, setCahedOptions] = useState([]);
+
+  const fetchFkDataForColumn = (currentColumn, currentIndex) => {
```
ganesh8056, line 94: can we rename **currentColumn** to **foreignKey** if it is meant to be a foreign key

frontend/src/TooljetDatabase/Table/index.jsx
```diff
@@ -89,6 +89,80 @@ const Table = ({ collapseSidebar }) => {
...
+    // responsible to fetch the fk column details for all available fk's and cache them
+    return new Promise((resolve, reject) => {
+      if (!currentColumn?.referenced_column_names?.length) return resolve();
```
ganesh8056, line 97: Rather than checking `currentColumn?.referenced_table_id,` on line 104 should we check here

frontend/src/TooljetDatabase/Table/index.jsx
```diff
@@ -89,6 +89,80 @@ const Table = ({ collapseSidebar }) => {
...
+
+  const fetchAllFkData = async () => {
+    const dataToCache = {};
+
+    const results = await Promise.allSettled(
+      foreignKeys.map((currentColumn, currentIndex) => fetchFkDataForColumn(currentColumn, currentIndex))
+    );
+
+    results.forEach((result, index) => {
+      console.log('kiran ::', { result });
```
ganesh8056, line 147: could you remove the console log

frontend/src/TooljetDatabase/Table/index.jsx
```diff
@@ -89,6 +89,80 @@ const Table = ({ collapseSidebar }) => {
   const [isCellUpdateInProgress, setIsCellUpdateInProgress] = useState(false);
   const [referencedColumnDetails, setReferencedColumnDetails] = useState([]);
 
+  const [cachedOptions, setCahedOptions] = useState([]);
```
ganesh8056, line 92: Since cached options are **Object** by default `const dataToCache = {};` as mentioned in line no 140. Can we make cachedOptions default state as empty object.

akshaysasidrn (changes requested):

akshaysasidrn (approved):

PR #9856, opened 2024-05-24
Fix: Foreign key drop down scroll bar moves to the top on each render

ganesh8056:
**Issue**: Every render, the scroll bar for the foreign key drop-down moves to the top.

**Solution**: Scroll bar should not move to the top, rather it should be at the start of newly added items to the list.


akshaysasidrn (approved):

Asilkan5807 (commented):
- [ ] Hjnnkk

akshaysasidrn (approved):

Asilkan5807 (commented):
d9d1878822a046e550f6642d5c95b69da3c8d747

PR #9873, opened 2024-05-27
Fix : Table is crashing on app load, and selecting the table crashes the whole app.

### LINKED ISSUE #9850: Table crashes on app load and selection due to null column values
**Problem Description:**
Table component crashes upon app loading and selecting the table, crashes the whole application. This problem occurs primarily because columns that have been deleted are still stored as null values. When the table attempts to render or interact with these null values, it leads to a crash, disrupting the user experience.

**Root Cause:**
1. Deleted columns are stored as null values in the table's configuration. These null values are not handled properly during the table rendering and interaction processes.

stepinfwd:
closes #9850


johnsoncherian (approved):

stepinfwd:
Merged to LTS

PR #9877, opened 2024-05-27
fix : Delete column query throws error if the column name starts with uppercase in ToolJet database

S-Abdul-Rahman:
fix : Delete column query throws error if the column name starts with uppercase in ToolJet database

### DIFF
--- a/server/src/services/tooljet_db.service.ts
+++ b/server/src/services/tooljet_db.service.ts
@@ -610,7 +610,7 @@ export class TooljetDbService {
 
     if (!internalTable) throw new NotFoundException('Internal table not found: ' + tableName);
 
-    const query = `ALTER TABLE "${internalTable.id}" DROP COLUMN ${column['column_name']}`;
+    const query = `ALTER TABLE "${internalTable.id}" DROP COLUMN "${column['column_name']}"`;
     try {
       const result = await this.tooljetDbManager.query(query);
       await this.tooljetDbManager.query("NOTIFY pgrst, 'reload schema'");


server/src/services/tooljet_db.service.ts
```diff
@@ -610,7 +610,7 @@ export class TooljetDbService {
 
     if (!internalTable) throw new NotFoundException('Internal table not found: ' + tableName);
 
-    const query = `ALTER TABLE "${internalTable.id}" DROP COLUMN ${column['column_name']}`;
+    const query = `ALTER TABLE "${internalTable.id}" DROP COLUMN "${column['column_name']}"`;
```
akshaysasidrn, line 613: Can you convert this into a parameterised query since the column name is a param?

server/src/services/tooljet_db.service.ts
```diff
@@ -610,12 +610,20 @@ export class TooljetDbService {
...
     try {
-      const result = await this.tooljetDbManager.query(query);
+      const result = await tjdbQueryRunnner.dropColumn(internalTable.id, column['column_name']);
+      await tjdbQueryRunnner.commitTransaction();
       await this.tooljetDbManager.query("NOTIFY pgrst, 'reload schema'");
+      await tjdbQueryRunnner.release();
       return result;
     } catch (error) {
+      await tjdbQueryRunnner.rollbackTransaction();
+      await tjdbQueryRunnner.release();
```
akshaysasidrn, line 626: Since there is only one query, we don't have to start a transaction. Transaction is only needed when we have to rollback multiple queries.

Actually, you can do something like:

```js
try {
  const result = await this.tooljetDbManager.queryRunner.dropColumn(internalTable.id, column['column_name']);

  await this.tooljetDbManager.query("NOTIFY pgrst, 'reload schema'");
  return result;
} catch(error) {
......

```

akshaysasidrn (changes requested):

akshaysasidrn (approved):

PR #9895, opened 2024-05-28
Fix: Remove null values from the foreign key drop-down list

ganesh8056:
Problem:
The Foreign key drop-down should not include null values.

Solution:
If there are any null values, remove them and if there is an empty string, put it at the top of the foreign key drop down list.

### DIFF
--- a/frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/SelectBox.jsx
+++ b/frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/SelectBox.jsx
@@ -190,16 +190,12 @@ function DataSourceSelect({
       columnDataType === 'character varying'
         ? filterQuery.ilike(referencedColumns?.referenced_column_names[0], `%${searchValue}%`)
         : filterQuery.eq(referencedColumns?.referenced_column_names[0], searchValue);
-      // Filtering out null values & bringing empty values to top
-      filterQuery.is(referencedColumns?.referenced_column_names[0], 'notNull');
-      orderQuery.order(referencedColumns?.referenced_column_names[0], 'nullsfirst');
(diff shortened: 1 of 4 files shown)


frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/SelectBox.jsx
```diff
@@ -182,14 +182,23 @@ function DataSourceSelect({
...
+      // Filtering out null values & bringing empty values to top
+      filterQuery.is(referencedColumns?.referenced_column_names[0], 'notNull');
+      orderQuery.order(referencedColumns?.referenced_column_names[0], 'nullsfirst');
+      query = query + `&${filterQuery.url.toString()}&${orderQuery.url.toString()}`;
+    } else {
+      // Filtering out null values & bringing empty values to top
+      filterQuery.is(referencedColumns?.referenced_column_names[0], 'notNull');
+      orderQuery.order(referencedColumns?.referenced_column_names[0], 'nullsfirst');
+      query = query + `&${filterQuery.url.toString()}&${orderQuery.url.toString()}`;
     }
```
akshaysasidrn, line 202: can make the conditional specific?

```js
if (!isEmpty(searchValue)) {
    if (columnDataType === 'character varying') {
        filterQuery.ilike(referencedColumns?.referenced_column_names[0], `%${searchValue}%`);
    } else {
        filterQuery.eq(referencedColumns?.referenced_column_names[0], searchValue);
    }
}

// Filtering out null values & bringing empty values to top
filterQuery.is(referencedColumns?.referenced_column_names[0], 'notNull');
orderQuery.order(referencedColumns?.referenced_column_names[0], 'nullsfirst');

query = query + `&${filterQuery.url.toString()}&${orderQuery.url.toString()}`;
```

akshaysasidrn (approved):

PR #9910, opened 2024-05-29
Hotfix: Salesforce plugin

akshaysasidrn:


marketplace/plugins/salesforce/lib/manifest.json
```diff
@@ -13,11 +13,31 @@
       "rawData": {}
     },
     "options": {
-      "api_version":{
```
akshaysasidrn, line 16: need to add api version back

ganesh8056 (approved):

PR #9921, opened 2024-05-30
Fix table all column editable

nakulnagargade:
1. Fix all columns editable not working as expected.
2. Fix Table pagination reseting to client side even after changing to server side.


johnsoncherian (approved):

cum8197:
a3116ec79482d5744270b0b4e28c48f0838dd0cf

PR #9963, opened 2024-06-03
fix : Table breaking for nested array of objects

stepinfwd:
Try this api in table : https://dummyjson.com/products


stepinfwd:
https://github.com/ToolJet/ToolJet/pull/9984 here

PR #10003, opened 2024-06-06
hotfix: Hidden component's box height not shrinking 

### LINKED ISSUE #10000: Fix `hidden` component's layering and resizing behavior
**Problem Description:**
We have observed a change in the behavior of hidden components in our recent release. Previously, components with visibility set to false were not interactive: they appeared to be shrunk to the very top of the window and were locked, making them non-clickable and adjustable only on width.

However, with the recent update, the borders of hidden components are not getting shrunk, and users can modify both the height and width of these components. 

Root cause:
This change is attributed to the different handling of element dimensions by the new library. Unlike the previous library, which expanded or collapsed components based on their content, the new library does not support this behavior.

**Solution**
To resolve this issue, we need to restore the previous behavior,
1.  Shrink to the very top of the component
2.  Resize on width level only

stepinfwd:
closes #10000

Fix hidden component's layering and resizing behavior


frontend/src/Editor/Container.jsx
```diff
@@ -1046,10 +1046,12 @@ const WidgetWrapper = ({
   const isWidgetActive = (isSelected || isDragging) && mode !== 'view';
```
johnsoncherian, line 1046: Can we check if we need same logic in subcontainer as well?

johnsoncherian (approved):

johnsoncherian (approved):

PR #10016, opened 2024-06-07
fix : Modal component not firing event

stepinfwd:


frontend/src/Editor/Components/Modal.jsx
```diff
@@ -223,6 +223,7 @@ export const Modal = function Modal({
             event.stopPropagation();
             setShowModal(true);
             setExposedVariable('show', true);
+            fireEvent('onOpen');
```
johnsoncherian, line 226: we are already fireing an event from useEffect for exposed variable. Can we remove it or else there might be a chance that some future changes in any of the parent element cause the event to trigger twice

johnsoncherian (approved):

emidhun:
- On closing the modal the event is triggering two times (cloud).
@GrowwithNaveen Please prioritise.

johnsoncherian (approved):

Issue #9538, opened 2024-04-29
Click filter not working on edges in query manager


GrowwithNaveen:
**Description:** Clicking the filter button in the query manager has no effect on certain areas around the edge of the UI.

**Current behavior:** Clicking the filter button does not apply the filter to data points located near the edges of the query manager UI.

**Expected behavior:** Clicking the filter button should apply the filter to all data points within the query manager view, regardless of location.

**Steps to reproduce:**
1. Open the query manager.
2. Add few queries.
3. Apply a filter by clicking the filter button.
4. Observe that data points near the edges of the UI might not be filtered as expected.

yanishbansal:
Hi Naveen, its a bit unclear about the problem and the desired expected behaviour , can you please attach the screenshot for this bug ?

karan-rathod-316:
@GrowwithNaveen

GrowwithNaveen:
@yanishbansal sure attaching a video grab for reference. The cursor only changes to a pointer (clickable) over the text part of the element, not the edges or corners. Ideally, the entire element should be clickable, not just the text.


https://github.com/ToolJet/ToolJet/assets/81761482/3fb58bca-4c8b-437c-8d58-cfc344d49200

2001asjad:
Hey @yanishbansal, would you be going forward with this issue?

609harsh:
hi @2001asjad  please assign this to me i am interested

2001asjad:
Sure @609harsh , assigning this issue to you. Happy coding :)

PriteshKiri:
Hey @609harsh ,

Any updates on this issue?

MK884:
I am interested in solving this issue

PriteshKiri:
Sure! @MK884 

Assigned it to you. Happy coding :)

PriteshKiri:
Thanks for submitting the PR @MK884 . Moving it to review.

PriteshKiri:
Closed with PR https://github.com/ToolJet/ToolJet/pull/10039

PR #10039, opened 2024-06-11
fix:QueryManager filter-list button issue

MK884:
# Problem
  The list of filters in the query manager header is only clickable on textual parts of the list button but is expected to be clickable on the entire button. #9538 

# Changes Made
1. The click function is only applied to `button.UnstyledButton` so I move it into the outer `div` to make entire button clickable 
   File => frontend/src/Editor/QueryPanel/FilterandSortPopup.jsx => MenuButton
2. To make a cursor pointer on the entire button(div) I made changes in styling of class `tj-list-btn` in
   File => frontend/src/_styles/theme.scss

### DIFF
--- a/frontend/src/Editor/QueryPanel/FilterandSortPopup.jsx
+++ b/frontend/src/Editor/QueryPanel/FilterandSortPopup.jsx
@@ -310,8 +310,8 @@ const MenuButton = ({
   };
 
   return (
-    <div className={`field p-2 ${noMargin ? '' : 'mx-1'} tj-list-btn`}>
-      <Button.UnstyledButton onClick={handleOnClick} disabled={disabled} classNames="d-flex justify-content-between">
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/QueryPanel/FilterandSortPopup.jsx
```diff
@@ -310,8 +310,8 @@ const MenuButton = ({
   };
 
   return (
-    <div className={`field p-2 ${noMargin ? '' : 'mx-1'} tj-list-btn`}>
-      <Button.UnstyledButton onClick={handleOnClick} disabled={disabled} classNames="d-flex justify-content-between">
+    <div onClick={handleOnClick} className={`field p-2 ${noMargin ? '' : 'mx-1'} tj-list-btn`}>
```
johnsoncherian, line 313: instead of adding onclick to div can we remove the extra padding to div and increase the size of button? Adding onClick to div might lead to user accessibility issues like unable to navigate by tab button etc

johnsoncherian (approved):

PR #10071, opened 2024-06-13
Fix password input width not rendered correctly on editor mount

nakulnagargade:


johnsoncherian:
Already fixed in https://github.com/ToolJet/ToolJet/pull/10509

PR #9922, opened 2024-05-30
Performance release 0.1

kavinvenkatachalam:
# Reduce the current state dependency 
### Improvements
- [x] Improves rendering of query manager on initial app load
- [x] Fixed: performance bottlenecks with our current state management, The old setter function, which updates the entire state tree, is causing significant delays and unnecessary re-renders.
- [x] Fixes: Order of query executions for app load an page load. Expected order- app load queries >> page load queries
- [x] Fixes: page load queries should be triggered even if the one or all the app load queries fails
- [x] - reduced calling resolver functions upto 50% times for components' rendering
      - moved to new resolver function for resolving component properties: more efficient as 
      -  utilises hash table to resolve references and only uses recursive function for 
      resolving JS code by inferring the query
      - reduced re-rendering of Box component upto 50-60%
      - app load time improved by 30% : as resolver function is more efficient and 50% 
      reduction on rendering of Box.jsx
- [x] Remove current state dependencies from query manager
- [x] Remove current state dependencies from utilities functions
- [x] Improvements Container and Sub-container (drag layers) to reduce re-rendering and efficient computations

PR #10104, opened 2024-06-17
[fix]: Replaced regex logic with acorn for reserved keywords

kavinvenkatachalam:
Regex logic searches the complete code and returns true even if the code has reserved code attached with any other texts. To fix that, replaced regex with acorn


kavinvenkatachalam:
Fixed in a diff PR #9922. Closing this

PR #10128, opened 2024-06-19
Fix dropdown and multiselect crash on integer labels

vjaris42:


frontend/src/Editor/Components/DropdownV2/CustomOption.jsx
```diff
@@ -14,7 +14,7 @@ const CustomOption = (props) => {
           </span>
         )}
         <span style={{ color: props.isDisabled ? '#889096' : 'unset', wordBreak: 'break-all' }}>
-          {highlightText(props.label, props.selectProps.inputValue)}
+          {highlightText(props.label.toString(), props.selectProps.inputValue)}
```
johnsoncherian, line 17: Can we add a null check for label? Else if label is undefined or null from a resolved value it might crash

johnsoncherian (approved):

PR #10175, opened 2024-06-25
New Postgrest change for render related to Aggregate and group by fun…

adishM98:
…ctionality


.github/workflows/render-preview-deploy.yml
```diff
@@ -80,7 +80,19 @@ jobs:
...
+                "key": "TOOLJET_DB_STATEMENT_TIMEOUT",
+                "value": "60000"
+              },
+              {
+                "key": "TOOLJET_DB_RECONFIG",
+                "value": "true"
+              },
+              {
+                "key": "PGRST_DB_ENABLE_AGGREGATE",
+                "value": "info"
```
ganesh8056, line 94: value is not "infor" it should be boolean "true"

PR #10239, opened 2024-07-01
Fix : Bugs from aggregate feature

manishkushare:
Resolves
1. Right after adding first aggregate condition, hiding the select section in join table operation
2. Deleting an only available aggregate condition, will result in deleting all available groups by values
3. Adjusted width of the table name in group by section and tooltip, avoiding rendering dropdown when joining table is not selected for the same.
4. show delete confirmation modal, when we are deleting last filled value , and all other values are half filled or empty

### DIFF
--- a/frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/AggregateUI/index.jsx
+++ b/frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/AggregateUI/index.jsx
@@ -271,12 +271,10 @@ export const AggregateFilter = ({ darkMode, operation = '' }) => {
       return option || {};
     }
     if (option === 'column') {
-      if (option === 'column') {
-        if (operation === 'joinTable') {
-          return getJoinTableOption(value, tableId);
-        } else if (operation === 'listRows') {
-          return getListRowsOption(value);
-        }
+      if (operation === 'joinTable') {
+        return getJoinTableOption(value, tableId);
+      } else if (operation === 'listRows') {
+        return getListRowsOption(value);
       }
     }
   };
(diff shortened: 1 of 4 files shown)


frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/JoinTable.jsx
```diff
@@ -89,8 +89,10 @@ const SelectTableMenu = ({ darkMode }) => {
 
   const showSelectSection = () => {
     const groupBy = joinTableOptions?.group_by || {};
+    const aggregates = joinTableOptions?.aggregates || {};
     const isGroupByUsed = Object?.values(groupBy)?.some((condition) => condition?.length >= 1);
```
ganesh8056, line 93: condition can be renamed to **columnList**

frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/SelectBox.jsx
```diff
@@ -197,11 +197,11 @@ function DataSourceSelect({
...
 
-  let optionsCount = options.length;
+  let optionsCount = options?.length;
 
-  options.forEach((item) => {
-    if (item.options && item.options.length > 0) {
-      optionsCount += item.options.length;
+  options?.forEach((item) => {
+    if (item?.options && item?.options?.length > 0) {
+      optionsCount += item.options?.length;
```
ganesh8056, line 204: Since validation is done on line 203,  optional chaining would not be needed here wdyt ?

frontend/src/Editor/QueryManager/QueryEditors/TooljetDatabase/AggregateUI/index.jsx
```diff
@@ -244,35 +243,40 @@ export const AggregateFilter = ({ darkMode, operation = '' }) => {
...
-              }
-            });
-            return acc;
-          }, {});
-          return option || {};
-        }
-        case 'listRows': {
-          const option = columnAccessorsOptions?.find((option) => option?.value === value);
-          return option || {};
+      if (option === 'column') {
```
ganesh8056, line 274: if condition duplicated

akshaysasidrn (approved):

akshaysasidrn (approved):

PR #10283, opened 2024-07-05
Getting 403 error when using multiple apps of different workspace in different tabs (public-apps) 

shah21:
#### How to reproduce

Open AppA of workspaceA in preview mode
Open AppB of workspaceB in edit mode.
refresh AppB
refresh AppA - App will redirect to invalid link


gsmithun4:
Releasing on next LTS

PR #10403, opened 2024-07-18
Resizing codehinter based on suggestion

shaurya-sharma064:
Issue: Suggestion would get cropped due to fixed text length.
Fix: Changed codehinter CSS width property


johnsoncherian:
Need to relook into UI with design team

PR #10455, opened 2024-07-25
fix/tab-switch-event

shaurya-sharma064:
Issue: Tab switch event was being fired even when the selected tab was clicked.
Fix: Added the check for checking if tab selected is the current tab or not.

### DIFF
--- a/frontend/src/Editor/Components/Tabs.jsx
+++ b/frontend/src/Editor/Components/Tabs.jsx
@@ -96,7 +96,7 @@ export const Tabs = function Tabs({
   useEffect(() => {
     const exposedVariables = {
       setTab: async function (id) {
-        if (id) {
+        if (id && currentTab !== id) {
           setCurrentTab(id);
           setExposedVariable('currentTab', id);
           fireEvent('onTabSwitch');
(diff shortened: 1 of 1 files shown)


frontend/src/Editor/Components/Tabs.jsx
```diff
@@ -162,6 +162,7 @@ export const Tabs = function Tabs({
             className="nav-item"
             style={{ opacity: tab?.disabled && '0.5', width: tabWidth == 'split' && '33.3%' }}
             onClick={() => {
+              if (currentTab === tab.id) return;
```
TooljetBot, line 165: <div>

<strong>Bito Code Review Agent Run</strong> <a href=https://github.com/ToolJet/ToolJet/pull/10455#issuecomment-2249755641>#211828</a><span> - 07/25/2024, 08:27 am</span>
<br>

<div id="suggestion">
<div id="issue">
<b>Issue:</b> The early return statement 'if (currentTab === tab.id) return;' prevents the rest of the onClick handler from executing if the current tab is already selected. This is a good optimization but it should be placed after the check for 'tab?.disabled' to ensure that the disabled state is respected.
</div>
<div id="fix">
<b>Fix:</b> Move the early return statement after the check for 'tab?.disabled' to ensure that the disabled state is respected.
</div>
<details>
<summary>
<b>Code suggestion</b>
</summary>
<div id="code">

```
 @@ -164,0 +165,0 @@
 +              if (tab?.disabled) return;
 +              if (currentTab === tab.id) return;
```
</div>
</details>
</div>
<br>


</div>

johnsoncherian (approved):

frontend/src/Editor/Components/Tabs.jsx
shaurya-sharma064: No point as none of the code should be executed if currentTab === tab.id

PR #10477, opened 2024-07-26
fix/listview-record-click

shaurya-sharma064:
Issue: Record click event is fired before the state is updated causing the previous values to be displayed.
Fix: Have added a setTimeout to send the fireEvent event to the end of the task queue so state is updated first.

### DIFF
--- a/frontend/src/Editor/Components/Listview.jsx
+++ b/frontend/src/Editor/Components/Listview.jsx
@@ -62,7 +62,9 @@ export const Listview = function Listview({
     if (isEditorReady) {
       setExposedVariables(exposedVariables);
     }
-    fireEvent('onRecordClicked');
+    setTimeout(() => {
+      fireEvent('onRecordClicked');
+    }, 0);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }
   function onRowClicked(index) {
@@ -74,7 +76,9 @@ export const Listview = function Listview({
     if (isEditorReady) {
       setExposedVariables(exposedVariables);
     }
-    fireEvent('onRowClicked');
+    setTimeout(() => {
+      fireEvent('onRowClicked');
+    }, 0);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }
 


frontend/src/Editor/Components/Listview.jsx
```diff
@@ -62,7 +62,9 @@ export const Listview = function Listview({
     if (isEditorReady) {
       setExposedVariables(exposedVariables);
     }
-    fireEvent('onRecordClicked');
+    setTimeout(() => {
+      fireEvent('onRecordClicked');
+    }, 0);
```
TooljetBot, line 67: <div>

<strong>Bito Code Review Agent Run</strong> <a href=https://github.com/ToolJet/ToolJet/pull/10477#issuecomment-2252045944>#3dd014</a><span> - 07/26/2024, 06:13 am</span>
<br>

<div id="suggestion">
<div id="issue">
<b>Issue:</b> Using 'setTimeout' with a delay of 0 to ensure the state is updated before firing the event is a workaround that can lead to unpredictable behavior. This approach relies on the JavaScript event loop and may not be reliable across different environments or future changes in the codebase. A similar issue was also found in frontend/src/Editor/Components/Listview.jsx (line 79-81).
</div>
<div id="fix">
<b>Fix:</b> Instead of using 'setTimeout', consider using a 'useEffect' hook to watch for changes in the state and then fire the event. This ensures that the event is fired only after the state has been updated.
</div>
<details>
<summary>
<b>Code suggestion</b>
</summary>
<div id="code">

```
 @@ -65,3 +65,3 @@
 -    setTimeout(() => {
 -      fireEvent('onRecordClicked');
 -    }, 0);
 +    useEffect(() => {
 +      if (selectedRowIndex !== undefined) {
 +        fireEvent('onRecordClicked');
 +      }
 +    }, [selectedRowIndex]);
```
</div>
</details>
</div>
<br>


</div>

shaurya-sharma064: Have been used in multiple places in the code
Also using a useEffect will lead to having to create another state to track number of presses

shaurya-sharma064:
No longer needed fixed in lts-3.0 separately

PR #10499, opened 2024-07-29
Fix/inspector query run preview

shaurya-sharma064:
Issue: When we hit run query from inspector, preview is not visible in query manager.
Fix: Made changes such that query manager switches to the selected query and sets the preview.

### DIFF
--- a/frontend/src/Editor/Editor.jsx
+++ b/frontend/src/Editor/Editor.jsx
@@ -626,8 +626,14 @@ const EditorComponent = (props) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const handleRunQuery = (queryId, queryName, ...additionalArgs) => {
-    runQuery(getEditorRef(), queryId, queryName, ...additionalArgs);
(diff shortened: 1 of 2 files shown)


frontend/src/Editor/Editor.jsx
```diff
@@ -626,7 +626,9 @@ const EditorComponent = (props) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const handleRunQuery = (queryId, queryName) => runQuery(getEditorRef(), queryId, queryName);
+  const handleRunQuery = (queryId, queryName, ...additionalArgs) => {
```
johnsoncherian, line 629: a better approach here is to pass additionalArgs as an object.
Reason: your don't have to worry about order and setting value to null if one param in between is not requied

johnsoncherian (commented):
a better approach here is to pass additionalArgs as an object.
Reason: your don't have to worry about order and setting value to null if one param in between is not requied

johnsoncherian (approved):

PR #10531, opened 2024-07-31
Fix/query state issue

Kartik18g:
On currentState store initialisation we update state in `idleCallback` and pass data to update as a param. This causes issues because if state is manipulated during the time updation is in queue, update overrides the state. This pr updates the function to consume latest state using `getState`.

### DIFF
--- a/frontend/src/_stores/currentStateStore.js
+++ b/frontend/src/_stores/currentStateStore.js
@@ -106,16 +106,18 @@ useCurrentStateStore.subscribe((state) => {
 
     handleLowPriorityWork(
       () => {
+        const currentState = useCurrentStateStore.getState();
         useResolveStore.getState().actions.addAppSuggestions(
           {
-            queries: state.queries,
-            components: !isPageSwitched ? state.components : {},
-            globals: state.globals,
-            page: state.page,
-            variables: state.variables,
-            client: state.client,
-            server: state.server,
-            constants: state.constants,
+            // get queries from current state so that we can get the latest queries
+            queries: currentState.queries,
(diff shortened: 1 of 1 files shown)


frontend/src/_stores/currentStateStore.js
```diff
@@ -106,16 +106,18 @@ useCurrentStateStore.subscribe((state) => {
 
     handleLowPriorityWork(
       () => {
+        const currentState = useCurrentStateStore.getState();
```
TooljetBot, line 109: <div>

<strong>Bito Code Review Agent Run</strong> <a href=https://github.com/ToolJet/ToolJet/pull/10531#issuecomment-2259638588>#b3e7e8</a><span> - 07/31/2024, 04:51 am</span>
<br>

<div id="suggestion">
<div id="issue">
<b>Issue:</b> The current change fetches the latest state using 'useCurrentStateStore.getState()', but it does not ensure that the state update is atomic. This can still lead to race conditions if multiple updates are queued simultaneously.
</div>
<div id="fix">
<b>Fix:</b> To ensure atomic updates and avoid race conditions, use Zustand's 'set' function with a callback that receives the current state. This ensures that the state update is based on the most recent state.
</div>
<details>
<summary>
<b>Code suggestion</b>
</summary>
<div id="code">

```
 @@ -108,0 +109,1 @@
 +        useCurrentStateStore.set(state => ({ ...state, ...newState }));
```
</div>
</details>
</div>
<br>


</div>

arpitnath (approved):

johnsoncherian (approved):

PR #10599, opened 2024-08-07
[Hotfix] - Add TLS settings for redis plugin

xwellingtonx:
The Redis plugin was not exposing the TLS settings on the UI, which makes the usage of Azure Redis Cache impossible. This PR intention is to fix that.

Obs: This Redis plugin might have to be reviewed again, since at the current version it won't work with Redis clusters.

Video: 
[20240807-0928-21.6391904.zip](https://github.com/user-attachments/files/16522621/20240807-0928-21.6391904.zip)

### DIFF
--- a/plugins/packages/redis/lib/index.ts
+++ b/plugins/packages/redis/lib/index.ts
@@ -1,16 +1,6 @@
-import { ConnectionTestResult, QueryError, QueryResult, QueryService } from '@tooljet-plugins/common';
+import { ConnectionTestResult, isEmpty, QueryError, QueryResult, QueryService } from '@tooljet-plugins/common';
 import Redis from 'ioredis';
 import { SourceOptions, QueryOptions } from './types';
-
-function isEmpty(value: number | null | undefined | string) {
-  return (
-    value === undefined ||
-    value === null ||
-    !isNaN(value as number) ||
-    (typeof value === 'object' && Object.keys(value).length === 0) ||
-    (typeof value === 'string' && value.trim().length === 0)
-  );
-}
 
 export default class RedisQueryService implements QueryService {
   connectionOptions(sourceOptions: SourceOptions) {
--- a/plugins/packages/redis/lib/index.ts
+++ b/plugins/packages/redis/lib/index.ts
@@ -45,8 +45,8 @@ export default class RedisQueryService implements QueryService {
       if (sourceOptions.tls_certificate === 'ca_certificate') {
         tls.ca = sourceOptions.ca_cert;
       }
-      if (sourceOptions.tls_certificate === 'self_signed') {
-        tls.ca = sourceOptions.root_cert;
+      if (sourceOptions.tls_certificate === 'client_certificate') {
+        tls.ca = sourceOptions.ca_cert;
         tls.key = sourceOptions.client_key;
         tls.cert = sourceOptions.client_cert;
       }
--- a/plugins/packages/redis/lib/index.ts
+++ b/plugins/packages/redis/lib/index.ts
@@ -2,7 +2,30 @@ import { ConnectionTestResult, QueryError, QueryResult, QueryService } from '@to
 import Redis from 'ioredis';
 import { SourceOptions, QueryOptions } from './types';
 
+function isEmpty(value: number | null | undefined | string) {
+  return (
+    value === undefined ||
+    value === null ||
+    !isNaN(value as number) ||
+    (typeof value === 'object' && Object.keys(value).length === 0) ||
+    (typeof value === 'string' && value.trim().length === 0)
+  );
+}
+
 export default class RedisQueryService implements QueryService {
+  connectionOptions(sourceOptions: SourceOptions) {
+    const _connectionOptions = (sourceOptions.connection_options || []).filter((o) => {
+      return o.some((e) => !isEmpty(e));
+    });
+
+    const connectionOptions = Object.fromEntries(_connectionOptions);
+    Object.keys(connectionOptions).forEach((key) =>
+      connectionOptions[key] === '' ? delete connectionOptions[key] : {}
+    );
+
+    return connectionOptions;
+  }
+
   async run(sourceOptions: SourceOptions, queryOptions: QueryOptions, dataSourceId: string): Promise<QueryResult> {
     let result = {};
     const query = queryOptions.query;
@@ -36,8 +59,19 @@ export default class RedisQueryService implements QueryService {
     const host = sourceOptions.host;
     const password = sourceOptions.password;
     const port = sourceOptions.port;
+    let tls = undefined;
 
-    const client = new Redis(port, host, { maxRetriesPerRequest: 1, username, password });
-    return client;
+    if (sourceOptions.tls) {
+      tls = {
+        ...this.connectionOptions(sourceOptions),
+      };
+    }
+
+    return new Redis(port, host, {
+      maxRetriesPerRequest: 1,
+      username,
+      password,
+      tls: tls,
+    });
   }
 }
--- a/plugins/packages/redis/lib/index.ts
+++ b/plugins/packages/redis/lib/index.ts
@@ -25,7 +25,12 @@ export default class RedisQueryService implements QueryService {
 
   async testConnection(sourceOptions: SourceOptions): Promise<ConnectionTestResult> {
     const client = await this.getConnection(sourceOptions);
-    await client.ping();
+    try {
+      await client.ping();
+    } catch (err) {
+      client.disconnect();
+      throw new QueryError('Connection could not be established', err.message, {});
+    }
 
     return {
       status: 'ok',
--- a/plugins/packages/redis/lib/index.ts
+++ b/plugins/packages/redis/lib/index.ts
@@ -1,6 +1,7 @@
 import { ConnectionTestResult, isEmpty, QueryError, QueryResult, QueryService } from '@tooljet-plugins/common';
 import Redis from 'ioredis';
 import { SourceOptions, QueryOptions } from './types';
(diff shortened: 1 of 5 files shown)


plugins/packages/redis/lib/manifest.json
```diff
@@ -65,6 +71,17 @@
...
       "type": "password",
       "description": "Enter password"
+    },
+    "tls": {
+      "label": "TLS",
+      "key": "tls",
+      "type": "toggle",
+      "description": "Toggle for TLS"
+    },
+    "connection_options": {
```
akshaysasidrn, line 81: Since it is only added for tls option, can we rethink this part? Connection options are generally spread at the root level. One approach is just prefix it with `tls_*`. 

But instead of that, I think we can do it like how we've done it for other datasources. Please have a look at the manifest file for Postgres datasource for reference.
We can have dropdown for the type of tls selection with `None` and `CA certificate` as an option. That will solve it, isn't it?

akshaysasidrn (changes requested):

plugins/packages/redis/lib/manifest.json
xwellingtonx: @akshaysasidrn I have followed the postgress plugin, the TLS options for ioredis do not support certificates, at least if you are not using the Redis.Cluster, which it's the case here

plugins/packages/redis/lib/index.ts
```diff
@@ -2,7 +2,30 @@ import { ConnectionTestResult, QueryError, QueryResult, QueryService } from '@to
 import Redis from 'ioredis';
 import { SourceOptions, QueryOptions } from './types';
 
+function isEmpty(value: number | null | undefined | string) {
```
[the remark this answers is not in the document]
xwellingtonx: @akshaysasidrn fixed

plugins/packages/redis/lib/manifest.json
xwellingtonx: Could you give me some example of what you mean by prefixing?

xwellingtonx: I'm not quite sure what you mean here, the Connection Options on postgress plugin is shown all the time, it does not control if it will show or not the Connection options.

xwellingtonx: I have changed it, but I don't understand why I need to add the common fields under the tls_certificate section, it should be outside of this section, but if I let outside it doesn't render the UI.


https://github.com/user-attachments/assets/402005c6-a484-49f3-ba84-63188fbf90bd

xwellingtonx: @akshaysasidrn I've understood now. But we should use the connection options inside the TLS section, because we can or not inform HOST/PORT and Family, those parameters are not mandatory but can me informed if needed

akshaysasidrn: @xwellingtonx Is that something that blocks you at the moment? If not, we can add it later iteratively.

xwellingtonx: @akshaysasidrn No, it's not! I can remove it so, I just added in case someone else would need after the deployment. I'm going to remove it since I need this be deployed so it unblock my APP

akshaysasidrn: Cool lets go ahead then 🙌🏾

xwellingtonx: @akshaysasidrn  all done.

akshaysasidrn: @xwellingtonx thanks, will have this tested and then take it forward for a release.

akshaysasidrn:
@xwellingtonx We are testing it out and found an issue with ioredis package. When the TLS connection fails, it keeps on retrying to connect to redis infinitely.
I tried it against a [mock redis setup](https://github.com/akshaysasidrn/TjMocks/tree/main/redis) for these config to verify. We'd need to fix this before merging. Let me look into this.


redis

xwellingtonx:
> @xwellingtonx We are testing it out and found an issue with ioredis package. When the TLS connection fails, it keeps on retrying to connect to redis infinitely. I tried it against a [mock redis setup](https://github.com/akshaysasidrn/TjMocks/tree/main/redis) for these config to verify. We'd need to fix this before merging. Let me look into this.
> 
> redis [ redis ](https://private-user-images.githubusercontent.com/11629675/357418278-0f4f172e-fada-4bf1-9261-234c7d37b2a5.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM1NTU2MzYsIm5iZiI6MTcyMzU1NTMzNiwicGF0aCI6Ii8xMTYyOTY3NS8zNTc0MTgyNzgtMGY0ZjE3MmUtZmFkYS00YmYxLTkyNjEtMjM0YzdkMzdiMmE1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODEzVDEzMjIxNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTEwNWVjMzRiYTRmMmI3YmVlNTVhMzljOWMzYWUxNzEwNTc2NGQxNDljZDcyZjliOGVlNGFkNWNmYTkzYjIyNDAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.Yll7BGsLUEo1DShjgD8k5sVIF4kAu-suVoG_UQ0ydis) [ ](https://private-user-images.githubusercontent.com/11629675/357418278-0f4f172e-fada-4bf1-9261-234c7d37b2a5.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM1NTU2MzYsIm5iZiI6MTcyMzU1NTMzNiwicGF0aCI6Ii8xMTYyOTY3NS8zNTc0MTgyNzgtMGY0ZjE3MmUtZmFkYS00YmYxLTkyNjEtMjM0YzdkMzdiMmE1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTMlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODEzVDEzMjIxNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTEwNWVjMzRiYTRmMmI3YmVlNTVhMzljOWMzYWUxNzEwNTc2NGQxNDljZDcyZjliOGVlNGFkNWNmYTkzYjIyNDAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.Yll7BGsLUEo1DShjgD8k5sVIF4kAu-suVoG_UQ0ydis)

Hi @akshaysasidrn I noted this even when not using SSL, so I don't think is related to my changes, most likely it's due to how TestConnection function is called by Tooljet.

plugins/packages/redis/lib/manifest.json
```diff
@@ -39,32 +51,98 @@
...
+          "value": "ca_certificate",
+          "name": "CA certificate"
+        },
+        {
+          "value": "self_signed",
+          "name": "Self-signed certificate"
+        },
+        {
+          "value": "none",
+          "name": "None"
```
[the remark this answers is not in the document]
xwellingtonx: @akshaysasidrn this is done now

akshaysasidrn:
> Hi @akshaysasidrn I noted this even when not using SSL, so I don't think is related to my changes, most likely it's due to how TestConnection function is called by Tooljet.

@xwellingtonx Thanks for pointing it out to me. So we need to terminate the client if it fails to establish the connection. That is missed currently. Can you also add this change in? Something like:

```js
 async testConnection(sourceOptions: SourceOptions): Promise<ConnectionTestResult> {
    const client = await this.getConnection(sourceOptions);

    try {
      await client.ping();
    } catch (err) {
      client.disconnect();
      throw new QueryError('Connection could not be established', err.message, {});
    }

    return {
      status: 'ok',
    };
  }
  ```

xwellingtonx:
> ```js
> try {
>       await client.ping();
>     } catch (err) {
>       client.disconnect();
>       throw new QueryError('Connection could not be established', err.message, {});
>     }
> ```

@akshaysasidrn this is now done.

akshaysasidrn (approved):

xwellingtonx:
Hi @akshaysasidrn , any update of the testing outcome on this?

akshaysasidrn:
Hey @xwellingtonx , it looks good. We are planning to release this at the earliest for LTS.

Issue #10291, opened 2024-07-08
Hover on one table triggers tooltip on all nearby tables


amanregu:
### Environment

ToolJet Cloud

### What is the expected behaviour?

When hovering over a single table in the canvas, only a tooltip associated with that specific table should be displayed.

### What is the current behaviour?

When hovering over the action elements (Add new row, Download, Manage Columns) on a single table in the canvas, the intended tooltip for that table appears alongside unexpected tooltips from adjacent(one cell or less than one cell, upto 4 cells incase of Manage Columns) tables. **This does NOT affect the functionality as the button click works as expected**.

### Screenshots or Screencast

https://github.com/ToolJet/ToolJet/assets/48810579/4892536d-9445-485f-b770-bceab0fb2220

PriteshKiri:
Hey @maxdiplogit ,

Thank you so much for your contribution! 🙌

Please keep an eye on your PR, as we may add some review comments if needed before merging it.

PR #10707, opened 2024-08-30
fix: hover on one table now does not trigger tooltips on all nearby t…

maxdiplogit:
…ables

Fixes: #10291

This PR fixes the hover problem that was triggered when any of the button was hovered over. Now only the tooltip for the button hovered will be displayed and not for other tables. This was made possible by checking whether the button being hovered is for that particular table or not.

[Screencast from 08-30-2024 01:30:34 PM.webm](https://github.com/user-attachments/assets/f9236851-77f9-4cd6-a92d-7f7238b9af57)

Happy to make any further changes if required :smile:


PriteshKiri:
Hey @maxdiplogit

Apologies for the delay in the response.

Your PR is under review now.

johnsoncherian (approved):

johnsoncherian (approved):

PR #10765, opened 2024-09-17
Updated ToolJet image to EE-LTS-latest

Slancerk:


.do/deploy.template.yaml
```diff
@@ -2,11 +2,11 @@ spec:
   name: tooljet
   services:
   - name: tooljet
-    image:
+    image: 
       registry_type: DOCKER_HUB
       registry: tooljet
       repository: tooljet
-      tag: latest
+      tag: tooljet/tooljet:EE-LTS-latest
```
adishM98, line 9: only tag is enough "EE-LTS-latest"

adishM98 (approved):

russell-hutson-jp:
I think the tag should be `ee-lts-latest` as Docker Hub tags are case-sensitive.

PR #10800, opened 2024-09-21
Fix workplace login whitespace 

### LINKED ISSUE #10674: White space at the bottom in Workspace login screen
### Version Information
2.50.5.5.19

When scrolling to the bottom, there should not be white space visible. 

White space is visible when scrolling down. 

1. Log in to tooljet and go to workspace settings
2. Workspace login tab and scroll to the bottom

### Screenshots or Screencast
Screenshot 2024-08-26 134124

### Any possible solutions?
Fix the size work space login section and enable scroll to go the bottom in that section. Keep the buttons are the bottom i.e. `SAVE` and `CANCEL` visible at all times, only the list should be scrollable

YES

IWang20:
Closes #10674

- Whitespace is no longer present when adjusting window height

- Main fixes - 
1. added scroll to the right hand side of settings for small window widths
2. added scroll to login card body
3. adjusted card height to match other workspace setting cards

- Additional bugfixes - 
1. CSS typo in .card
2.  unnecessary background would spill beyond login card if window width was too small (see below)
settings bug


frontend/src/_styles/theme.scss
```diff
@@ -5396,7 +5396,7 @@ div#driver-page-overlay {
 .workspace-settings-page {
   width: 880px;
   margin: 0 auto;
-  background: var(--base);
+  // background: var(--base);
```
gsmithun4, line 5399: Shall we remove the commented line?

IWang20: Yes, that would be fine. Do you want to make the change?

priteshTJ: Yes @IWang20

PriteshKiri:
Hey @IWang20,

Are you still working on the issue?

PR #10823, opened 2024-09-25
Pending Backlog Issues

rohanlahori:
Fixes : 

Issues : 

- [x] 12
- [x] 16
- [x] 21
- [x] 23
- [x] 39
- [x] re render folder count on app search

Ref : https://docs.google.com/spreadsheets/d/17vZggZMwrJwt4yH5ISMNMHxVCSKMj4m0WmiQ63MA5xo/edit?gid=874123979#gid=874123979

Moved EE Merge Issues fixes to CE: 

- [x] 4 Able to see embedded app link before app release 
- [x] 2 While clicking back arrow page is naviagting back to users tab but groups tab is been selected 
- [x] 7 In create workspace modal, Need to trim the white space from both end.

**Major change : 
 Regex for email field has been modified to not include any numeric or special characters in the end**

### DIFF
--- a/server/src/services/folders.service.ts
+++ b/server/src/services/folders.service.ts
@@ -107,7 +107,7 @@ export class FoldersService {
         .innerJoin('folderApp.app', 'app', 'folderApp.folderId = :id', {
           id: folder.id,
         })
-        .where('app.name LIKE :name', { name: `%${searchKey}%` })
+        .where('LOWER(app.name) LIKE :name', { name: `%${searchKey && searchKey.toLowerCase()}%` })
         .getMany();
 
       const userPermission = await this.abilityService.resourceActionsPermission(user, {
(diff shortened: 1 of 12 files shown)


server/src/services/folders.service.ts
```diff
@@ -107,7 +107,7 @@ export class FoldersService {
         .innerJoin('folderApp.app', 'app', 'folderApp.folderId = :id', {
           id: folder.id,
         })
-        .where('app.name LIKE :name', { name: `%${searchKey}%` })
+        .where('LOWER(app.name) LIKE :name', { name: `%${searchKey && searchKey.toLowerCase()}%` })
```
shah21, line 110:
```suggestion
        .where('LOWER(app.name) LIKE :name', { name: `%${(searchKey ?? '').toLowerCase()}%` })
```

shah21 (changes requested):

server/src/services/folders.service.ts
rohanlahori: fixed this

Issue #10762, opened 2024-09-16
[TextInput] Issues with setting width


DevangAgarwal93:
There is an issue with the width control of the text input component. 

### Steps to reproduce - 
1. Add two input components and try to change the width of any one of them
2. On unchecking the autowidth the width gets changed, ideally it should remain same (as default is 33) and then user can change
3. Now if we change the width to some % (80%) and when you move to the other text component, the default will be the width of previous component (80%) ideally it should be 33% only.
4. Now when we uncheck the autowidth the width gets changed again and not to 80% which is showing in UI. 
5. Now if we slightly change the width and make it back to whatever it is (80%) it will show a different width in UI

### To Check 
Does the issue exist on Number Input or any other component as well.

### Screen Recording
https://github.com/user-attachments/assets/bbb26519-4606-4da4-852d-29b37930ddcf
### Slack 
https://tooljet.slack.com/archives/C02QT13RF51/p1726151270233719

theashu02:
Hey! @DevangAgarwal93 
I would like to work on this bug, could you please assign me this issue.

neehhaa06:
Hello @DevangAgarwal93 :)

fix: #10762

```
To fix the width control issue in the `TextInput` component, start by ensuring that when `autowidth` is unchecked, the width is set to the default (33%) and only changes when the user manually adjusts it. Each input should maintain its own independent width state to prevent the width from persisting between multiple inputs. Additionally, the UI should always reflect the actual width value in the component's state, ensuring consistent behavior after manual width changes. This will resolve the problem of inconsistent width and carryover between different input fields.
```

anasKhafaga:
Hello @DevangAgarwal93 ,
I have proposed a solution for this issue in this PR #10838 . I'd appreciate if you take a look.

anasKhafaga:
@PriteshKiri ,
Are you still interested in solving this issue?

PriteshKiri:
Hey @anasKhafaga , our internal team will handle this issue.  

For issues open for community, please check out open issues under the `community contribution`

anasKhafaga:
@PriteshKiri ,
BTW, there's a PR submitted for review that solves it, and you can let the team check it out to save their time for other private issues.

Let me ask a question to save our time in the future: Is the Tooljet team not accept any contribution to an issue without that mentioned label?

PriteshKiri:
Hey @anasKhafaga,

[...]

PR #10838, opened 2024-09-30
fix(#10762): unchanged width value between components switch

anasKhafaga:
related to #10762 

https://github.com/user-attachments/assets/54c23142-9d75-4d51-8644-48f28233bbdd


PriteshKiri:
Hey @anasKhafaga,

We really appreciate your contribution! 

Just a gentle reminder to please start working on an issue only after it has been officially confirmed by the team, as there might be similar fixes in the pipeline.

Thank you for your understanding!

PR #10843, opened 2024-09-30
[docs]:  Update Password and Text Input Doc According to the Style Guide

RudraPratik30:
Updated the password input and text input docs according to the style guide.


karan-rathod-316 (approved):
Looks good to me.

PR #11036, opened 2024-10-21
[docs]: Update Create Query and Query Panel docs

RudraPratik30:
Updates the following two docs according to the new query panel in ToolJet 3.0:
- /docs/beta/app-builder/walkthrough/create-queries
- /docs/beta/app-builder/query-panel

I have updated this only in the next version.


karan-rathod-316 (approved):
Looks good to me.

PR #11100, opened 2024-10-24
[docs]: Update documentation for Page Management

RudraPratik30:
Updated the documentation for page management according to the ToolJet 3.0 updates.

Added new settings panel in the page management which can used to customize the page navigation sidebar on the app.


karan-rathod-316 (approved):
Looks good to me.

PR #11117, opened 2024-10-25
Fix table going into infinite render when placed inside form

nakulnagargade:


frontend/src/AppBuilder/Widgets/Table/Table.jsx
```diff
@@ -974,12 +968,6 @@ export const Table = React.memo(
       });
     }, [JSON.stringify(tableDetails.selectedRowsDetails)]);
 
-    const pageData = page.map((row) => row.original);
```
nakulnagargade, line 977: Removed this ussEffect(duplicate) as this is already being done in another UseEffect.

johnsoncherian (approved):

PR #11344, opened 2024-11-15
Local data source discontinued changes

rohanlahori:
**Description:** 
With the release of v3.0, local data sources have been officially discontinued. 
This PR introduces an update inside the App Builder Query panel interface to notify users with apps currently utilizing local data sources about this change and inform them to update to global data source.

[Design File : 
](https://www.figma.com/design/idCYgis6JK6jRI97i8IXa2/Data-source?node-id=5731-11422&node-type=frame&m=dev)

**Note**: Although the default data sources inside the App Builder are technically local data sources, they will continue to function as expected and should work fine without any issues.

Screenshot 2024-11-18 at 6 53 32 PM


frontend/src/AppBuilder/QueryManager/Components/QueryManagerBody.jsx
```diff
@@ -310,14 +312,28 @@ export const QueryManagerBody = ({ darkMode, options, setOptions, activeTab }) =
       </>
     );
   };
+  useEffect(() => {
+    const showDeprecationBanner =
+      selectedDataSource == null &&
+      selectedQuery &&
+      selectedDataSource?.kind !== 'runjs' &&
```
shah21, line 319:
```suggestion
      const staticDataSources = ['runjs', 'runpy', 'tooljetdb'];
      // use this check instead
      !staticDatasources.includes(selectedDataSource?.kind) &&
```

frontend/src/AppBuilder/QueryManager/Components/QueryManagerBody.jsx
```diff
@@ -330,7 +346,14 @@ export const QueryManagerBody = ({ darkMode, options, setOptions, activeTab }) =
       }} // 40px for preview header height
     >
       {selectedDataSource === null || !selectedQuery ? (
-        renderDataSourcesList()
+        showLocalDataDeprecationBanner ? (
+          <>
+            <NotificationBanner />
```
shah21, line 351: Does support dark mode styles automatically? If yes, please add the dark mode variable inside the component itself or pass value from the referenced files.

shah21 (changes requested):
@rohanlahori I have added some comments. Please take a look.

rohanlahori (commented):
fixed

shah21 (approved):

Issue #11256, opened 2024-11-11
App actions pop-up breaking near screen edge


nechalmaggon:
The pop-up on the app card near the edge of the screen opens up outside the screen area - it should open up aligning the bottom of the pop-up instead of the top in such cases to ensure the user can always view it 

<img width="1292" alt="Screenshot 2024-11-11 at 1 49 27 PM" src="https://github.com/user-attachments/assets/a3e37387-fe2e-40c5-97b5-f5e48fc97253">

kushalsourav:
@nechalmaggon  
Hi, I'd like to work on this issue if it's open and unassigned. Could you please assign it to me?
Thank you!

kushalsourav:
@nechalmaggon
Hi, while fixing the issue, I noticed that in Frontend -> src -> HomePage -> AppMenu.jsx, changing the OverlayTrigger component's placement from "bottom" to "top-start" prevents the popup from breaking the screen. However, resizing the screen still causes issues in some cases due to the limited width between the margin and the cards within the container.

As a solution, would it be ideal to dynamically increase the width and height of the card wrapper whenever the popup is displayed?

nechalmaggon:
Hey @kushalsourav, not super clear on what you mean by `dynamically increase the width and height of the card wrapper whenever the popup is displayed` would it be possible for you to share some screenshot and draw over it? 

In essence what I need is just for the pop-up to always be visible in the frame of the screen. Don't want the dashboard UI to change in any way.

kushalsourav:
@nechalmaggon 
App actions card popup that was breaking the screen at the bottom(fixed)
![image](https://github.com/user-attachments/assets/d50550fc-e23f-42fa-b9c0-1d7fc9b77dfc)

But the popup is also breaking the screen on the right side (screen size between 1100px - 992px)
![image](https://github.com/user-attachments/assets/4b1c6dac-626d-4896-8a9c-e126f9ee10ed)

And at screen size 700px below 
![image](https://github.com/user-attachments/assets/08daa130-4efc-40c2-a4fa-f0007aeb8fde)

I need clarity on the popup actions breaking the right side of the screen.
* I tried allowing the popup to be visible on the left side of the app bar
* ![image](https://github.com/user-attachments/assets/cc85da78-b840-4515-854e-8fc111590bb9)
*  The popup of the card also moves up and down as the user scrolls when the card popup is clicked.

[...]

PR #11381, opened 2024-11-21
Fix: popup breaking screen

kushalsourav:
fixed auto close popover when the user scrolls past the app card and didn't fix right side screen breaking 

Fix: #11256

### DIFF
--- a/frontend/src/HomePage/AppCard.jsx
+++ b/frontend/src/HomePage/AppCard.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, useEffect } from 'react';
+import React, { useState, useCallback, useEffect, useRef } from 'react';
 import cx from 'classnames';
 import { AppMenu } from './AppMenu';
 import moment from 'moment';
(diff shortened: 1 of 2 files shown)


frontend/src/HomePage/AppCard.jsx
```diff
@@ -90,7 +110,9 @@ export default function AppCard({
                   canUpdateApp={canUpdateApp(app)}
                   deleteApp={() => deleteApp(app)}
                   exportApp={() => exportApp(app)}
-                  isMenuOpen={isMenuOpen}
+                  isMenuOpen={setMenuOpen}
```
rudeUltra, line 113: Hey, @kushalsourav why was this changed to setMenuOpen

kushalsourav: Hey @rudeUltra  thank you for noticing this. I forgot to remove that line. The `isMenuOpen` prop is not defined as a prop in  `AppMenu.jsx`. I think we don’t need that line

* The issue was to fix the screen breaking at the bottom edge when the last row's app menu popup is visible
* I got an update on that issue: when the user clicks the button to open the pop-up, scrolling past the `App Card` linked to the pop-up closes it
* After fixing that, I realized the Home page has pagination, as users won’t be scrolling down when the screen width is at its maximum
* Popups auto close works only when screen-Y (Y-axis) is scrollable

kushalsourav:
@rudeUltra 

https://github.com/user-attachments/assets/7d79e486-b65d-438b-a35a-7298c7e434b1

Please let me know if there are any changes or suggestions

rudeUltra (approved):

PR #11407, opened 2024-11-25
[docs]: Add guide on How-to Display Table Record on new Page

RudraPratik30:
`/beta/how-to/display-table-record-on-new-page`

Added a how-to guide on displaying table records of selected row on a new page in the 3.1.0-Beta version.


karan-rathod-316 (changes requested):
@RudraPratik30 We don't need to run the query again on the new page. We need to store the entire data as the variable and then display it on the next page. Keep the current example, but also add one more example where everything is saved in the variable and then displayed in the new page.

RudraPratik30:
@karan-rathod-316 updated

karan-rathod-316 (approved):
Looks good to me.

karan-rathod-316 (changes requested):
Please make the required changes as discussed on call.

PR #11436, opened 2024-11-27
[docs]: Add Guide on How-to Use Custom Component

RudraPratik30:
- `/beta/how-to/use-custom-component`


karan-rathod-316 (changes requested):
This document looks a bit confusing to follow. Few suggestions from my side:

- Have complete examples where where you are passing data through the Data property and utilizing the same properties in the Code. Let's not have a separate section for Data as it is creating confusing. 
- Title to be changed to `Use Custom Components`

PR #11437, opened 2024-11-27
Fix: App Builder code cleanup

devanshu052000:
Closes: Move all widgets to Appbuilder folder and remove all dependancy on widgets from Editor folder. Also delete all widget related code from Editor folder


frontend/src/AppBuilder/WidgetManager/configs/widgetConfig.js
shaurya-sharma064, line 1: Remove kanbanBoardConfig, this will add a Kanban board component, but dropping it will cause the component to break.

frontend/src/AppBuilder/WidgetManager/configs/widgetConfig.js
```diff
@@ -53,6 +53,7 @@ import {
   linkConfig,
   iconConfig,
   boundedBoxConfig,
+  kanbanBoardConfig,
```
shaurya-sharma064: Remove kanbanBoardConfig, this will add a Kanban board component, but dropping it will cause the component to break.

shaurya-sharma064 (changes requested):

shaurya-sharma064 (dismissed):

Issue #4318, opened 2022-10-05
SQL Server with encryption fails as a data source


keskinonur:
### Environment
ToolJet 1.26.1 with Dorcke

### What is the expected behaviour?

Using SQL Server as a data source.

### What is the current behaviour?

SQL Server connection with encryption and possibly with trust server certification fails.

### How to reproduce the issue?

1. Create or edit an app inside ToolJet admin
2. Add or edit SQL Server database which encrypt its connection
3. Test connection or query anything like SELECT @@Version or else

### Screenshots or Screencast

https://imgur.com/QCAqIGo
https://imgur.com/JOd7IZl

### Any possible solutions?
Add an option inside .env config or in the GUI for enabling encryption and trusting server certification for SQL Server.

shubh22:
Hi @sudOnur ,

GUI option for this would be a good solution.
Would be willing to raise a PR for this?

pkoz:
I encountered the same problem, so I'd happily create a fix. 

Below is my suggestion for the solution:
<img width="755" alt="image" src="https://github.com/user-attachments/assets/7cbc5dfa-1100-4e22-9dc6-28912bb731a7">

Comments:
1. When a value of the connection options equals `true`, `false`, `yes`, or `no`, the value is translated to a boolean.
2. The label "Azure" of the toggle is not precise. From the code, it sends the option `encrypt: true`, which is expected not only for Azure but also for AWS or other providers.

Please let me know if that works.

pkoz:
@ganesh8056 I noted you pushed one commit from the PR, but there are two commits in there

PriteshKiri:
Closing as the issue no longer exists!

PR #11491, opened 2024-12-03
#4318 SQL Server with encryption fails

pkoz:
The angle of attack: some SQL Server providers require specific connection parameters; therefore, inspired by the PostgreSQL data source configurator, I've added key-value pairs that are further translated to parameters. 

If the parameter's value is `true`, `false`, `yes`, or `no`, then the type of the value is assumed to be a boolean. If the parameter's value can be parsed as an `integer,` the type is `number`.

### DIFF
--- a/plugins/packages/mssql/lib/index.ts
+++ b/plugins/packages/mssql/lib/index.ts
@@ -16,8 +16,8 @@ const recognizedBooleans = {
   no: false,
 };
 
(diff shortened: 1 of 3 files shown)


akshaysasidrn:
@pkoz Thank you for your contribution. We need to tweak a bit of change here and then have it released.
We will cherry-pick your commits and take it forward. Plan to release it in the LTS version first.

PR #11519, opened 2024-12-06
Fixed query transformations getting swapped while changing queries

shaurya-sharma064:


shaurya-sharma064:
Already merged in other PR

PR #11564, opened 2024-12-09
Bug fixes

shaurya-sharma064:
Fixed following issues:

1. Duplicating page causes components with parent and reference to not be duplicated.
2. Event changes don't occur synchronously.
3. Form when disabled doesn't disable all the components.
4. Fixed page reloading on setting page events.
5. Two handle range slider not taking array as default value.
6. HTML support added.
7. Fixed hiding show download button hides column selector button aswell


frontend/src/Editor/Components/Table/GenerateEachCellValue.jsx
```diff
@@ -117,7 +117,7 @@ export default function GenerateEachCellValue({
     }
   }
   useEffect(() => {
-    if (mounted && _.isEmpty(rowChangeSet)) {
+    if (_.isEmpty(rowChangeSet)) {
```
TooljetBot, line 120: <div>


<div id="suggestion">
<div id="issue"><b>Consider impact of removing mounted check</b></div>
<div id="fix">

Consider if removing the `mounted` check from the dependency array could affect component lifecycle handling. The `useEffect` hook might now trigger in scenarios where the component is not fully mounted.
</div>
<details>
<summary>
<b>Code suggestion</b>
</summary>
<blockquote>Check the AI-generated fix before applying</blockquote>
<div id="code">


```suggestion
    if (mounted && _.isEmpty(rowChangeSet)) {
```

</div>
</details>
</div>


<small><i>Code Review Run <a href=https://github.com/ToolJet/ToolJet/pull/11564#issuecomment-2538244716>#7c809d</a></i></small>
</div>

---

johnsoncherian (approved):

Issue #11739, opened 2025-01-03
Add cypress test cases for group permissions


YuktiGoyal02:
### Specify the kind of test

Cypress integration-test

### Describe the test
Privileges for Builder and Admin
- Add query in the app
- Export and import  the app

PR #11740, opened 2025-01-03
added query and import/export privileges

YuktiGoyal02:
Github Issue - #11739


ajith-k-v (approved):
Always add reusable functions to the `utils` file.

PR #11788, opened 2025-01-11
Updated private and public apps test cypress cases

ajith-k-v:


cypress-tests/cypress/commands/apiCommands.js
```diff
@@ -168,6 +168,8 @@ Cypress.Commands.add("apiCreateWorkspace", (workspaceName, workspaceSlug) => {
       expect(response.status).to.equal(201);
     });
   });
+  cy.logoutApi();
```
emidhun, line 171: Can we skip this part, and can we set the env from the response?

emidhun (commented):
@ajith-k-v Can we follow the naming conventions for the api commands?
eg: `apiLogin`, `api<commandName>`
I think it will give us more readability on specs.

cypress-tests/cypress/commands/apiCommands.js
ajith-k-v: Will address this comments in the upcoming PR's

gsmithun4 (approved):

PR #11822, opened 2025-01-17
[docs]: Update ToolJet API

PriteshKiri:
- Added User Role
- Updated the Update user Dto
- Removed Metadata APIs


karan-rathod-316 (approved):
Looks good to me.

Issue #11880, opened 2025-01-29
Add cypress test cases for workspace constants


YuktiGoyal02:
Specify the kind of test
Cypress integration-test

Describe the test
Add test cases for:

Global and secret constants in components, inspector and data source
Global and secrets in query preview, preview
Global and secrets in datasource form

PR #11884, opened 2025-01-29
added cases for workspace constants

YuktiGoyal02:
#11880


ajith-k-v (approved):
@YuktiGoyal02 From the next iteration onwards, use API commands for CRUD operations if they are being used multiple times.

PR #11944, opened 2025-02-06
Fix: GraphQL plugin appends an empty URL parameter (?=) to the endpoint.

ganesh8056:
Problem:
In GraphQL plugin even when url parameters are not provided, empty parameters are appended in the endpoint ( ?= ).


ganesh8056:
Created a new PR : #12353

PR #11963, opened 2025-02-10
Feat/modal footer header 5

nithindavid:


frontend/src/AppBuilder/RightSideBar/Inspector/Components/ModalV2.jsx
```diff
@@ -0,0 +1,92 @@
...
+    apps,
+    allComponents,
+    validations,
+    darkMode,
+    [],
+    additionalActions
+  );
+
+  const [optionsItems] = conditionalAccordionItems(component);
+  accordionItems.splice(2, 0, optionsItems);
```
nakulnagargade, line 89: can we please replace this magic number 2 with constant variable?

frontend/src/_styles/widgets/modal.scss
```diff
@@ -0,0 +1,60 @@
+.modal-content.modal-component.tj-modal-widget-content {
```
nakulnagargade, line 1: I think this styles can go under modals directory as we already have created a file to keep it separate from older modal styles.

kavinvenkatachalam (commented):
@nithindavid Overall looks good to me. 
Can you add `React.memo` to the header and footer of the modal component? This will stop unwanted re-renders if something is changed on the parent

PR #11975, opened 2025-02-11
feat: Add support such that absolute size of a component is maintained while dropping or copy/pasting a component in and out of the container components

### LINKED ISSUE #11271: Absolute size of a component to be maintained while moving a component in and out of the container components
### Behaviour for dropping a new small component directly in a large container
When components are dropped directly on the containers from component library panel - the size of the component reduces compared to what it is when they are dropped on the canvas.

### Solution
The component should maintain its default size by rounding off the absolute size.

### Behaviour when moving a small component inside large container from the canvas - 
When the component is moved from the canvas to the container components mentioned above - the size remains intact. 

### Solution
This is expected behaviour.

## Copy Pasting Small components into the container and from the container
1. The size decreases when components is pasted into the container
2. The size increases when component is copied from the container and pasted on the canvas.

Size should remain intact when copy pasting from the container and into the container.

### Behaviour for dropping a new large component directly in a small container
1. The component gets placed accurately where the developer intends to place it.
2. However it resizes to fit inside the container from the right side.
3. The bottom boundary coincides with the bottom boundary of the container (a scroll gets added to the container to fit in the large component vertically).

### Behaviour when moving a large component inside small container from the canvas - 
1. The component resizes to fit horizontally (both from left and right side) inside the container and snaps to the left boundary of the container.
2. The bottom boundary coincides with the bottom boundary of the container (a scroll gets added to the container to fit in the large component vertically).

### Solution for both cases
1. Component should resize horizontally to fit inside the container. 
2. 

> To check - should the component snap to the left boundary - this will allow the component to occupy maximum width, or should it get placed where the developer drops it which will allow the developer to accurately place the component, however, if the developer is placing a large component inside the container most likely they intend to resize it and placing it where they drop it would allow them extra padding to use the resize handle more easily.

 
3. Component should not resize vertically, since we intend to introduce dynamic height later which will increase the height of the container automatically to fit the component. 
4. Left boundary of the component coincides with the left boundary of the container and the right boundary with the right boundary.

Another Related Issue - [[Container Bug] Difficult to resize large components inside container](https://github.com/ToolJet/ToolJet/issues/10645)

nakulnagargade:
Fixes #11271

### DIFF
--- a/frontend/src/AppBuilder/AppCanvas/appCanvasUtils.js
+++ b/frontend/src/AppBuilder/AppCanvas/appCanvasUtils.js
@@ -5,6 +5,7 @@ import useStore from '@/AppBuilder/_stores/store';
 import { toast } from 'react-hot-toast';
 import { CANVAS_WIDTHS, NO_OF_GRIDS, WIDGETS_WITH_DEFAULT_CHILDREN } from './appCanvasConstants';
 import _ from 'lodash';
+import { useGridStore } from '@/_stores/gridStore';
 
 export function snapToGrid(canvasWidth, x, y) {
   const gridX = canvasWidth / 43;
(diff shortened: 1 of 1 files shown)


frontend/src/AppBuilder/AppCanvas/appCanvasUtils.js
```diff
@@ -409,6 +413,16 @@ export function pasteComponents(parentId, copiedComponentObj) {
     componentData.definition.others.showOnDesktop.value = currentLayout === 'desktop' ? `{{true}}` : `{{false}}`;
     componentData.definition.others.showOnMobile.value = currentLayout === 'mobile' ? `{{true}}` : `{{false}}`;
 
+    // Adjust width if parent changed
+    let width = component.layouts.desktop.width;
```
johnsoncherian, line 417: add null checks just incase we get a corrupted data

johnsoncherian (dismissed):

PR #12033, opened 2025-02-24
[docs]: Update Devices Section

vj-codes:
Changed devices description to a tabular format which is consistent with other component docs


RudraPratik30 (approved):
Looks good to me.

PR #12102, opened 2025-03-04
[docs]: Update TJ API and Map Const w/ Env Var

RudraPratik30:
1. Added Get All App ID in ToolJet API Docs
2. Added Mapping Workspace Constants to Environment Variables


karan-rathod-316 (dismissed):
Looks good to me.

gsmithun4 (approved):

PR #12143, opened 2025-03-06
Fixed transformation issue causing query to not retain multiple trans…

shaurya-sharma064:
…formation types


johnsoncherian:
reopened in https://github.com/ToolJet/ToolJet/pulls?q=head%3Afeat%2Ftransformation-fallback

PR #12164, opened 2025-03-08
[docs]: Update Datasources - Baserow and Twilio

akshatvirmani:


akshatvirmani:
Updated Twilio and Baserow

- Updates Images
- Updated Example Responses
- Added Example Values

PR #12186, opened 2025-03-10
Git Sync backend fixes

rohanlahori:


server/src/modules/app-git/module.ts
```diff
@@ -0,0 +1,94 @@
...
+    const importPath = await getImportPath(configs?.IS_GET_CONTEXT);
+    const { AppGitController } = await import(`${importPath}/app-git/controller`);
+    const { AppGitService } = await import(`${importPath}/app-git/service`);
+    const { AppGitUtilService } = await import(`${importPath}/app-git/util.service`);
+    const { GitSyncUtilService } = await import(`${importPath}/git-sync/util.service`);
+    const { TooljetDbImportExportService } = await import(
+      `${importPath}/tooljet-db/services/tooljet-db-import-export.service`
+    );
+    const { ImportExportResourcesService } = await import(`${importPath}/import-export-resources/service`);
+    const { AppsService } = await import(`${importPath}/apps/service`);
```
AnantshreeChandola, line 33: Try to avoid importing direct service files of other modules, can move the implementation to utility files and import those instead.

server/src/modules/app-git/module.ts
```diff
@@ -0,0 +1,94 @@
...
+    const { VersionService } = await import(`${importPath}/versions/service`);
+    const { AppImportExportService } = await import(`${importPath}/apps/services/app-import-export.service`);
+    const { TooljetDbTableOperationsService } = await import(
+      `${importPath}/tooljet-db/services/tooljet-db-table-operations.service`
+    );
+    const { AppEnvironmentUtilService } = await import(`${importPath}/app-environments/util.service`);
```
AnantshreeChandola, line 39: Most util services are exported from their modules, can import the module itself. No need to add util services in providers.

PR #12248, opened 2025-03-15
fix: replace unpkg with jsdelivr

SukkaW:
UNPKG has not been actively maintained and has been down since `Mar 15, 2025` (https://github.com/unpkg/unpkg/issues/412).

The PR replaces UNPKG usage with jsDelivr.

Since UNPKG usage has gone, we can safely remove UNPKG's references in the server's CSP directives as well.


SukkaW:
UNPKG was down again from `Mar 31, 2025, 7:00 AM UTC` and lasted 8 hours (https://github.com/unpkg/unpkg/issues/443), I urge everyone to immediately migrate away from unpkg to jsDelivr now.

Issue #12039, opened 2025-02-24
Add cypress spec for Baserow data sources


Srimanitejas123:
### Specify the kind of test

Smoke test suite

Issue #12009, opened 2025-02-18
Added cypress test spec for airtable


Srimanitejas123:
### Specify the kind of test
 Smoke test suit for Airtable

### Describe the test

Added cypress test spec for airtable

PR #12263, opened 2025-03-17
cypress spec for airtable and baserow

Srimanitejas123:
#12039 , #12009


cypress-tests/cypress/constants/selectors/Plugins.js
Mekhla-Asopa, line 1: There is no need to add "baserow" as a prefix in the selector key because it is present inside "baserowSelectors"

cypress-tests/cypress/e2e/happyPath/marketplace/commonTestcases/data-source/aitTable.cy.js
Mekhla-Asopa, line 1: Please check file name, it should be airTable

Mekhla-Asopa (approved):

PR #12308, opened 2025-03-20
Docs: Adding PostgREST as a mandatory installation

adishM98:
- PostgREST needs to be mandatory installation since it is required for ToolJetDB


karan-rathod-316 (approved):
Looks good to me.

PR #12325, opened 2025-03-21
Added feature to add bulk upsert using PK

### LINKED ISSUE #12209: TJ DB - [New Operation] bulk upsert using primary key
### Problem
Currently, users cannot perform bulk insert and update together in TJ DB. 

Also in current self hosted beta version, this can be done with ToolJet DB SQL editor, but again complex queries and transformations have to be written for the same. 

### Solution 
We need to build a new GUI operation `Bulk upsert using Primary Key" in TJ DB query builder. 
Upon selecting this operation, Primary key of the selected table will be shown and another input having the array of objects(change set) of rows to be updated. Note that each object in this array need to specify primary key column(s) with their values. 

Note: For inserting new rows, PK column(s) will only be required if PK is not of serial type.

Step 1: Select the table and this operation
Step 2: Input the changeset with updated row values

**Single Primary Key**

**Composite Primary Key**

manishkushare:
Resolves #12209


akshaysasidrn (changes requested):

akshaysasidrn (approved):

ganesh8056 (changes requested):

server/src/modules/tooljet-db/services/tooljet-db-bulk-upload.service.ts
```diff
@@ -368,4 +370,195 @@ export class TooljetDbBulkUploadService {
...
+        rows: [],
+      };
+    }
+
+    const result = await this.tableOperationsService.perform(organizationId, 'view_table', {
+      id: tableId,
+    });
+    const tableColumns = result?.columns || [];
+
+    const invalidPrimaryKeys = primaryKeyColumns.filter((pk) => !tableColumns.some((col) => col.column_name === pk));
```
ganesh8056, line 439: In the check for invalid Primary key columns, there is validation only for a column exists in table, and a check for if column is also Primary key is missed.

server/src/modules/tooljet-db/services/tooljet-db-bulk-upload.service.ts
```diff
@@ -368,4 +370,195 @@ export class TooljetDbBulkUploadService {
...
+        rows: [],
+      };
+    }
+
+    const serialTypeColumns = tableColumns
+      .filter((col) => col.data_type === 'integer' && /^nextval\(/.test(col.column_default))
+      .map((col) => col.column_name);
+
+    // Identify serial vs. non-serial primary keys
+    const serialPrimaryKeys = primaryKeyColumns.filter((pk) => serialTypeColumns.includes(pk));
```
ganesh8056, line 455:
```
    const serialPrimaryKeys = primaryKeyColumns.filter((pk) => serialTypeColumns.includes(pk));
    const nonSerialPrimaryKeys = primaryKeyColumns.filter((pk) => !serialTypeColumns.includes(pk));
```
can we make it to single loop like
```
const serialPrimaryKeys = []
const nonSerialPrimaryKeys = []
primaryKeyColumns.forEach((pk) => serialTypeColumns.includes(pk) ? serialPrimaryKeys.push(pk) ? nonSerialPrimaryKeys.push(pk));
```

server/src/modules/tooljet-db/services/tooljet-db-bulk-upload.service.ts
```diff
@@ -368,4 +370,195 @@ export class TooljetDbBulkUploadService {
...
+    const nonSerialPrimaryKeys = primaryKeyColumns.filter((pk) => !serialTypeColumns.includes(pk));
+
+    // Group rows by the exact set of provided keys
+    const rowGroups = new Map<string, Record<string, any>[]>();
+    for (const row of rowsToUpsert) {
+      // Ensure required non-serial PKs are provided
+      const missingNonSerialPKs = nonSerialPrimaryKeys.filter((pk) => row[pk] === undefined);
+      if (missingNonSerialPKs.length > 0) {
```
ganesh8056, line 463: While we throw error for missing PK values. We are not providing the information on which record it is.

manishkushare: @ganesh8056 
Screenshot 2025-04-11 at 12 41 06 PM

ganesh8056 (dismissed):

akshaysasidrn (approved):

PR #12338, opened 2025-03-21
[docs]: Update GitSync

vj-codes:


docs/static/img/gitsync/gitsync-v4.png
PriteshKiri, line 1: Can you blur the workspace name at the bottom left?

PR #12345, opened 2025-03-24
Fix: Correction in AI whitelisting message

adishM98:
- Correction: Azure Container App and EC2 deployment for AI whitelisting message.


karan-rathod-316 (approved):
Looks good to me.

Issue #11798, opened 2025-01-15
Dark mode - tree select widget poor visibility


abulka:
### Version Information

Version 3.0.24-cloud-lts

### Environment
ToolJet Cloud

### What is the expected behaviour?
Should see the treeview items properly.

### What is the current behaviour?

Treeview items barely visible in dark mode.

### How to reproduce the issue?

1. in dark mode, drag out tree select widget
2. notice how hard it is to read the text of the tree

### Screenshots or Screencast

<img width="416" alt="image" src="https://github.com/user-attachments/assets/896ed9fa-95a5-4845-af00-8fcf0943d8f0" />

neelpatel19:
Hey @PriteshKiri can i work on this , can you assign it to me !!

PriteshKiri:
Hey @neelpatel19,

Are you still interested in working on this issue?

GuntreddyHemanth:
Hey @PriteshKiri can I work on this, please assign it to me !

PriteshKiri:
@GuntreddyHemanth  assigned!!

Please make sure that,
 - You sign the CLA after submitting your PR. 
 - You complete this task in 5 days from the day of the assignment.   
 
Thank you for showing interest in contributing to ToolJet! Feel free to reach out to us in our [Slack community](https://join.slack.com/t/tooljet/shared_invite/zt-2rk4w42t0-ZV_KJcWU9VL1BBEjnSHLCA) if you have any questions.

GuntreddyHemanth:
PR raised @PriteshKiri

PriteshKiri:
Hey @GuntreddyHemanth!

Can you please check the comment made on your PR?

ramakrishna67:
@PriteshKiri is the issue still open?

PriteshKiri:
@ramakrishna67 This issue is already assigned. Could you please check other issues with the **community contribution** label?

PriteshKiri:
@ramakrishna67 are you still interested in working on this issue?

ramakrishna67:
Hello @PriteshKiri.
Thank you for reaching me. 
I would like to work on it but consider the pr raised by @Dante-Nephilim as he already worked on it.

PriteshKiri:
@Dante-Nephilim you can raise the PR for the same.

@ramakrishna67 feel free to check out other issues with `community contribution`

Dante-Nephilim:
@PriteshKiri, PR is raised

[...]

PR #12369, opened 2025-03-25
fix `TreeSelect` colors in dark mode

Dante-Nephilim:
Changes:
- remove hard-coded `color: #3e525b;`. The `textColor` is already set in `TreeSelect.jsx` according to current theme.
- replace  hard-coded url svg `content` with html unicode character (Black Medium Right-Pointing Triangle Centred) https://www.compart.com/en/unicode/U+2BC8
- update rotate in expanded state according to the new unicode character

Screenshots:

Dark

Light


Dante-Nephilim:
fixes #11798

PR #12384, opened 2025-03-26
Added detailed error descriptions in BigQuery Plugin

### LINKED ISSUE #10999: BigQuery - Add detailed error descriptions + link documentation in query builder
Currently we are not adding detailed error descriptions directly from the respective SQL servers when a query fails.

We need a standard way of showing errors. There errors from all 3 of SQL plugins are expected to return in the following format

```
{
    status:"",
    message:"",
    description:"",
    data:{} //// The exact error object sent by the data source with specifc details of error
}

```

The data attribute needs to return the exact error sent from the respective SQL Servers. Their documentation can be found on the following link https://cloud.google.com/bigquery/docs/error-messages

thesynthax:
Resolves #10999

### DIFF
--- a/plugins/packages/bigquery/lib/index.ts
+++ b/plugins/packages/bigquery/lib/index.ts
@@ -104,7 +104,37 @@ export default class Bigquery implements QueryService {
       }
     } catch (error) {
       console.log(error);
-      throw new QueryError('Query could not be completed', error.message, {});
+      const errorMessage = error.message || "An unknown error occurred.";
+      let errorDetails: any = {};
+      
+      const errorSuggestions = {
+        "notFound": "Check if the table or dataset exists in the specified location.",
+        "accessDenied": "Verify that the service account has the necessary permissions.",
+        "invalidQuery": "Check the SQL syntax and ensure that all referenced columns exist.",
+        "rateLimitExceeded": "You are making too many requests. Try again after some time.",
+        "backendError": "BigQuery encountered an internal error. Retry the request after some time.",
+        "quotaExceeded": "You have exceeded your quota limits. Consider upgrading your plan or reducing query size.",
+        "duplicate": "A resource with this name already exists. Try using a different name.",
+        "badRequest": "Check the request parameters and ensure they are correctly formatted.",
+      };
+
+      if (error && error instanceof Error) {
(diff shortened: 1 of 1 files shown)


plugins/packages/bigquery/lib/index.ts
```diff
@@ -104,7 +104,37 @@ export default class Bigquery implements QueryService {
       }
     } catch (error) {
       console.log(error);
-      throw new QueryError('Query could not be completed', error.message, {});
+      const errorMessage = error.message || "An unknown error occurred.";
+      let errorDetails: any = {};
+      
+      const errorSuggestions = {
+        "notFound": "Check if the table or dataset exists in the specified location.",
```
akshaysasidrn, line 111: @thesynthax the suggestions you have provided correspoding to the error reason is from their documentation?

plugins/packages/bigquery/lib/index.ts
```diff
@@ -104,7 +104,37 @@ export default class Bigquery implements QueryService {
...
+
+        const reason = bigqueryError.response?.status?.errorResult?.reason || "unknownError";
+        errorDetails.reason = reason;
+        errorDetails.message = errorMessage;
+        errorDetails.jobId = bigqueryError.response?.jobReference?.jobId;
+        errorDetails.location = bigqueryError.response?.jobReference?.location;
+        errorDetails.query = bigqueryError.response?.configuration?.query?.query;
+
+
+        const suggestion = errorSuggestions[reason] || "Check your request and try again.";
```
akshaysasidrn, line 133: The fallback message here might mislead the user for errors on the data source side. We can let it be empty.

akshaysasidrn (changes requested):

plugins/packages/bigquery/lib/index.ts
thesynthax: Yes. 
https://cloud.google.com/bigquery/docs/error-messages#errortable

plugins/packages/bigquery/lib/index.ts
thesynthax: Okay got it

akshaysasidrn (approved):

PR #12398, opened 2025-03-27
add metadata to QueryError and populating metadata in inspector on error

### LINKED ISSUE #12310: [REST API] Send metadata when query fails
Problem: When a REST API query runs successfully, its request headers, response and response headers gets added to  `metadata` key, these are used further in other queries to query secured resources like` metadata.responseHeaders.accessToken`
But when the query fails, metadata does not gets poulated hence causing issues in the app flow. 

Solution: 
Both query success and query failure should have same response structure with details in `metadata`

thesynthax:
resolves #12310

### DIFF
--- a/plugins/packages/common/lib/query.error.ts
+++ b/plugins/packages/common/lib/query.error.ts
@@ -1,8 +1,8 @@
 export class QueryError extends Error {
   data: Record<string, unknown>;
   description: any;
-  metadata?: any;
-  constructor(message: string | undefined, description: any, data: Record<string, unknown>, metadata?: any) {
+  metadata?: unknown;
+  constructor(message: string | undefined, description: string, data: Record<string, unknown>, metadata?: unknown) {
     super(message);
     this.name = this.constructor.name;
     this.data = data;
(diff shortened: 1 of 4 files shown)


plugins/packages/common/lib/query.error.ts
```diff
@@ -1,11 +1,13 @@
 export class QueryError extends Error {
   data: Record<string, unknown>;
   description: any;
-  constructor(message: string | undefined, description: any, data: Record<string, unknown>) {
+  metadata?: any;
```
akshaysasidrn, line 4: Prefer `unknown` over `any`. Let's change the description to `string`, if there are type issues then make it `unknown`.

plugins/packages/restapi/lib/index.ts
```diff
@@ -217,24 +218,35 @@ export default class RestapiQueryService implements QueryService {
...
+          statusCode: error.response.statusCode,
+          responseBody: error.response.body,
+        }
+
+        metadata = {
+          request: requestObject,
+          response: responseObject,
+        };
+
+        result = {
```
akshaysasidrn, line 238: Can you add a comment above stating:
```js
// TODO: Need to remove the request/response related information in result in next MAJOR release. 
// This is now shared in `metadata` key. Keeping this here for backward compatibility.
```

akshaysasidrn (changes requested):

PR #12440, opened 2025-04-02
Fix: App menu popup breaking screen

devanshu-gupta2002:
Fix for: #11256 
Fixed App Menu popup opening out of view area by auto closing popup when the user scrolls past app menu viewing area.

Fix SR:

https://github.com/user-attachments/assets/8310be41-73f3-4a28-b390-c4231aa66f0b

### DIFF
--- a/frontend/src/HomePage/AppMenu.jsx
+++ b/frontend/src/HomePage/AppMenu.jsx
@@ -47,6 +47,7 @@ export const AppMenu = function AppMenu({
       onToggle={onMenuOpen}
       onExit={() => setMenuOpen(false)}
       show={popoverVisible}
+      container={document.getElementsByClassName('home-page-content')[0]}
       overlay={
         <div>
           <Popover id="popover-app-menu" className={darkMode && 'dark-theme'} placement="bottom">
--- a/frontend/src/HomePage/AppMenu.jsx
+++ b/frontend/src/HomePage/AppMenu.jsx
@@ -13,6 +13,8 @@ export const AppMenu = function AppMenu({
   openAppActionModal,
   darkMode,
   currentFolder,
+  popoverVisible,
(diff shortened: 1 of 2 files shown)


frontend/src/HomePage/AppMenu.jsx
```diff
@@ -40,9 +42,11 @@ export const AppMenu = function AppMenu({
   return (
     <OverlayTrigger
       trigger="click"
-      placement="bottom-end"
+      placement="top-start"
       rootClose
       onToggle={onMenuOpen}
+      onExit={() => setMenuOpen(false)}
+      show={popoverVisible}
```
kavinvenkatachalam, line 49:
```suggestion
      container={document.getElementsByClassName('home-page-content')[0]}
```

Adding the container might fix the issue

kavinvenkatachalam (changes requested):
@devanshu-gupta2002 Can you try adding the `container` and polacement to `top-start`? If it works then rest of the changes are not needed

frontend/src/HomePage/AppMenu.jsx
devanshu-gupta2002: should I set the container placement to `top start` as default or calculate if the container is going out of view and then only set the placement to `top-start`

devanshu-gupta2002: Adding the container did fix the issue but only keeping the `container` logic leads to the app menu following the scroll and the menu does not close automatically.
I have attached both views in the below recording, if you want i can either keep the previous changes or revert it back.

https://github.com/user-attachments/assets/199831bd-1388-42fc-8fc0-87d87642b2bd

shah21 (approved):

kavinvenkatachalam (approved):

PR #12501, opened 2025-04-08
add metadata to QueryError and populating metadata in inspector on error

### LINKED ISSUE #12310: [REST API] Send metadata when query fails

thesynthax:
resolves #12310

### DIFF
--- a/plugins/packages/restapi/lib/index.ts
+++ b/plugins/packages/restapi/lib/index.ts
@@ -228,7 +228,7 @@ export default class RestapiQueryService implements QueryService {
         responseObject = {
           statusCode: error.response.statusCode,
           responseBody: error.response.body,
-          headers: error.response.headers,
+          headers: redactHeaders(error.response.headers),
         }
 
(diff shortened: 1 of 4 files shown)


plugins/packages/restapi/lib/index.ts
```diff
@@ -228,6 +228,7 @@ export default class RestapiQueryService implements QueryService {
         responseObject = {
           statusCode: error.response.statusCode,
           responseBody: error.response.body,
+          headers: error.response.headers,
```
ganesh8056, line 231: Can we clean the sensitive information from headers. There is existing method ( **redactHeaders** )

PR #12534, opened 2025-04-09
Fix: restricted input of whitespaces in new password input in reset p…

devanshu-gupta2002:
Fix for: [Issue](https://github.com/orgs/ToolJet/projects/28/views/1?sliceBy%5Bvalue%5D=_noValue&filterQuery=epic%3A%22Post+Modularisation+Fixes%22+assignee%3A&groupedBy%5BcolumnId%5D=Assignees&pane=issue&itemId=104473641)

### DIFF
--- a/frontend/src/SettingsPage/SettingsPage.jsx
+++ b/frontend/src/SettingsPage/SettingsPage.jsx
@@ -87,12 +87,12 @@ function SettingsPage(props) {
   };
 
   const handlePasswordInput = (input) => {
-    const sanitizedInput = input.replace(/\s/g, '');
-    setNewPassword(sanitizedInput);
-    if (sanitizedInput.length > 100) {
+    setNewPassword(input);
+    const trimmedInput = input.trim();
+    if (trimmedInput.length > 100) {
       setHelperText('Password should be Max 100 characters');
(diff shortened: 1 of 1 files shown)


frontend/src/SettingsPage/SettingsPage.jsx
```diff
@@ -87,11 +87,12 @@ function SettingsPage(props) {
   };
 
   const handlePasswordInput = (input) => {
-    setNewPassword(input);
-    if (input.length > 100) {
+    const sanitizedInput = input.replace(/\s/g, '');
```
shah21, line 90: Even though adding whitespace between words is not a common practice when choosing passwords, we shouldn't remove all whitespace. Instead, we should only remove spaces from the beginning and end.

So shall we do this instead ?
`const sanitizedInput = input.trim();
`

shah21 (changes requested):

frontend/src/SettingsPage/SettingsPage.jsx
devanshu-gupta2002: sure I'll change it to trim

shah21 (approved):

PR #12559, opened 2025-04-11
Fix/remove metadata ce

devanshu-gupta2002:
Fix for: [Issue](https://github.com/orgs/ToolJet/projects/28/views/1?sliceBy%5Bvalue%5D=_noValue&filterQuery=epic%3A%22Post+Modularisation+Fixes%22+assignee%3A&groupedBy%5BcolumnId%5D=Assignees&pane=issue&itemId=104437693)

- Removed Metadata column, server response for CE
- Added bulk user upload, organization-users in EE submodule with Metadata
- Added new sample_upload_ce.csv for CE

### DIFF
--- a/frontend/src/modules/WorkspaceSettings/pages/Users/InviteUsersForm.jsx
+++ b/frontend/src/modules/WorkspaceSettings/pages/Users/InviteUsersForm.jsx
@@ -13,6 +13,7 @@ import { EDIT_ROLE_MESSAGE } from '@/modules/common/constants';
 import ModalBase from '@/_ui/Modal';
 import { UserMetadata } from './components';
 import LicenseBanner from '@/modules/common/components/LicenseBanner';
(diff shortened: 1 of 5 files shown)


frontend/src/modules/WorkspaceSettings/pages/Users/InviteUsersForm.jsx
```diff
@@ -68,6 +69,7 @@ function InviteUsersForm({
 
   const { super_admin } = authenticationService.currentSessionValue;
   const [featureAccess, setFeatureAccess] = useState({});
+  const edition = fetchEdition();
```
shah21, line 72: As per the modularization standards, using the edition inside components is not recommended. If there are any edition-specific changes, they should be controlled using props and base components.

@devanshu-gupta2002, please go through the other modularized components once and revert this shortly. Let’s keep it this way for now.

shah21 (approved):

PR #12578, opened 2025-04-14
[post-modularisation] Fixed: Archiving user from the personal workspace on all users page

shah21:


server/src/modules/organization-users/controller.ts
```diff
@@ -78,8 +78,8 @@ export class OrganizationUsersController implements IOrganizationUsersController
 
   @InitFeature(FEATURE_KEY.USER_ARCHIVE)
   @Post(':id/archive')
-  async archive(@User() user: UserEntity, @Param('id') id: string) {
-    const organizationId = user.organizationId;
+  async archive(@User() user: UserEntity, @Param('id') id: string, @Body() body) {
+    const organizationId = body.organizationId ? body.organizationId : user.organizationId;
```
gsmithun4, line 82: user.organizationId will be available all cases right?

shah21: Yes, technically this can cause some bugs if we don’t pass an ID from the front end. However, the areas where we are using this are the workspace settings page and the instance settings page. Currently, we are passing the ID from the instance users page, so it should be fine.

Issue #12584, opened 2025-04-15
RESTAPI headers sending invalid values on resolved Js embedds


akshaysasidrn:
### Version Information
v3.5.27-ee-lts and v3.9.0

### Environment
ToolJet Cloud / Self-hosted (Docker/Linux/MacOS)

### What is the expected behaviour?
The REST API data source should handle empty values (undefined, null, NaN, empty strings) consistently, with either:
1. An option to omit parameters with empty values, or
2. Convert JavaScript values to their stringified versions

### What is the current behaviour?
Currently, the REST API data source:
- Only skips parameters when the key is empty
- Throws errors when values are null or NaN
- Sends undefined values as "undefined" (or skips them)
- Sends empty strings as-is

This inconsistent behavior causes issues when working with dynamically resolved JavaScript expressions.

### How to reproduce the issue?
1. Create a REST API data source
2. Add parameters with different "empty" values (undefined, null, NaN, empty string)
3. Try to execute the query
4. Observe errors when null or NaN values are present

### Screenshots or Screencast

![Image](https://github.com/user-attachments/assets/60dd4b37-7efb-4b3d-a492-e4b20b5fba21)

![Image](https://github.com/user-attachments/assets/2ade3e4b-165a-425c-8b35-19ee24682be2)

![Image](https://github.com/user-attachments/assets/91723e61-89dd-46a1-aa0d-6e00290a88a7)

![Image](https://github.com/user-attachments/assets/fad4c40f-aef8-47a9-8a72-9649c59f2e52)

### Any possible solutions?
1. Implement a consistent approach to handle all empty/special values:
   - Add a checkbox option "Omit empty values" that, when enabled, would skip parameters with undefined, null, NaN, or empty string values
   - When disabled, convert these values to their string representations ("undefined", "null", "NaN", "")

2. Standardize the behavior across all ToolJet versions to make API requests more predictable.

### Can you identify the location in the driver source code where the problem exists?
The issue is likely in the REST API data source handler where request parameters are processed before sending.

### If the bug is confirmed, would you be willing to submit a PR?
Yes, I can help implement a solution for this issue if the approach is agreed upon.

PR #12602, opened 2025-04-16
Hotfix: Sanitize invalid headers

parthy007:
- This fixes for RestAPI & Graphql
Closes: [#12584](https://github.com/ToolJet/ToolJet/issues/12584)


plugins/packages/common/lib/utils.helper.ts
```diff
@@ -83,14 +83,23 @@ export const sanitizeHeaders = (
   hasDataSource = true
 ): { [k: string]: string } => {
   const cleanHeaders = (headers) => headers.filter(([k, _]) => k !== '').map(([k, v]) => [k.trim(), v]);
+  const filterValidHeaderEntries = (headers) => {
```
akshaysasidrn, line 86: This handles specific case of `undefined` and `null` resolved. I recommend adding type validation to filter out other invalid types like `NaN`, function, objects and arrays. Please put some thought into it, also check from the query manager how it is currently handled and how got library expects types of headers to be in. Looking at their type definition will help.

akshaysasidrn (changes requested):

plugins/packages/common/lib/utils.helper.ts
akshaysasidrn: Prompted solution for ideation, please verify:

```js
/**
 * Demonstrates header sanitization with detailed console logs
 * Shows how different JavaScript types are handled in HTTP headers
 */
const demonstrateHeaderSanitization = () => {
  // Sample header entries with various JavaScript types
  const sampleHeaders = [
    ['Content-Type', 'application/json'],        // Valid: string
    ['X-Number', 42],                           // Valid: number
    ['X-Boolean', true],                        // Valid: boolean
    ['X-Null', null],                           // Invalid: null
    ['X-Undefined', undefined],                 // Invalid: undefined
    ['X-NaN', NaN],                             // Invalid: NaN
    ['X-Infinity', Infinity],                   // Invalid: Infinity
    ['X-NegativeInfinity', -Infinity],          // Invalid: -Infinity
    ['X-Object', { key: 'value' }],             // Invalid: object
    ['X-Array', [1, 2, 3]],                     // Invalid: array
    ['X-Function', () => console.log('test')],  // Invalid: function
    ['X-Symbol', Symbol('sym')],                // Invalid: symbol
    ['', 'Empty key'],                          // Invalid: empty key
    ['  X-Trim-Me  ', 'needs trimming'],        // Valid after trimming
    ['X-Empty-String', '']                      // Valid: empty string
  ];

  console.log('=== SANITIZATION DEMONSTRATION ===');
  console.log('Original headers:', sampleHeaders);

  // Step 1: Clean header keys (remove empty keys and trim)
  const cleanKeys = sampleHeaders
    .filter(([key, _]) => key !== '')
    .map(([key, value]) => [key.trim(), value]);
  
  console.log('\n=== AFTER CLEANING KEYS ===');
  console.log(cleanKeys);

  // Step 2: Filter valid types (only string, number, boolean)
  // Explicitly exclude null, undefined, NaN, Infinity
  const validTypes = cleanKeys.filter(([_, value]) => {
    // Check if value is undefined or null
    if (value === undefined || value === null) {
      return false;
    }
    
    // Get the type
    const type = typeof value;
    
    // Check if it's a valid primitive type
    const isValidType = type === 'string' || type === 'number' || type === 'boolean';
    
    // Special check for NaN and Infinity
    const isInvalidNumber = type === 'number' && (isNaN(value) || !isFinite(value));
    
    return isValidType && !isInvalidNumber;
  });
  
  console.log('\n=== AFTER FILTERING INVALID TYPES ===');
  console.log(validTypes);

  // Step 3: Convert all values to strings (required for HTTP headers)
  const stringified = validTypes.map(([key, value]) => [
    key, 
    typeof value === 'string' ? value : String(value)
  ]);
  
  console.log('\n=== AFTER CONVERTING TO STRINGS ===');
  console.log(stringified);

  // Final result as an object
  const finalHeaders = Object.fromEntries(stringified);
  console.log('\n=== FINAL HEADERS OBJECT ===');
  console.log(finalHeaders);
  
  return finalHeaders;
};

// Run the demonstration
demonstrateHeaderSanitization();
```

PR #12611, opened 2025-04-16
Fix/smtp creds env

devanshu-gupta2002:
Fix for: [Issue](https://docs.google.com/spreadsheets/d/1jav_EV2Cv8Qs7dU0tVLlPIlqsCgbIaCa9IJE-DegUhs/edit?gid=0#gid=0&range=41:41)


server/src/modules/smtp/module.ts
```diff
@@ -2,18 +2,20 @@ import { getImportPath } from '@modules/app/constants';
...
     const importPath = await getImportPath(configs?.IS_GET_CONTEXT);
     const { SMTPService } = await import(`${importPath}/smtp/service`);
     const { SmtpController } = await import(`${importPath}/smtp/controller`);
+    const { SMTPUtilService } = await import(`${importPath}/smtp/util.service`)
     return {
       module: SMTPModule,
       imports: [await InstanceSettingsModule.register(configs)],
       controllers: [SmtpController],
-      providers: [SMTPService, FeatureAbilityFactory],
+      providers: [SMTPService, FeatureAbilityFactory, SMTPUtilService, InstanceSettingsService],
```
gsmithun4, line 18: Rules to follow.
Never use other module's service as provider for another. use only util service and import respective module.
should remove `InstanceSettingsService` from here

gsmithun4 (changes requested):

PR #12615, opened 2025-04-16
fixed plugin duplication bug

### LINKED ISSUE #12462: Add uniqueness contraint on plugins to prevent plugin to be installed multiple times in a workspace
A plugin should only be installed once in a workspace

A plugin can be installed multiple times in a workspace

1. Login with valid credential 
2. Navigate to the marketplace plugin installation page
3. Open the same page on a different tab such that there are two tabs of the same page
4. Click on install on each of the tabs for the same plugin
5. Goto the data source configuration page and check the plugins installed to find the duplicates

### Screenshots or Screencast

Image

thesynthax:
resolves #12462

### DIFF
--- a/server/src/entities/plugin.entity.ts
+++ b/server/src/entities/plugin.entity.ts
@@ -6,9 +6,11 @@ import {
   OneToOne,
   PrimaryGeneratedColumn,
   UpdateDateColumn,
+  Unique
 } from 'typeorm';
 import { File } from 'src/entities/file.entity';
 
+@Unique(['pluginId'])
 @Entity({ name: 'plugins' })
 export class Plugin {
   @PrimaryGeneratedColumn()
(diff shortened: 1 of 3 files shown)


server/src/modules/plugins/service.ts
```diff
@@ -21,6 +21,12 @@ export class PluginsService implements IPluginsService {
 
   async install(body: CreatePluginDto) {
     const { id, repo } = body;
+    
+    const existingPlugin = await dbTransactionWrap((manager: EntityManager) => {
+      return manager.findOne(Plugin, { where: { pluginId: id } });
```
akshaysasidrn, line 26: Can we also add a uniqueness constraint at the DB level to avoid duplicates on requests with race conditions?

akshaysasidrn (changes requested):

server/src/entities/plugin.entity.ts
```diff
@@ -6,9 +6,11 @@ import {
   OneToOne,
   PrimaryGeneratedColumn,
   UpdateDateColumn,
+  Unique
 } from 'typeorm';
 import { File } from 'src/entities/file.entity';
 
+@Unique(['pluginId'])
```
akshaysasidrn, line 13: We cannot make use of this as it works with TypeORM synchronisation feature. We write custom migrations and will have to create one for this.

Check the /migrations directory

akshaysasidrn (changes requested):

PR #12627, opened 2025-04-20
External API Modularisation 

rudeUltra:


server/src/modules/external-apis/dto/external_apis.dto.ts
```diff
@@ -12,6 +12,7 @@ import {
   IsNotEmpty,
 } from 'class-validator';
 import { Type } from 'class-transformer';
```
gsmithun4, line 14: Rename
index.ts

PR #12657, opened 2025-04-21
Fixes for static data sources and build with permissions

AnantshreeChandola:


server/src/modules/app/guards/ability.guard.ts
```diff
@@ -94,11 +94,7 @@ export abstract class AbilityGuard implements CanActivate {
 
       // Validate all features against resource if any
       if (!features.every((feature) => ability.can(feature, this.getSubjectType(), resourceId || undefined))) {
-        throw new ForbiddenException(
-          JSON.stringify({
-            organizationId: app?.organizationId,
-          })
-        );
+        throw new ForbiddenException(`You do not have permissions to perform this action.`);
```
gsmithun4, line 97: This changes breaks the app redirection flow

gsmithun4 (changes requested):

PR #12671, opened 2025-04-22
Fix CE Edition Licensing Terms

rudeUltra:


server/src/modules/licensing/configs/LicenseBase.ts
```diff
@@ -42,8 +41,17 @@ export default class LicenseBase {
...
   private _isExternalApis: boolean;
   private _isAppWhiteLabelling: boolean;
+  private BASIC_PLAN_TERMS: Partial<Terms>;
+
+  constructor(
+    licenseData?: Partial<Terms>,
+    updatedDate?: Date,
+    startDate?: Date,
+    expiryDate?: Date,
+    BASIC_PLAN_TERMS?: Partial<Terms>
```
gsmithun4, line 51: shall we make the BASIC_PLAN_TERMS first argument?

gsmithun4 (changes requested):

PR #12744, opened 2025-04-30
feat: Audit logs 2.0 Batch-1

devanshu-gupta2002:
Change Log:
- Added `resource_data` column in `audit_logs` table to save addition properties
- Updated existing user actions with additional properties
- Added audit logs for new user actions.

Batch-0 requested changes: [Sheet](https://docs.google.com/spreadsheets/d/1_ZywNyhInRSUj5VgzBeF2BO2pnj2bmxl13UrgqjAAvQ/edit?gid=529434162#gid=529434162)


server/src/modules/profile/util.service.ts
```diff
@@ -5,10 +5,28 @@ import { FilesRepository } from '@modules/files/repository';
...
+import { AUDIT_LOGS_REQUEST_CONTEXT_KEY } from '@modules/app/constants';
+import { User } from '@entities/user.entity';
 
 @Injectable()
 export class ProfileUtilService implements IProfileUtilService {
-  constructor(protected readonly filesRepository: FilesRepository, protected userRepository: UserRepository) {}
+  constructor(protected readonly filesRepository: FilesRepository, protected readonly userRepository: UserRepository) {}
+
+  private async setAuditLogForUser(userId: string, manager: EntityManager): Promise<void> {
+    const user = await manager.findOneOrFail(User, {
```
shah21, line 18: You can pass user dto from controller to service functions. if it it applies for other events, please follow that instead of querying it one more time

server/src/modules/session/util.service.ts
```diff
@@ -44,6 +45,21 @@ export class SessionUtilService {
     protected readonly encryptionService: EncryptionService,
     protected readonly jwtService: JwtService
   ) {}
+
+  async setAuditLogForUser(userId: string, manager: EntityManager): Promise<void> {
```
shah21, line 49: If there are no additional session-related audit log events, you can call it directly from the service function. Alternatively, if the function in users.util.server.ts shares the same logic, please use a common utility function across multiple services.

shah21 (commented):
I have added some comments. please take a look @devanshu-gupta2002

server/src/modules/session/util.service.ts
devanshu-gupta2002: Currently the function shares the same logic for the most part because of default properties. But we plan on having additional properties for all the actions so sharing the same function may seem a bit difficult.
I'll try to reuse the function wherever possible.

server/src/modules/profile/util.service.ts
devanshu-gupta2002: `manager.update`  only returns the affected rows and not the updated user so i have to query the user regardless because the dto does not contain sufficient data for audit logs

PR #12788, opened 2025-05-02
Prevent Stringify Query Resolved Value

rudeUltra:


server/test/modules/data-queries/util.service.spec.ts
```diff
@@ -454,13 +455,13 @@ describe('DataQueriesUtilService', () => {
 
       expect(result).toEqual({
         string: 'test string',
-        number: 42,
-        boolean: true,
+        number: '42',
```
akshaysasidrn, line 458: Should resolve to the user given data type itself, not stringified data.
This is the problem we are trying to solve.

akshaysasidrn (changes requested):

PR #12802, opened 2025-05-05
added cypress spec for user invite with metadata

trishna-tooljet:


cypress-tests/cypress/constants/selectors/common.js
```diff
@@ -402,4 +406,20 @@ export const commonWidgetSelector = {
   appLinkField: '[data-cy="app-link-field"]',
   appSlugErrorLabel: '[data-cy="app-slug-error-label"]',
   appLinkSucessLabel: '[data-cy="app-link-success-label"]',
+  iconHidden: '[data-cy="icon-hidden"]',
+  inspectornodeglobals: '[data-cy="inspector-node-globals"]',
+  inspectorNodeCurrentUser: '[data-cy="inspector-node-currentuser"]',
+  inspectorNodeMetadata: '[data-cy="inspector-node-metadata"]',
+  defaultValueInputField: '[data-cy="default-value-input-field"]',
+  draggableWidgetTextInput: '[data-cy="draggable-widget-textinput1"]',
```
ajith-k-v, line 414: Please reuse the existing selectors instead of redefining them.

cypress-tests/cypress/e2e/happyPath/platform/eeTestcases/editmetadata.cy.js
```diff
@@ -0,0 +1,43 @@
+
```
ajith-k-v, line 1: Consider including all CRUD operations in a single spec file instead of creating separate specs for each.

cypress-tests/cypress/support/utils/manageUsers.js
```diff
@@ -401,3 +401,37 @@ export const inviteUserWithUserRole = (firstName, email, role) => {
   cy.get(commonSelectors.acceptInviteButton).click();
   cy.get(commonSelectors.homePageLogo, { timeout: 10000 }).should("be.visible");
 };
+
+export const inviteUserWithUserRoleAndMetadata = (firstName, email, role) => {
```
ajith-k-v, line 405: Please reuse the existing function instead of creating a new one. If minor adjustments are needed for this use case, consider adding parameters to the existing function to handle the variation.

ajith-k-v (changes requested):

ajith-k-v (changes requested):
@trishna-tooljet  Please update the changes as per the review comments. Always make sure to create the PR without including the `package-lock.json` and sub module files.

PR #12803, opened 2025-05-05
Feat/audit logs user

devanshu-gupta2002:
[Issue](https://github.com/ToolJet/tj-ee/issues/913)
Change Log:
- Added `resource_data` column in `audit_logs` table to save addition properties
- Updated existing user actions with additional properties
- Added audit logs for new user actions.
- Added `/user/instance` route 

Batch-1 requested changes: [Sheet](https://docs.google.com/spreadsheets/d/1_ZywNyhInRSUj5VgzBeF2BO2pnj2bmxl13UrgqjAAvQ/edit?gid=529434162#gid=529434162)


server/migrations/1637909008629-CreateAuditLogs.ts
```diff
@@ -43,6 +43,11 @@ export class CreateAuditLogs1637909008629 implements MigrationInterface {
             type: 'varchar',
             isNullable: true,
           },
+          {
```
gsmithun4, line 46: Should not add migrations like this, should create new migration file and add column

server/src/modules/organization-users/util.service.ts
```diff
@@ -30,8 +30,8 @@ import { UserDetailsService } from './services/user-details.service';
 import { FetchUserResponse, InvitedUserType, RoleUpdate, UserFilterOptions } from './types';
 import { GroupPermissionsRepository } from '@modules/group-permissions/repository';
 import { ERROR_HANDLER, ERROR_HANDLER_TITLE } from '@modules/organizations/constants';
-import { MODULE_INFO } from '@modules/app/constants/module-info';
-import { MODULES } from '@modules/app/constants/modules';
+// import { MODULE_INFO } from '@modules/app/constants/module-info';
```
gsmithun4, line 33: remove these lines if not used

gsmithun4 (changes requested):

server/src/modules/profile/util.service.ts
```diff
@@ -5,10 +5,25 @@ import { FilesRepository } from '@modules/files/repository';
...
+import { RequestContext } from '@modules/request-context/service';
+import { AUDIT_LOGS_REQUEST_CONTEXT_KEY } from '@modules/app/constants';
+import { User } from '@entities/user.entity';
 
 @Injectable()
 export class ProfileUtilService implements IProfileUtilService {
-  constructor(protected readonly filesRepository: FilesRepository, protected userRepository: UserRepository) {}
+  constructor(protected readonly filesRepository: FilesRepository, protected readonly userRepository: UserRepository) {}
+
+  private async setAuditLogForUser(user: User, resourceData?: any): Promise<void> {
```
gsmithun4, line 17: try to add audit logs call on services, if the function on util service is used by any other service, it will write audit logs

server/src/modules/profile/util.service.ts
```diff
@@ -31,6 +46,51 @@ export class ProfileUtilService implements IProfileUtilService {
...
+      },
+      updated_user_details: {
+        first_name: firstName,
+        last_name: lastName,
+      },
+    };
+    await this.setAuditLogForUser(user, resourceData);
+  }
+
+  async updateUserPassword(userId: string, password: string, manager: EntityManager): Promise<void> {
```
gsmithun4, line 82: same here

server/src/modules/session/service.ts
```diff
@@ -34,6 +34,7 @@ export class SessionService {
     response.clearCookie('tj_auth_token');
     await dbTransactionWrap(async (manager: EntityManager) => {
       await manager.delete(UserSessions, { id: sessionId, userId });
+      await this.sessionUtilService.setAuditLogForUser(userId, manager);
```
gsmithun4, line 37: avoid using functions for audit logs, call directly

gsmithun4 (changes requested):

gsmithun4 (approved):

PR #12820, opened 2025-05-07
[docs]: GitSync Feedback Update

RudraPratik30:
Working on some enhancements suggested by the PMs for the docs. 

Main Feature - GitSync New Flow, merged through #12794


RudraPratik30:
[Slack Thread](https://tooljet.slack.com/archives/C03A0P0SQAD/p1746629911736219?thread_ts=1745927339.200719&cid=C03A0P0SQAD)

Feedback -
- App migration - If you've set up a different ToolJet instance for each environment, git sync can be used to seamlessly move applications from one instance to another. You can pull an app from git repo into your testing and production environments.

- Import app is still coming abruptly in the doc. Don't try to force everything into sections. The how-to parts make sense in sections because they're long. But the first part of the doc needs to be more oriented towards the user's flow. Build a concept narrative first that covers the purpose/value of the entire doc. Then go into details.

- You've explained pull changes into an existing app already but a user will first have to pull an app into the workspace (import). Write another line in the intro section stating that to pull an application to another instance or workspace, you'll have to first import it from git and it'll get imported with latest commit. After that if any new commits are there, the user can just pull changes into the same app. When an app is imported from git, they can be imported in two states based on your use case - editable and non-editable. Then move to your use cases.

PR #12834, opened 2025-05-08
[improvement] Default workspace

shah21:
https://github.com/orgs/ToolJet/projects/28/views/10?pane=issue&itemId=103823470&issue=ToolJet%7Ctj-ee%7C3405

**Changes**
This PR includes changes related to file sharing between enterprise repo.

Major changes
1. DB and data-migration
2. Allow personal workspace related changes
3. Logic changes in common signup function
4. Prop changes for passing default-org 
5. Removed/Commented the ENABLE_ONBOARDING_QUESTIONS_FOR_ALL_SIGN_UPS env support from EE and CE

#### Major Developer test cases

1. **Super-admin Setup**
   * Super-admin's workspace should be set as the default workspace.
2. **Signup (Form and SSO)**
   * New users should be added to the default workspace.
   * Verified instance-level and workspace-level signups via both form and SSO.
3. **Login Flow**
   * Login should function as before.
     * Form login: No changes made.
     * SSO login is working as expected.
4. **Default Workspace Visibility for Super-admin**
   * Super-admin should be able to view and edit the default workspace from the "All Workspaces" page.
6. **Default Workspace Restrictions**
   * Archiving the default workspace should not be allowed.
   * A tooltip should be shown indicating this restriction.
7. **Existing Users Migration**
   * For existing users, the first created workspace should be assigned as the default workspace.
8. **Environment Variable-Based Workspace**
   * If a user is using `TOOLJET_DEFAULT_WORKSPACE_URL`, that workspace should become the default during migration.
9. Did some CE all-round testing since I touched the CE code. (Login,Signup,invite)


server/data-migrations/1740401100000-SetDefaultWorkspace.ts
```diff
@@ -0,0 +1,66 @@
...
+            UPDATE organizations 
+            SET is_default = true 
+            WHERE slug = $1
+          `, [workspaceSlug]);
+          return;
+          }
+          console.log(`No active organization found with slug: ${workspaceSlug}`);
+        }
+      } catch (err) {
+        console.log('Invalid TOOLJET_DEFAULT_WORKSPACE_URL format');
```
gsmithun4, line 37: NOTE: Migration will be success even if it throws error

server/src/modules/session/util.service.ts
```diff
@@ -368,8 +368,8 @@ export class SessionUtilService {
   async #onboardingFlags(user: User) {
     let isFirstUserOnboardingCompleted = true;
     let isOnboardingCompleted = true;
-    const isOnboardingQuestionsEnabled =
-      this.configService.get<string>('ENABLE_ONBOARDING_QUESTIONS_FOR_ALL_SIGN_UPS') === 'true';
+    // const isOnboardingQuestionsEnabled =
```
gsmithun4, line 371: we have to keep this, since we are using cloud same as EE behaviour

server/data-migrations/1740401100000-SetDefaultWorkspace.ts
shah21: I think this should be fine. If those conditions or queries don’t pass, the query below will update the default workspace.

server/src/modules/session/util.service.ts
shah21: Sure. we can take this up when LTS cloud is testing.

adishM98 (approved):

adishM98 (approved):

PR #12848, opened 2025-05-11
Fix/custom theme

shaurya-sharma064:


shaurya-sharma064:
Already fixed in some other PR

PR #3496, opened 2022-07-05
added component specific action to Dropdown Widget

manishkushare:
Resolve #3495

PR #12862, opened 2025-05-13
Fix: Ensure secrets and constants mapped to encrypted fields are correctly imported with the app

parthy007:
Closes: [#3496](https://github.com/ToolJet/tj-ee/issues/3496)
Sub module PR: [#71](https://github.com/ToolJet/ee-server/pull/71)


akshaysasidrn (approved):

plugins/packages/common/lib/utils.helper.ts
```diff
@@ -17,8 +18,29 @@ export function cacheConnection(dataSourceId: string, connection: any): any {
...
+  return crypto.createHash('sha256').update(sortedEntries).digest('hex').substring(0, 16);
+}
+
+export function cacheConnectionWithConfiguration(dataSourceId: string, enhancedCacheKey: string, connection: any): any {
+  const updatedAt = new Date();
+  const allKeys = Object.keys(CACHED_CONNECTIONS);
+  const oldKeysForThisDatasource = allKeys.filter(
+    (key) => key.startsWith(`${dataSourceId}_`) && key !== enhancedCacheKey
+  );
+  oldKeysForThisDatasource.forEach((oldKey) => delete CACHED_CONNECTIONS[oldKey]);
```
ganesh8056, line 37: Above Filter & Foreach can be combined in one iteration

server/src/helpers/in_memory_cache.service.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@nestjs/common';
```
ganesh8056: We would not need this Cache logic. Also solution is yet to be fixed

server/src/modules/data-sources/module.ts
```diff
@@ -13,6 +13,7 @@ import { TooljetDbModule } from '@modules/tooljet-db/module';
 import { OrganizationRepository } from '@modules/organizations/repository';
```
ganesh8056, line 12: InMemoryCacheService - is not required as it only works, when we upgrade from Basic to Team plan

server/src/modules/data-sources/util.service.ts
```diff
@@ -21,6 +21,7 @@ import { PluginsServiceSelector } from './services/plugin-selector.service';
 import { OrganizationConstantsUtilService } from '@modules/organization-constants/util.service';
 import { DataSourceOptions } from '@entities/data_source_options.entity';
 import { IDataSourcesUtilService } from './interfaces/IUtilService';
+import { InMemoryCacheService } from '@helpers/in_memory_cache.service';
```
ganesh8056: We can fallback to old logic. InMemoryCacheService - only works, when we upgrade from Basic to Team plan

ganesh8056 (dismissed):

PR #12905, opened 2025-05-20
Fix: Small height of single line code editor when expanded

devanshu052000:


devanshu052000:
Fixed in this PR -> [#12906](https://github.com/ToolJet/ToolJet/pull/12906)

PR #12921, opened 2025-05-21
Added Scroll and Background Styling to Table Header & Container

shaurya-sharma064:


shaurya-sharma064:
Already merged as part of another PR

PR #12959, opened 2025-05-30
Fix grid issues

nakulnagargade:
This fixes 

1. Shadow size of the component is not same as the component when dropping first time into container, and while moving between canvases/containers.
2. When moving a button into a container, it's body flickers
3. When moving a component between containers, the component shadow disappears.
4. Component group selection glitches on scroll in Container and Form.
5. When I drop a large component on canvas, a shadow appears towards the top left corner of the component.
6. When dropping a component, shadow is coming on top of component manager.


nakulnagargade:
This PR is already merged via a different PR

PR #12995, opened 2025-06-09
Adding Cloud plugin scripts

adishM98:


server/scripts/plugins-install.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  console.log('Plugins to install:', pluginsToInstall);
+
+  for (const pluginId of pluginsToInstall) {
+    const pluginDto = Object.assign(new CreatePluginDto(), findPluginDetails(pluginId));
+    const validationErrors = validateSync(pluginDto);
+
+    if (validationErrors.length === 0) {
+      const plugin = await entityManager.findOne(Plugin, { where: { pluginId: pluginDto.id } });
+      plugin ? invalidPluginDtos.push(pluginDto) : validPluginDtos.push(pluginDto);
```
ganesh8056, line 40: nit: If plugins are already exists, should we need to categorise as Existing plugin and not as invalid plugins

PR #13011, opened 2025-06-12
Fix/audit logs filter

devanshu-gupta2002:
Change Log:
- Added `/audit-logs/resources` route to fetch resources and actions


gsmithun4:
Duplicate

PR #13018, opened 2025-06-13
design: Adds License buttons to header

nithindavid:
image


frontend/src/_ui/Header/index.jsx
```diff
@@ -140,20 +140,27 @@ function Header({
...
               })}
               data-cy="version-label"
             >
+              <LicenseNavBarActions
+                licenseStatus="unlicensed"
+                trialDaysRemaining={10}
+                onStartTrial={() => {}}
+                onGetConsultation={() => {}}
+                onTrialAction={() => {}}
+              />
```
nithindavid: Entry point to add the license logic

nithindavid:
@nechalmaggon Here are the different cases

1. unlicensed
image

2. trial
image

3. trial-ending
image

4. trial-expired
image

PR #13127, opened 2025-06-30
Tabs Migration

shaurya-sharma064:
On upgrading version or importing older tabs component the children were missing as the useDynamicOptions wasn't turned on by default.


shaurya-sharma064:
Merging into another branch

PR #13129, opened 2025-07-01
Fix Container and form on import adding header and footer by default

nakulnagargade:


emidhun:
@nakulnagargade  Please handle versioning on a different PR.

PR #13131, opened 2025-07-01
Fix for app count guard

rohanlahori:


server/src/modules/licensing/guards/app.guard.ts
```diff
@@ -5,20 +5,28 @@ import { EntityManager } from 'typeorm';
...
 
   //Dependancy error using licenseCountsService need to check
-  async fetchTotalAppCount(manager: EntityManager): Promise<number> {
-    const apps = await manager.find(App, {
-      where: {
-        type: APP_TYPES.FRONT_END,
-        organization: {
-          status: 'active',
-        },
+  async fetchTotalAppCount(manager: EntityManager, organizationId: string): Promise<number> {
```
gsmithun4, line 16: if edition is cloud and organizationId is empty or equals ORGANIZATION_INSTANCE_KEY
throw error

PR #13143, opened 2025-07-01
fix module permission

platform-ops123:


server/src/modules/apps/ability/guard.ts
```diff
@@ -15,6 +15,10 @@ export class FeatureAbilityGuard extends AbilityGuard {
         return {
           resourceType: MODULES.APP,
         };
+      case APP_TYPES.FRONT_END || APP_TYPES.MODULE:
```
gsmithun4, line 18: case APP_TYPES.FRONT_END:
        return {
          resourceType: MODULES.APP,
        };
        
        
        why do we need to keep this?

gsmithun4:
Pushed changes

server/src/modules/apps/ability/guard.ts
platform-ops123: to handle module app type, there is no default case. It was returning null.

PR #13151, opened 2025-07-02
fix: add fallback for useModuleId hook in query panel

johnsoncherian:
Handle case where ModuleContext is not available by wrapping
useModuleId in try-catch with 'canvas' fallback


johnsoncherian:
This is closed in https://github.com/ToolJet/ToolJet/pull/13176

PR #13162, opened 2025-07-02
Feat: New fields 'client id' and 'client secret' have been introduced in the Slack datasource configuration page in pre-release

parthy007:


frontend/src/_components/Slack.jsx
```diff
@@ -97,6 +146,59 @@ const Slack = ({
               </label>
             </div>
           </div>
+          <div>
+            <label className="form-label mt-3">Slack app</label>
+            <Select
+              options={[
+                { value: 'from_env', name: 'Use environment variables' },
```
ganesh8056: For this string, please check if it is cloud : checkIfToolJetCloud using method

if - cloud : use string ( 'ToolJet slack app' ) 
if - not cloud : use string ( 'Use environment variables' )

ganesh8056 (changes requested):

server/data-migrations/1751480059780-BackfillSlackOptions.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
ganesh8056: server/data-migrations/1747338135635-BackfillSlackAppOption.ts

Migration file name timestamp should be same as LTS. Else the migration would again re-run for the customer from LTS. Which may update the existing options selected

ganesh8056 (changes requested):

ganesh8056 (approved):

ganesh8056 (approved):

PR #13180, opened 2025-07-03
Bugfixes/whitelabelling apis

rohanlahori:


server/src/modules/white-labelling/constant/feature.ts
```diff
@@ -5,9 +5,10 @@ import { FEATURE_KEY } from '.';
 
 export const FEATURES: FeaturesConfig = {
   [MODULES.WHITE_LABELLING]: {
-    [FEATURE_KEY.GET]: {},
+    [FEATURE_KEY.GET]: { isSuperAdminFeature: true },
```
gsmithun4: should be ispublic: true

PR #13233, opened 2025-07-07
Fix/appbuilder 02

johnsoncherian:


server/src/modules/data-queries/module.ts
```diff
@@ -37,7 +37,7 @@ export class DataQueriesModule extends SubModule {
         AppFeatureAbilityFactory,
         DataSourceFeatureAbilityFactory,
       ],
-      exports: [DataQueriesUtilService],
+      exports: [DataQueriesUtilService, DataQueriesService],
```
gsmithun4: Should not export  DataQueriesService. move functions to util.service if its reused

server/src/modules/data-queries/service.ts
```diff
@@ -24,6 +24,13 @@ export class DataQueriesService implements IDataQueriesService {
     protected readonly dataSourceRepository: DataSourcesRepository
   ) { }
 
+  async findOne(dataQueryId: string): Promise<DataQuery> {
```
gsmithun4: use data queries repository

emidhun (approved):

karan-rathod-316 (approved):
Looks good to me.

kavinvenkatachalam (approved):

PR #13285, opened 2025-07-09
Hide Prompt input if AI feature flag is false and also block prompt submission via enter if ai credits are exhausted

NishidhJain11:


johnsoncherian:
This is already handled in updating the merge commit for main module

PR #13291, opened 2025-07-09
Enhance: Extract the general OAuth flow for plugins [PRE-RELEASE]

parthy007:


frontend/src/_ui/OAuth/GrantTypes.jsx
```diff
@@ -4,127 +4,208 @@ import Select from '@/_ui/Select';
...
   const { selectedDataSource, options } = opt;
+  const { oauthTypes } = oauth_configs || {};
+
+  React.useEffect(() => {
+    if (oauthTypes?.default_value && !options?.oauth_type?.value) {
+      optionchanged('oauth_type', oauthTypes.default_value);
+    }
+  }, []);
+
+  const oauthTypeOptions = [
```
ganesh8056: Can we update the label of the drop down based on Version, specifically for cloud, similar to slack

PR #13325, opened 2025-07-10
Feature/add aws bedrock

satyy2301:
added aws-bedrock plugin


marketplace/plugins/aws-bedrock/lib/operations.json
```diff
@@ -0,0 +1,93 @@
...
+          {"value": "EMBEDDING", "name": "Embedding"}
+        ]
+      },
+      "provider": {
+        "label": "Provider",
+        "key": "provider",
+        "type": "codehinter",
+        "description": "Provider filter (optional)",
+        "className": "codehinter-plugins",
+        "placeholder": "Enter provider name"
```
ganesh8056, line 89: Placeholder is incorrect

marketplace/plugins/aws-bedrock/lib/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  QueryService,
+  ConnectionTestResult,
+} from "@tooljet-marketplace/common";
+import {
+  BedrockRuntimeClient,
+  BedrockRuntimeServiceException,
+} from "@aws-sdk/client-bedrock-runtime";
+import { SourceOptions, QueryOptions } from "./types";
+import { generateContent, listFoundationModels } from "./query_operations";
+import { BedrockServiceException } from "@aws-sdk/client-bedrock";
```
ganesh8056, line 13: Two imports can be combined, as module is same

marketplace/plugins/aws-bedrock/lib/index.ts
```diff
@@ -0,0 +1,219 @@
...
+                  model_id: !queryOptions.model_id ? "missing" : "valid",
+                  request_body: !queryOptions.request_body ? "missing" : "valid"
+                }
+              }
+            );
+          }
+          data = await generateContent(runtimeClient, {
+            model_id: queryOptions.model_id,
+            request_body: queryOptions.request_body,
+            content_type: queryOptions.content_type,
```
ganesh8056, line 46: content type is a optional params, only if we provide content type we need to add the key and value

marketplace/plugins/aws-bedrock/lib/index.ts
```diff
@@ -0,0 +1,219 @@
...
+import { generateContent, listFoundationModels } from "./query_operations";
+import { BedrockServiceException } from "@aws-sdk/client-bedrock";
+import { BedrockClient, ListFoundationModelsCommand } from "@aws-sdk/client-bedrock";
+export default class AWSBedrock implements QueryService {
+  async run(
+    sourceOptions: SourceOptions,
+    queryOptions: QueryOptions,
+    dataSourceId: string
+  ): Promise<QueryResult> {
+    const runtimeClient = this.createClient(sourceOptions);
```
ganesh8056, line 21: Based on the operation, you can init the respective client. It's not necessary to init both the clients at the same time.

marketplace/plugins/aws-bedrock/lib/index.ts
```diff
@@ -0,0 +1,219 @@
...
+          data = await generateContent(runtimeClient, {
+            model_id: queryOptions.model_id,
+            request_body: queryOptions.request_body,
+            content_type: queryOptions.content_type,
+          });
+          break;
+
+        case "list_foundation_models":
+          data = await listFoundationModels(bedrockClient, {
+            customization_type: queryOptions.customization_type,
```
ganesh8056, line 52: All the below mentioned parameters are optional. Only if there is value mentioned, respective key and value should be included. or check for default values i.e ( empty array or respective value )

marketplace/plugins/aws-bedrock/lib/query_operations.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (options.customization_type)
+    input.byCustomizationType = options.customization_type;
+  if (options.inference_type) input.byInferenceType = options.inference_type;
+  if (options.output_modality) input.byOutputModality = options.output_modality;
+  if (options.provider) input.byProvider = options.provider;
+
+  const command = new ListFoundationModelsCommand(input);
+  const response = await client.send(command);
+  if (!response.modelSummaries || response.modelSummaries.length === 0) {
+    throw new Error("No foundation models were found for the selected filters. Please adjust your criteria and try again.");
```
ganesh8056, line 62: When model summaries list is 0, it doesn't mean it's a error. We can just return Emtpy list.

marketplace/plugins/aws-bedrock/lib/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  private createCredentials(sourceOptions: SourceOptions) {
+    this.validateCredentials(sourceOptions);
+    if (sourceOptions.access_key && sourceOptions.secret_access_key) {
+      return {
+        accessKeyId: sourceOptions.access_key,
+        secretAccessKey: sourceOptions.secret_access_key,
+        sessionToken: sourceOptions.session_token ? sourceOptions.session_token : undefined
+      };
+    }
+    throw new QueryError(
```
ganesh8056, line 200: This scenario is already handled in `this.validateCredentials(sourceOptions);` . It's not required

marketplace/plugins/aws-bedrock/lib/index.ts
```diff
@@ -0,0 +1,219 @@
...
+          stack: process.env.NODE_ENV === 'development' ? error.stack : undefined
+        }
+      );
+    }
+  }
+
+  async testConnection(
+    sourceOptions: SourceOptions
+  ): Promise<ConnectionTestResult> {
+    this.validateCredentials(sourceOptions);
```
ganesh8056, line 111: `this.validateCredentials(sourceOptions);` is not required here. As the validateCredentials method is invoke again inside `this.createCredentials(sourceOptions);`

PR #13352, opened 2025-07-11
LTS Bug Fixes

rudeUltra:


server/src/modules/organizations/service.ts
```diff
@@ -25,17 +29,34 @@ export class OrganizationsService implements IOrganizationsService {
...
-      return this.organizationRepository.fetchOrganizationsForRegularUser(
+      const { organizations, totalCount } = await this.organizationRepository.fetchOrganizationsForRegularUser(
         user,
         status,
         currentPage,
         perPageCount,
         name
       );
+
+      const updatedOrganizations = await Promise.all(
```
gsmithun4, line 44: this is required only for cloud

PR #13370, opened 2025-07-13
LTS Bug Fixes

rudeUltra:


server/src/modules/versions/ability/app-version.ability.ts
```diff
@@ -125,4 +125,21 @@ export function defineAppVersionAbility(
   ) {
     can([FEATURE_KEY.GET_EVENTS], App);
   }
+
+  if (isBuilder) {
```
gsmithun4: This will allow editors to access all apps and leads to Security issues 

Shall we verify the app type is module here?

PR #13401, opened 2025-07-15
fix for data query permissions

gsmithun4:


server/src/modules/data-queries/guards/validate-query-source.guard.ts
```diff
@@ -45,7 +45,6 @@ export class ValidateQuerySourceGuard implements CanActivate {
     // Attach the found app to the request
     request.tj_data_source = dataSource;
     request.tj_resource_id = dataSource.id;
-    request.resource_type = dataSource?.type;
```
rohanlahori, line 48: Just confirming : We don't require this resource_type anywhere else where this guard is being used right?

server/src/modules/data-queries/controller.ts
```diff
@@ -129,7 +128,7 @@ export class DataQueriesController implements IDataQueriesController {
   }
 
   @InitFeature(FEATURE_KEY.RUN_VIEWER)
-  @UseGuards(QueryAuthGuard, RunQuerySourceGuard)
+  @UseGuards(QueryAuthGuard, AppFeatureAbilityGuard)
   @Post(':id/run')
   async runQuery(
```
rohanlahori, line 133: do we need the RunQuerySourceGuard somewhere, I can see it was only used in this api endpoint.
We can maybe remove that if it's not required.

server/src/modules/app-permissions/ability/index.ts
```diff
@@ -77,21 +76,5 @@ export class FeatureAbilityFactory extends AbilityFactory<FEATURE_KEY, Subjects>
       );
       return;
     }
```
rohanlahori, line 78: can we please double check this : all of these access permissions are not required for users with view access ?

server/src/modules/data-queries/controller.ts
gsmithun4: RunQuerySourceGuard and QueryAuthGuard are doing same so removed one

server/src/modules/app-permissions/ability/index.ts
gsmithun4: checked with Devanshu, its not needed. loading everything from backend and construction definition based on permissions

server/src/modules/data-queries/guards/validate-query-source.guard.ts
gsmithun4: not needed for data source, required only for app permissions

server/src/modules/data-queries/controller.ts
rohanlahori: can do it later : delete the RunQuerySourceGuard itself if it's not used at any place.

PR #13427, opened 2025-07-15
Fix: Add Performance Metrics and In-Memory Permission Filtering to App & Version APIs

devanshu052000:


server/src/modules/group-permissions/module.ts
```diff
@@ -46,7 +46,7 @@ export class GroupPermissionsModule extends SubModule {
         GroupPermissionsRepository,
         FeatureAbilityFactory,
       ],
-      exports: [GroupPermissionsUtilService],
+      exports: [GroupPermissionsUtilService, GroupPermissionsRepository],
```
gsmithun4: should not export repository. import it as a provider in other repositories

server/src/modules/modules/module.ts
```diff
@@ -40,6 +41,7 @@ export class ModulesModule {
         await AiModule.register(configs),
         await AppPermissionsModule.register(configs),
         await ImportExportResourcesModule.register(configs),
+        await GroupPermissionsModule.register(configs),
```
gsmithun4: not needed, import group permission repository as provider

server/src/modules/versions/module.ts
```diff
@@ -49,6 +51,7 @@ export class VersionModule extends SubModule {
         await AppEnvironmentsModule.register(configs),
         await ThemesModule.register(configs),
         await AppPermissionsModule.register(configs),
+        await GroupPermissionsModule.register(configs),
```
gsmithun4: can remove if group permission util service is not used inside version module

server/src/modules/workflows/module.ts
```diff
@@ -115,6 +117,7 @@ export class WorkflowsModule extends SubModule {
         await ThemesModule.register(configs),
         await AiModule.register(configs),
         await AppPermissionsModule.register(configs),
+        await GroupPermissionsModule.register(configs),
```
gsmithun4: same here

gsmithun4 (changes requested):

server/src/modules/group-permissions/module.ts
devanshu052000: Got it will fix it

frontend/src/AppBuilder/_stores/slices/appSlice.js
```diff
@@ -231,19 +232,19 @@ export const createAppSlice = (set, get) => ({
...
+    toNavigate = `${subpath ? `${subpath}` : ''}/${isPreview ? 'applications' : `${getWorkspaceId() + '/apps'}`}/${
+      slug ?? appId
+    }/${handle}?${queryParamsString}`;
+    navigate(toNavigate, {
+      state: {
+        isSwitchingPage: true,
+        id: pageId,
+        handle: handle,
+      },
+    });
```
devanshu052000: @vjaris42 Hope we are not breaking the back and forward navigation flow with this inside editor, preview and released apps. Keeping in mind on going back and forward, page should not refresh being a SPA. Please test this

.husky/pre-commit
devanshu052000, line 1: is this needed?

docker/cloud/cloud-server.Dockerfile
devanshu052000, line 1: is this needed?

johnsoncherian (approved):

PR #13477, opened 2025-07-18
Add UPS plugin integration

### LINKED ISSUE #12643: UPS- new marketplace plugin
Integrate with [UPS](https://developer.ups.com/catalog?loc=en_US) API to create and track shipments of small package, LTL, and air freight shipments.

priyxansh:
Add the UPS marketplace plugin with initial implementation.
Closes #12643


marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,218 @@
...
+      const method = operation.toUpperCase();
+      const hasBody = ["POST", "PUT", "PATCH"].includes(method);
+
+      const requestOptions: RequestInit = {
+        method,
+        headers,
+        body: hasBody ? JSON.stringify(params.request) : undefined,
+      };
+
+      const response = await fetch(fullUrl.toString(), requestOptions);
```
parthy007: Use the `got` library, reference googlesheets

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    const { client_id, client_secret, base_url, shipper_number } =
+      sourceOptions;
+
+    const headers = new Headers();
+
+    headers.append("x-merchant-id", shipper_number);
+    headers.append("Content-Type", "application/x-www-form-urlencoded");
+
+    // Basic authorization header
+    const base64Credentials = btoa(`${client_id}:${client_secret}`);
```
parthy007: Can we use something like `Buffer.from(...).toString('base64')`

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    const requestOptions = {
+      method: "POST",
+      headers: headers,
+      body: body,
+    };
+
+    const baseUrl = base_url;
+    const url = `${baseUrl}/security/v1/oauth/token`;
+
+    const response = await fetch(url, requestOptions);
```
parthy007: Use `got` library

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,218 @@
...
+        {
+          cause: error,
+        }
+      );
+      queryError.name = "QueryError";
+      throw queryError;
+    }
+  }
+
+  async testConnection(
```
parthy007, line 80: In testConnection, you'll have to run a basic query to check it the creds that you are using are working correctly. Run a simple operations that could work for everyone without providing additional inputs. You can refer testConnection of other plugins

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      const requestOptions: RequestInit = {
+        method,
+        headers,
+        body: hasBody ? JSON.stringify(params.request) : undefined,
+      };
+
+      const response = await fetch(fullUrl.toString(), requestOptions);
+
+      if (!response.ok) {
```
parthy007: Can we remove this and catch errors in catch block. When we use `got` library it will automatically throw you to catch block if the response is not ok. Refer postgresql plugin's catch block to keep the code simple. Directly use `throw new QueryError('Query could not be completed', errorMessage, errorDetails)`. Fill the errorMessage and errorDetails accordingly

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,303 @@
...
+      await got(fullUrl.toString(), {
+        method: method.toLowerCase() as Method,
+        headers,
+      });
+
+      return {
+        status: "ok",
+        message: "Connection successful",
+      };
+    } catch (error) {
```
parthy007: I see try block inside another try block, we can keep things clean by using something like 

`try {
  const response = await got(...);
  ...
} 
catch (error) {
  throw parseHttpError(error, "Failed to fetch data");
}`

Here `parseHttpError` can be another function which can be reused somewhere else too.

frontend/src/_styles/theme.scss
```diff
@@ -11883,9 +11883,9 @@ tbody {
   }
 }
 
-.selected-ds.row>img {
-  padding: 0 !important;
-}
+// .selected-ds.row>img {
```
parthy007: Can't comment styles on the common theme file, might affect other places

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,256 @@
...
+      // Append query parameters
+      for (const [key, value] of Object.entries(params.query || {})) {
+        if (value !== undefined && value !== null) {
+          fullUrl.searchParams.append(key, value);
+        }
+      }
+
+      // Prepare headers
+      const headers: Record<string, string> = {
+        "Content-Type": "application/json",
```
parthy007: Add `'Content-Type': 'application/json'` only when body is present

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,256 @@
...
+      const method = operation.toUpperCase();
+      const hasBody = ["POST", "PUT", "PATCH"].includes(method);
+
+      const response = await got(fullUrl.toString(), {
+        method: method.toLowerCase() as Method,
+        headers,
+        json: hasBody ? params.request : undefined,
+      });
+
+      const responseData = JSON.parse(response.body);
```
parthy007: I don't think we need manual parsing with `got` library, please confirm this first then make a change.

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,256 @@
...
+      this.validateSourceOptions(sourceOptions);
+      const { accessToken } = await this.generateOAuthToken(sourceOptions);
+
+      // Use a simple endpoint to test the connection
+      const baseUrl = sourceOptions.base_url + "/api";
+      const resolvedPath = "/track/v1/details/testing"; // Example path for testing
+      const fullUrl = new URL(`${baseUrl}${resolvedPath}`);
+
+      const headers: Record<string, string> = {
+        "Content-Type": "application/json",
```
parthy007: Won't need `'Content-Type': 'application/json'` with `GET` method

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    const headers = {
+      "x-merchant-id": shipper_number,
+      "Content-Type": "application/x-www-form-urlencoded",
+      Authorization: `Basic ${Buffer.from(
+        `${client_id}:${client_secret}`
+      ).toString("base64")}`,
+    };
+
+    const body = new URLSearchParams();
```
parthy007: Won't need this with `got` library. Use `form`

```
const headers = {
    "x-merchant-id": shipper_number,
    Authorization: `Basic ${Buffer.from(
      `${client_id}:${client_secret}`
    ).toString("base64")}`,
  };
  
  const formData = {
    grant_type: "client_credentials",
  };
  
  const url = `${base_url}/security/v1/oauth/token`;
  
  try {
    const response = await got(url, {
      method: "post",
      headers,
      form: formData,
      responseType: 'json' ---> This is optional just in case the got library is not automatically parsing
    });
    .....
```

marketplace/plugins/ups/lib/index.ts
```diff
@@ -0,0 +1,256 @@
...
+    const url = `${base_url}/security/v1/oauth/token`;
+
+    try {
+      const response = await got(url, {
+        method: "post",
+        headers,
+        body: body.toString(),
+      });
+
+      const data = JSON.parse(response.body);
```
parthy007: Try to remove this manual parsing otherwise can give issue with already parsed response from the `got` library

PR #13494, opened 2025-07-18
Feature/easypost plugin

satyy2301:
added easypost plugin


marketplace/package.json
```diff
@@ -30,5 +30,7 @@
     "lint": "eslint .  '**/*.ts'",
     "format": "eslint . --fix '**/*.ts'"
   },
-  "dependencies": {}
+  "dependencies": {
```
ganesh8056: EasyPost dependencies is not required, at the Root of marketplace package.json

marketplace/plugins/easypost/lib/manifest.json
```diff
@@ -0,0 +1,46 @@
...
+  "type": "api",
+  "source": {
+    "name": "EasyPost",
+    "kind": "easypost",
+    "exposedVariables": {
+      "isLoading": false,
+      "data": {},
+      "rawData": {}
+    },
+    "options": {}
```
ganesh8056: API Key to be encrypted by adding it in Options. please take of other similar plugins

marketplace/plugins/easypost/lib/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Handle body parameters
+    const bodyParams = queryOptions['params']['request'] || {};
+    const jsonBody = Object.keys(bodyParams).length > 0 ? bodyParams : undefined;
+
+    try {
+      const response = await got(url, {
+        method: operation.toLowerCase(),
+        headers: this.authHeader(apiKey),
+        searchParams: queryParams,
+        json: jsonBody
```
ganesh8056: If JSON body is empty don't send the key json

ganesh8056 (changes requested):

marketplace/plugins/easypost/lib/manifest.json
ganesh8056: It should be like, Type should be 'string' in the option
`        "type": "string",
        "encrypted": true`

marketplace/plugins/easypost/lib/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    const pathParams = queryOptions['params']['path'] || {};
+    for (const param of Object.keys(pathParams)) {
+      url = url.replace(`{${param}}`, pathParams[param]);
+    }
+
+    // Handle query parameters
+    const queryParams = queryOptions['params']['query'] || {};
+
+    // Handle body parameters
+    const bodyParams = queryOptions['params']['request'] || {};
```
ganesh8056: can we rename bodyParams to requestBody, variable naming is confusing

ganesh8056 (dismissed):

PR #13537, opened 2025-07-29
Pending Issues Pre-release platform

rohanlahori:


server/src/helpers/utils.helper.ts
```diff
@@ -488,3 +488,22 @@ export function decamelizeKeysExcept(obj: any, ignoreKeys: string[]): any {
 
   return obj;
 }
+export function objectGUIDtoString(buffer) {
```
gsmithun4, line 491: are we using this anywhere?

PR #13552, opened 2025-07-30
Fix: Query run failing for public released app

devanshu052000:
This PR updates the query execution logic to skip the query permission check if the released app is public. Previously, even public apps were subject to permission checks, which could prevent queries from running as we used to access undefined user object to get organizationId

[ee-server](https://github.com/ToolJet/ee-server/pull/190)


devanshu052000:
Fixed in [PR](https://github.com/ToolJet/ToolJet/pull/13556)

PR #13612, opened 2025-08-01
Fix: Response, Request and Metadata coming out to be undefined on query failure and success inside Inspector

devanshu052000:
This PR fixes an issue where the Inspector panel was showing undefined for Response, Request, and Metadata fields after query execution (both on success and failure). The root cause was incorrect data extraction from the query result and error objects.

Changes:
- Corrected the assignment of finalData in processQueryResults to use the proper data structure.
- Updated error handling for REST API queries to extract requestObject, responseObject, and responseHeaders from the correct nested location (errorData.data).
- Ensured consistent data flow to the Inspector for both successful and failed queries.


devanshu052000:
Fixed in the following PR -> [#13576](https://github.com/ToolJet/ToolJet/pull/13576)

PR #13618, opened 2025-08-01
Fix widget dragging width calculation in tab containers

johnsoncherian:
## Summary
  - Fixes incorrect width calculation when dragging widgets within tab containers
  - Adds grid parent type tracking to distinguish tab containers from regular containers
  - Prevents substring transformation of canvas parent IDs for tab containers


johnsoncherian:
This is fixed in #13688

PR #13735, opened 2025-08-08
[docs]: GitSync Migration Changes

RudraPratik30:


RudraPratik30:
1. Choose Your Connection Method

- [x] Branch Configuration → Both are configurable from the UI (From 3.16.0).
- [x] Default Branch → Both have default branch main (From 3.16.0).
- [x] Require GitHub App → This field should be removed as GitLab is also supported, so this is not correct.

2. Configure Git Sync Page
- [x] Second line → By default GitSync is configured for the master branch → Update to default branch is main (From 3.16.0).
- [x] Steps 6 to 10 can be combined into fewer points (maximum 3), as they only describe adding the SSH key to the repository deploy keys, which is currently spread across 5 steps.
- [x] Using the Environment Variable → Remove this section completely, as there is no .env support (From 3.16.0).

3. Configure Git Repo Manager
- [ ] Reduce the number of points. For example, GitLab currently has 8 points and some of them can be combined.
       [@Pratik](https://tooljet.slack.com/team/U07HEN1UJUW) can you please review this section and decide how it should be structured?

4. GitHub Configuration
- [x] Add the following Repository permissions → This is important and should be highlighted or displayed differently so it stands out clearly for users.
- [x] Here, make sure to copy the App ID → Keep this as a note or info text instead of including it within the same step.
- [x] can we please modify the below point for better clarity (can try breaking it into 2 points) :
    Current point:
       After creating the GitHub App, you will be directed to the GitHub App Settings page. Here, make sure to copy the App ID. Next, generate the       Private key (.pem key), download it, and store it securely. This information will be essential later when configuring GitSync.

5. GitLab Configuration  : This section looks clear and to the point.

6. For GitLab and GitHub Configurations
- [x] Do not use the phrase instead of the traditional SSH-based method.
- [x] Explain simply that HTTPS is used here and mention its benefits.

karan-rathod-316 (approved):
Looks good to me.

PR #13749, opened 2025-08-11
Fix: Apply default values for missing props in Multiselect dynamic Options

vaishnaviparabkar90:
This PR updates the MultiSelect component logic so that dynamic options work correctly with only the essential properties (label, value, default) without requiring disable and visible to be explicitly defined


RudraPratik30:
Closing this PR since the issue was not open for community contribution and the PR was raised without prior assignment.

PR #13764, opened 2025-08-11
Fix: AfterShip plugin API key option naming updated

ganesh8056:
**Changes :**
The most significant changes involve renaming the API key property from `as-api-key` to `apiKey` throughout the implementation, manifest, and type definitions, as well as minor formatting improvements.

### DIFF
--- a/marketplace/plugins/aftership/lib/manifest.json
+++ b/marketplace/plugins/aftership/lib/manifest.json
@@ -13,7 +13,7 @@
       "rawData": {}
     },
     "options": {
-      "as-api-key": {
+      "apiKey": {
         "type": "string",
         "encrypted": true
       }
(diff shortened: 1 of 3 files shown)


marketplace/plugins/aftership/lib/manifest.json
```diff
@@ -13,23 +13,23 @@
...
   "defaults": {},
   "properties": {
-    "as-api-key": {
+    "apiKey": {
       "label": "API key",
-      "key": "as-api-key",
+      "key": "apiKey",
       "type": "password",
       "description": "Enter your Aftership API key",
       "hint": "You can find your API key in your Aftership account settings."
```
parthy007, line 29: Here we use `helpText` instead of `hint`

PR #13850, opened 2025-08-18
Xero Plugin Intergration

Pratush613:
This PR introduces a new ToolJet plugin to integrate with Xero’s API, enabling seamless access to accounting, invoicing, and financial data from connected Xero organizations. The plugin supports OAuth 2.0 authentication, allowing users to securely connect multiple Xero tenants.

Key Features
OAuth 2.0 based authentication flow for secure access
Fetch and manage invoices, contacts, and payments data
Create and update accounting records directly from ToolJet apps
Retrieve financial reports and statements for better insights
Support for working with multiple Xero organizations via tenant IDs

This integration empowers users to build custom finance dashboards, automate workflows, and streamline internal financial tools within ToolJet


marketplace/plugins/xero/lib/index.ts
```diff
@@ -0,0 +1,347 @@
...
+
+
+    let url = `https://api.xero.com/api.xro/2.0${path}`;
+
+    if (specType === "finance") {
+      url = `https://api.xero.com/finance.xro/1.0${path}`
+    }
+
+    for (const param in pathParams) {
+      url = url.replace(`{${param}}`, pathParams[param]);
```
parthy007: Use `url = url.replace({${param}}, encodeURIComponent(pathParams[param]));`

marketplace/plugins/xero/lib/index.ts
```diff
@@ -0,0 +1,347 @@
...
+
+      if (operation && typeof operation === 'string' && !['get', 'delete'].includes(operation)) {
+        if (bodyParams && Object.keys(bodyParams).length > 0) {
+          requestOptions['json'] = bodyParams;
+        }
+      }
+    }
+    try {
+      const response = await got(url, requestOptions);
+      if (response.statusCode !== 200) {
```
parthy007: Can we remove this if block and let the final catch block handle errors, will give us a cleaner approach

marketplace/plugins/xero/lib/index.ts
```diff
@@ -0,0 +1,347 @@
...
+      }
+      const errorMessage =
+        parsed?.Title ||
+        parsed?.error_description ||
+        parsed?.error ||
+        'Xero API request failed';
+
+      const errorDetails = {
+        statusCode: error?.response?.statusCode,
+        response: parsed,
```
parthy007: Please confirm that the server returns error details as compact key-value pairs that are immediately useful to end users. Avoid returning the entire response body, which can become verbose and may include unnecessary data like file paths. Would you mind testing this to ensure it’s not the case?

marketplace/plugins/xero/lib/manifest.json
```diff
@@ -0,0 +1,97 @@
...
+      "key": "oauth",
+      "type": "react-component-oauth",
+      "description": "A component for Xero integration",
+      "oauth_configs": {
+        "allowed_field_groups": {
+          "authorization_code": [
+            "client_id",
+            "client_secret"
+          ],
+          "basic": [
```
parthy007: If Basic flow is not needed then you don't need to add this, confirm the product requirement

marketplace/plugins/xero/lib/manifest.json
```diff
@@ -0,0 +1,97 @@
...
+    "scopes": {
+      "value": "openid profile email offline_access accounting.transactions accounting.contacts accounting.reports.read accounting.settings"
+    },
+    "oauth_type": {
+      "value": "custom_app"
+    },
+    "access_token_url": {
+      "value": "https://identity.xero.com/connect/token"
+    },
+    "database": {
```
parthy007: Don't need database, schema, warehouse. Try to understand the flow and logic

marketplace/plugins/xero/lib/index.ts
```diff
@@ -0,0 +1,347 @@
...
+      client_id: clientId,
+      client_secret: clientSecret,
+      grant_type: 'authorization_code',
+      redirect_uri: redirectUri,
+    });
+
+    try {
+      const response = await got('https://identity.xero.com/connect/token', {
+        method: 'post',
+        body: data.toString(),
```
parthy007: Use `form` instead of `body`. I have added `responseType: 'json'` so you don't have to parse manually, do test once. Also you won't need to add `Content-Type` with `form`. Do the same thing for refreshToken as well.

```
const data = {
  client_id: sourceOptions.client_id,
  client_secret: sourceOptions.client_secret,
  grant_type: 'refresh_token',
  refresh_token,
};

const response = await got('https://identity.xero.com/connect/token', {
  method: 'post',
  form: data,
  responseType: 'json'
});
```

marketplace/plugins/xero/lib/index.ts
```diff
@@ -0,0 +1,347 @@
...
+      header_prefix: 'Bearer ',
+      audience: '',
+      client_auth: 'header',
+      headers: [['', ''], ['tj-x-forwarded-for', '::1']],
+      custom_query_params: [['', '']],
+      custom_auth_params: [['', '']],
+      access_token_custom_headers: [['', '']],
+      ssl_certificate: 'none',
+      retry_network_errors: true,
+      scopes: encodeURIComponent(`openid profile email offline_access accounting.transactions accounting.transactions.read accounting.reports.read accounting.reports.tenninetynine.read accounting.journals.read accounting.settings accounting.settings.read accounting.contacts accounting.contacts.read accounting.attachments accounting.attachments.read accounting.budgets.read finance.statements.read finance.accountingactivity.read finance.cashvalidation.read finance.bankstatementsplus.read`)
```
parthy007: Isn't there a better way to add all scopes?

marketplace/plugins/xero/lib/index.ts
```diff
@@ -0,0 +1,347 @@
...
+      if (_requestOptions.status === 'needs_oauth') return _requestOptions;
+      requestOptions = _requestOptions.data as OptionsOfTextResponseBody;
+      requestOptions.headers = {
+        ...(requestOptions.headers || {}),
+        'Xero-tenant-id': sourceOptions['tenant_id'],
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+      };
+    } else {
+      requestOptions = {
```
parthy007, line 328: I'm not understanding how this will work? If multi-auth is disabled then `const accessToken = sourceOptions['access_token'];`, can you check if this works correctly

marketplace/plugins/xero/lib/manifest.json
```diff
@@ -54,14 +54,20 @@
...
-  },
+    },
   "properties": {
     "tenant_id": {
       "key": "tenant_id",
       "type": "text",
       "label": "Tenant ID",
       "description": "Enter your Xero Tenant ID (get it from https://api.xero.com/connections)"
       },
+      "scopes": {
```
parthy007: Won't need this separate scope field, just turn this key to true `"allowed_scope_field"`

marketplace/plugins/xero/lib/index.ts
```diff
@@ -0,0 +1,371 @@
...
+    let requestOptions;
+    if (sourceOptions['multiple_auth_enabled']) {
+      const customHeaders = { 'tj-x-forwarded-for': '::1' };
+      const newSourceOptions = this.constructSourceOptions(sourceOptions);
+      const authValidatedRequestOptions = this.convertQueryOptions(queryOptions, customHeaders);
+
+      const _requestOptions = await validateAndSetRequestOptionsBasedOnAuthType(
+        newSourceOptions,
+        context,
+        authValidatedRequestOptions as any
```
parthy007: You'll have to add one more option
```
validateAndSetRequestOptionsBasedOnAuthType(
        newSourceOptions,
        context,
        authValidatedRequestOptions as any
        { kind: 'xero' }          <---- Add this
```

marketplace/plugins/xero/lib/manifest.json
```diff
@@ -45,18 +45,12 @@
     "oauth_type": {
       "value": "custom_app"
     },
+    "client_secret": {
```
parthy007: This doesn't go in defaults rather in options object. Defaults object only takes care of filling in default value, encryption and initial type setting is done by options object

marketplace/plugins/xero/lib/manifest.json
```diff
@@ -0,0 +1,93 @@
...
+      "value": "http://localhost:8082/oauth2/authorize",
+      "encrypted": false
+    },
+    "auth_url": {
+      "value": "https://login.xero.com/identity/connect/authorize",
+      "encrypted": false
+    },
+    "allowed_auth_types": {
+      "value": "oauth2",
+      "encrypted": false
```
parthy007: Don't add these encrypted tag inside default object to any key. Also encrypted false is set for everyone until its true, so its never required

PR #13930, opened 2025-08-26
fix:storing the UTM params in the local storage and passing the UTM p…

Dheeraj-P-Girish:
Implemented storing of UTM parameters in local storage

Ensured UTM parameters are appended and passed correctly to tooljet.ai


RudraPratik30:
Looks good to me

PR #14012, opened 2025-09-09
Feat: Add FedEx plugin

priyxansh:
Introduce FedEx plugin to the ToolJet Marketplace.


marketplace/plugins/fedex/lib/manifest.json
```diff
@@ -0,0 +1,160 @@
...
+      "child_key": {
+        "type": "string"
+      },
+      "child_secret": {
+        "type": "string"
+      },
+      "base_url": {
+        "type": "string"
+      },
+      "child_key": {
```
parthy007, line 31: ```child_key``` is written twice

marketplace/plugins/fedex/lib/manifest.json
```diff
@@ -0,0 +1,160 @@
...
+        "type": "string"
+      },
+      "child_secret": {
+        "type": "string",
+        "encrypted": true
+      }
+    }
+  },
+  "defaults": {
```
parthy007, line 40: Inside this default object encrypted tag is not required for anyone, only applies for the options object

marketplace/plugins/fedex/lib/manifest.json
```diff
@@ -0,0 +1,160 @@
...
+    "oauth": {
+      "key": "oauth",
+      "type": "react-component-oauth",
+      "description": "OAuth configuration for FedEx",
+      "oauth_configs": {
+        "allowed_field_groups": {
+          "client_credentials": [
+            "client_id",
+            "client_secret",
+            "child_key",
```
parthy007: ```child_key``` & ```child_secret``` won't be put here, oauth component won't render these component. Also can you put this oauth object as the last object inside properties (will make the data config page more consistent with other plugins)

marketplace/plugins/fedex/lib/index.ts
```diff
@@ -0,0 +1,275 @@
...
+    let errorMessage = defaultMessage;
+    let errorDetails: Record<string, any> = {};
+
+    if (error instanceof HTTPError && error.response) {
+      errorMessage = `HTTP ${error.response.statusCode}: ${error.response.statusMessage || "Request failed"
+        }`;
+      errorDetails = {
+        status: error.response.statusCode,
+        statusText: error.response.statusMessage,
+        errors: error.response.body,
```
parthy007: Lets not pass the whole body, and follow key-value approach, the whole body might have sensitive/unwanted info

marketplace/plugins/fedex/lib/index.ts
```diff
@@ -0,0 +1,275 @@
...
+
+      // Use the access token to make a test API call to verify connectivity
+      const testUrl = `${base_url}/rate/v1/rates/quotes`;
+
+      await got.post(testUrl, {
+        headers: {
+          Authorization: `Bearer ${accessToken}`,
+          Accept: "application/json",
+        },
+        json: {},
```
parthy007, line 82: Does this works with empty body, rather can we use any get request for testing purpose?

marketplace/plugins/fedex/lib/index.ts
```diff
@@ -0,0 +1,275 @@
...
+
+      const responseData = response.body;
+
+      return {
+        status: "ok",
+        data: responseData as any,
+      };
+    } catch (error) {
+      if (error instanceof QueryError) {
+        throw error;
```
parthy007, line 60: Will this be properly caught as per our error standardisation?

priyxansh: Yes. All the errors thrown from the helper functions used in the run functions are of type `QueryError`, so they'll be thrown as is with correct information (message, data, etc.). And if it's not an instance of `QueryError` it'll be handled by the `parseHttpError` function, converting those errors into `QueryError`.

Issue #13976, opened 2025-09-04
OAuth multi user flow to redirect user to login on config page only rather than just save


mansukh-tj:
Problem:
When multi user auth is toggled on, we save the config without testing if Oauth app creds are valid or not.

Solution:
We let the user who is configuring the data source, connect via his credentials and save it for this user in ToolJet.
Note: That any new user/builder who executes queries with this data source has to be redirected and authenticated with his own credentials.

PR #14043, opened 2025-09-12
Feature: OAuth multi user flow to redirect user to login on config page only rather than just save

parthy007:
Closes: [#13976](https://github.com/ToolJet/ToolJet/issues/13976)


plugins/packages/snowflake/lib/index.ts
```diff
@@ -231,7 +238,8 @@ export default class Snowflake implements QueryService {
   ): Promise<any> {
     if (checkCache) {
       const optionsHash = generateSourceOptionsHash(sourceOptions);
-      const enhancedCacheKey = `${dataSourceId}_${optionsHash}`;
+      const userId = context?.user?.id;
```
ganesh8056, line 241: Only if we have used based auth, we would need to cache based on User Id ? Also this edge case should be addressed in all of the cache logic

server/src/modules/data-sources/util.service.ts
```diff
@@ -724,18 +764,34 @@ export class DataSourcesUtilService implements IDataSourcesUtilService {
   }
 
   /* this function only for getting auth token for googlesheets and related plugins*/
-  async fetchAPITokenFromPlugins(dataSource: DataSource, code: string, sourceOptions: any) {
+  async fetchAPITokenFromPlugins(
```
ganesh8056, line 765: Since user based auth flow also added in config page, there is redundant code for Fetching access token

ganesh8056: Let's skip this review comment.

PR #14055, opened 2025-09-15
Feature/design changes

rohanlahori:


frontend/src/App/App.jsx
```diff
@@ -265,15 +266,31 @@ class AppComponent extends React.Component {
...
-                {auth(this.props)}
-                <Route path="/sso/:origin/:configId" exact element={<Oauth {...this.props} />} />
-                <Route path="/sso/:origin" exact element={<Oauth {...this.props} />} />
+                {onboarding({ ...this.props, darkMode })}
+                {auth({ ...this.props, darkMode })}
+                <Route
+                  path="/sso/:origin/:configId"
+                  exact
+                  element={
+                    <PrivateRoute darkMode={darkMode}>
```
rohanlahori: @nithindavid 
why are we adding Private Route guard on sso routes?
This breaks the sso flow -> this route is expected to be public.

can we please revert adding PrivateRoute guard to extra routes where it's not expected

frontend/src/App/App.jsx
```diff
@@ -403,29 +429,43 @@ class AppComponent extends React.Component {
                   </Route>
                 )}
```
rohanlahori, line 433: why are we adding private route guard here? is it required or it's done by mistake ?
@nithindavid

frontend/src/App/App.jsx
```diff
@@ -403,29 +429,43 @@ class AppComponent extends React.Component {
...
+                      <Navigate to="/:workspaceId" />
+                    </PrivateRoute>
+                  }
+                />
                 <Route
                   exact
                   path="/error/:errorType"
-                  element={<ErrorPage switchDarkMode={this.switchDarkMode} darkMode={darkMode} />}
+                  element={
+                    <PrivateRoute darkMode={darkMode}>
```
rohanlahori: The error page can come for public apps also : so we should avoid adding a private route here.

@nithindavid

frontend/src/App/App.jsx
```diff
@@ -403,29 +429,43 @@ class AppComponent extends React.Component {
...
                   element={
-                    <SwitchWorkspacePage
-                      switchDarkMode={this.switchDarkMode}
-                      darkMode={darkMode}
-                      archived={true}
-                      isAppUrl={true}
-                    />
+                    <PrivateRoute darkMode={darkMode}>
+                      <SwitchWorkspacePage
+                        switchDarkMode={this.switchDarkMode}
```
rohanlahori: Let's remove the private route guard from here also can be an issue for public apps

PR #14064, opened 2025-09-15
design: Setup shadcn, upgrade storybook, story for Login page

nithindavid:
image


frontend/src/components/ui/Input/Input.jsx
```diff
@@ -1,11 +1,28 @@
...
+    size: {
+      small: 'tw-text-[12px]/[18px] tw-px-[10px] tw-py-[6px] tw-rounded-[6px]',
+      medium: 'tw-text-[12px]/[18px] tw-px-3 tw-py-[7px] tw-rounded-[6px]',
+      large: 'tw-text-[14px]/[20px] tw-px-3 tw-py-[7px] tw-rounded-[8px]',
+    },
+  },
+  defaultVariants: {
+    size: 'medium',
+  },
+});
```
devanshu052000: What is the use of this? since we are adding buttonSizeVariants to the input instead of a button

frontend/src/components/ui/Input/Input.jsx
```diff
@@ -23,12 +40,20 @@ const Input = React.forwardRef(
 
     const validationClass = response === true ? 'valid-textarea' : response === false ? 'invalid-textarea' : '';
 
+    const inputStyle = ` placeholder:tw-text-text-placeholder ${
+      isValid === true ? '!tw-border-border-success-strong' : isValid === false ? '!tw-border-border-danger-strong' : ''
+    }`;
+
```
devanshu052000: Aren't we already passing this down from the wrapper components `NumberInput` and `TextInput` components? If we are adding it to the base component,  can we make sure we aren't passing down redundant classes from wrapper ones? Will make the code a lot cleaner I feel. What are your thoughts?

frontend/src/components/ui/Input/Input.jsx
```diff
@@ -37,23 +62,34 @@ const Input = React.forwardRef(
...
-            className={cn(
-              inputVariants({ size }),
-              `tw-peer tw-flex tw-text-[12px]/[18px] tw-w-full tw-rounded-[8px] tw-border-[1px] tw-border-solid tw-bg-background-surface-layer-01 tw-py-[7px] tw-text-text-default focus-visible:tw-ring-[1px] focus-visible:tw-ring-offset-[1px] focus-visible:tw-ring-border-accent-strong focus-visible:tw-ring-offset-border-accent-strong focus-visible:tw-border-transparent disabled:tw-cursor-not-allowed ${props.styles}`,
-              className
-            )}
-            ref={ref}
-            {...props}
-          />
-        )}
-        {isPasswordField && !multiline && (
```
devanshu052000, line 51: Can we still keep the multiline condition? Since it might look weird if someone passes down type as password but also adds multiline as true

frontend/src/components/ui/Input/Input.jsx
```diff
@@ -37,23 +62,34 @@ const Input = React.forwardRef(
...
-          <div onClick={togglePasswordVisibility}>
-            {isPasswordVisible ? (
-              <SolidIcon className="eye-icon" name="eye" />
-            ) : (
-              <SolidIcon className="eye-icon" name="eyedisable" />
+          <div className="tw-relative">
```
devanshu052000: Can we revert this if not necessary? This might break the existing component being used inside the codebase.

devanshu052000 (changes requested):

nithindavid:
@devanshu052000 Thanks for the detailed review, I have fully revamped the Input & Child components to fully utilise the cva to support all types of sizes and variants.

Also, extracted the logic for Encrypted field & workspace Constant [Needs to test]

frontend/src/components/ui/Input/Input.jsx
nithindavid: Separated the password from Input, so it won't be multiline

frontend/src/components/ui/Input/Input.jsx
nithindavid: Makes sense, created a Variants.jsx and moved common styles to Input/Input.jsx

PR #14203, opened 2025-09-29
base image fix

slancerk:
testing purpose


adishM98:
Need to add the same changes to the pre-release EE image as well

PR #14242, opened 2025-10-03
Fix: Updated code editor UI for test parameters 

medhansh-alt:


frontend/src/AppBuilder/CodeEditor/MultiLineCodeEditor.jsx
```diff
@@ -255,6 +263,7 @@ const MultiLineCodeEditor = (props) => {
           tip="Pop out code editor into a new window"
           isMultiEditor={true}
           isQueryManager={isInsideQueryPane}
+          position={{ height: height }}
```
shah21:
```suggestion
          position={{ height }}
```

frontend/src/AppBuilder/CodeEditor/MultiLineCodeEditor.jsx
```diff
@@ -264,21 +273,24 @@ const MultiLineCodeEditor = (props) => {
...
+              data-cy={`${cyLabel}-input-field`}
+              style={{ height: heightValue }}
+            >
               <CodeMirror
                 ref={editorRef}
                 value={initialValueWithReplacedIds}
                 placeholder={placeholder}
-                height={'100%'}
-                minHeight={heightInPx}
-                {...(isInsideQueryPane ? { maxHeight: '100%' } : {})}
```
shah21, line 281: Is the height value the same for every multi-line code editor throughout the entire product now?

medhansh-alt: The default height is 300px when no height prop is passed on, before the styling had 300px hardcoded, so passing height prop had no effect on the component, changed that and the fallback value is now 300px

PR #14257, opened 2025-10-06
feat: implement SSRF protection with URL validation across plugins

adishM98:


plugins/packages/common/lib/ssrf-protection.ts
```diff
@@ -0,0 +1,580 @@
+import { URL } from 'url';
+import { QueryError } from './query.error';
+import * as dns from 'dns/promises';
+import { lookup as dnsLookup } from 'dns';
+import * as net from 'net';
+
+const CLOUD_METADATA_ENDPOINTS = [
+  // AWS EC2 metadata endpoint (most common SSRF target)
+  /^169\.254\.169\.254$/,
```
ganesh8056, line 9: IBM & OpenStack as well uses the same IP address. Could you mention it over comments

plugins/packages/common/lib/ssrf-protection.ts
```diff
@@ -0,0 +1,580 @@
...
+// Dangerous schemes that could lead to SSRF vulnerabilities
+const DANGEROUS_SCHEMES = [
+  'file',
+  'ftp',
+  'dict',
+  'gopher',
+  'jar',
+  'data',
+  'javascript',
+];
```
ganesh8056: `DANGEROUS_SCHEMES` is un-used. Also we are making use of environment variable for Blacklisting the schemes. Shouldn't we consider having basic blacklist already set from our end ?

ganesh8056: Or can we document this. So it can give more clarity.

plugins/packages/common/lib/ssrf-protection.ts
```diff
@@ -0,0 +1,580 @@
...
+
+  return CLOUD_METADATA_ENDPOINTS.some(pattern => pattern.test(normalized));
+}
+
+/**
+ * Normalize alternative IP formats to standard decimal notation
+ * Handles: hex (0xA9FEA9FE), decimal (2852039166), octal (0251.0376.0251.0376)
+ */
+function normalizeIPFormat(ip: string): string {
+  if (!ip) return ip;
```
ganesh8056, line 94: can we consider trimming the IP address for any whitespace

plugins/packages/common/lib/ssrf-protection.ts
```diff
@@ -0,0 +1,580 @@
...
+
```
ganesh8056, line 95: we can pick up later : 

Mixed-base dotted formats (classic bypass trick)
127.0x0.0x0.1
0177.0.0.1
0x7f.0.0.1

In the case of Mixed format, later we can upgrade our code to handle the logic in more normalised way

plugins/packages/common/lib/ssrf-protection.ts
```diff
@@ -0,0 +1,580 @@
...
+  // Carrier-grade NAT: 100.64.0.0/10
+  if (a === 100 && b >= 64 && b <= 127) {
+    return true;
+  }
+
+  // Broadcast: 255.255.255.255
+  if (a === 255 && b === 255 && c === 255 && d === 255) {
+    return true;
+  }
+
```
ganesh8056, line 223: Should we need to consider 

1. 224.0.0.0/4  ->  Multicast
2. 240.0.0.0/4 -> Reserved

plugins/packages/common/lib/ssrf-protection.ts
```diff
@@ -0,0 +1,580 @@
...
+      if (ipv4Part.includes('.')) {
+        return isPrivateIP(ipv4Part);
+      }
+    }
+  }
+
+  return false;
+}
+
+export async function resolvesToCloudMetadata(hostname: string): Promise<boolean> {
```
ganesh8056: `resolvesToCloudMetadata` is a un-used method.

ganesh8056: Also the check for **cloud metadata endpoint** is already exists in `resolvesToPrivateIP() -> isPrivateIP()` flow

plugins/packages/common/lib/ssrf-protection.ts
```diff
@@ -0,0 +1,580 @@
...
+        return callback(error);
+      }
+
+      // Cache the validated resolution to prevent TOCTOU attacks
+      dnsCache.set(hostname, { address, family, timestamp: now });
+
+      // Clean up old cache entries to prevent memory leak
+      if (dnsCache.size > 1000) {
+        const entriesToDelete: string[] = [];
+        dnsCache.forEach((value, key) => {
```
ganesh8056, line 468: We can delete the cached values in the same iteration itself

plugins/packages/openapi/lib/index.ts
```diff
@@ -55,6 +57,10 @@ export default class Openapi implements QueryService {
     const { request, query, header, path: pathParams } = params;
     const resolvedHost = sourceOptions.host || host;
     const url = new URL(resolvedHost + this.resolvePathParams(pathParams, path));
+
+    // SSRF Protection: Validate URL before making request
+    await validateUrlForSSRF(url.toString());
```
ganesh8056, line 62: What if `URL` has redirect url in its query params. Should we need to consider those scneario as well 

Example `/product/nextProduct?currentProductId=6&path=http://evil-user.net`

ganesh8056 (dismissed):

PR #14258, opened 2025-10-06
[feat]: App history

kavinvenkatachalam:
Closes https://github.com/ToolJet/tj-ee/issues/3775

EE frontend - https://github.com/ToolJet/ee-frontend/pull/256
EE server - https://github.com/ToolJet/ee-server/pull/271

### DIFF
--- a/frontend/src/AppBuilder/LeftSidebar/AppHistory/_styles/app-history.scss
+++ b/frontend/src/AppBuilder/LeftSidebar/AppHistory/_styles/app-history.scss
@@ -24,7 +24,7 @@
     display: flex;
     align-items: center;
     justify-content: space-between;
-    padding: 12px 16px 12px 20px;
+    padding: 16px;
     border-bottom: 1px solid var(--border-weak, #E4E7EB);
 
     .app-history-header-title {
@@ -64,7 +64,7 @@
 
       .history-container {
         width: 100%;
-        
+
         .history-entry {
           height: 52px;
           display: flex;
@@ -72,12 +72,12 @@
           justify-content: space-between;
           gap: 8px;
           padding: 7px 0;
-  
+
           &:hover {
             border-radius: 6px;
             margin: 0 -6px;
             padding: 7px 6px;
-            background: var(--interactive-hover, rgba(136, 144, 153, 0.12));
+            background: rgba(136, 144, 153, 0.12);
 
             .entry-option-btn-wrapper {
               .entry-option-btn {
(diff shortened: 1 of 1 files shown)


server/src/entities/app_history.entity.ts
```diff
@@ -0,0 +1,89 @@
...
+  (history_type = 'snapshot' AND jsonb_typeof(change_payload) = 'object') OR
+  (history_type = 'delta' AND jsonb_typeof(change_payload) = 'array')
+`
+)
+export class AppHistory extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @Column({ name: 'app_id', type: 'uuid' })
+  appId: string;
```
gsmithun4: app id can get from 
app_version -> app_id

won't be a good idea to add multiple relations

server/src/entities/app_history.entity.ts
```diff
@@ -0,0 +1,89 @@
...
+  @CreateDateColumn({ default: () => 'now()', name: 'created_at' })
+  createdAt: Date;
+
+  @UpdateDateColumn({ default: () => 'now()', name: 'updated_at' })
+  updatedAt: Date;
+
+  // Relations
+  @ManyToOne(() => App, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'app_id' })
+  app: App;
```
gsmithun4: Can use join table and get app
use AppBase entity 
App entity has subscriber which loads latest version on load

server/src/modules/app-history/module.ts
```diff
@@ -0,0 +1,68 @@
...
+import { getImportPath } from '@modules/app/constants';
+import { VersionRepository } from '@modules/versions/repository';
+import { AppsRepository } from '@modules/apps/repository';
+import { BullModule } from '@nestjs/bull';
+import { FeatureAbilityFactory } from './ability';
+import { NameResolverRepository } from '@modules/app-history/repositories/name-resolver.repository';
+import { AppHistoryRepository } from '@modules/app-history/repository';
+@Module({})
+export class AppHistoryModule extends SubModule {
+  private static isProcessorRegistered = false;
```
gsmithun4: instead of this we can use
```
static async register(configs: { IS_GET_CONTEXT: boolean }, isMainImport: boolean = false): Promise<DynamicModule> {
```
isMainImport - true only once from main.ts 
if (isMainImport) {
const { HistoryQueueProcessor } = await import(`${importPath}/app-history/queue/history-queue.processor`);
      providers.push(HistoryQueueProcessor);
}
If you don't want this to be added for migrations app load
IS_GET_CONTEXT = true only if its used on migration

if (isMainImport && !IS_GET_CONTEXT) {
// Its from main.ts and not from migrations
const { HistoryQueueProcessor } = await import(`${importPath}/app-history/queue/history-queue.processor`);
      providers.push(HistoryQueueProcessor);
}

johnsoncherian (approved):

PR #14277, opened 2025-10-07
Hotfix: MariaDB datasource connection handling

ganesh8056:
This PR has changes for :

1. The default connection limit in the connection pool has been updated from 5 to 10.
2. Instead of a connection pool, a single TCP connection is created for testing connection flow.
3.


plugins/packages/mariadb/lib/index.ts
```diff
@@ -61,18 +95,69 @@ export default class Mariadb implements QueryService {
...
+    dataSourceId: string,
+    dataSourceUpdatedAt?: string
+  ): Promise<any> {
+    if (checkCache) {
+      const optionsHash = generateSourceOptionsHash(sourceOptions);
+      const enhancedCacheKey = `${dataSourceId}_${optionsHash}`;
+      const cachedConnectionPool = await getCachedConnection(enhancedCacheKey, dataSourceUpdatedAt);
+      if (cachedConnectionPool) return cachedConnectionPool;
+
+      const connectionPool = await this.buildConnectionPool(sourceOptions);
```
parthy007: ```await``` is not required

parthy007 (approved):

PR #14336, opened 2025-10-15
Enhancement: enhance UI components with new icons and styling adjustments

manishkushare:
Resolves [Issue](https://github.com/ToolJet/tj-ee/issues/4147)


frontend/src/AppBuilder/Header/EnvironmentManager/EnvironmentSelectBox.jsx
```diff
@@ -71,11 +71,13 @@ const EnvironmentSelectBox = React.memo(function EnvironmentSelectBox({ options,
       <div className="d-inline-flex align-items-center env-header" onClick={() => setShowOptions(!showOptions)}>
         <Layers width="16" height="16" />
         <div data-cy="list-current-env-name">{capitalize(currentEnv.name)}</div>
+        Manish
```
nakulnagargade: .

frontend/src/_ui/Icon/solidIcons/PanelBottomCloseLucide.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { PanelBottomClose } from 'lucide-react';
```
nakulnagargade: Do we need to add this wrappers?, IMO, don't think we need to create this wrappers as this will add extra overhead to maintain in future.

nakulnagargade (approved):

frontend/src/components/ui/Button/Button.jsx
```diff
@@ -141,6 +141,8 @@ const Button = forwardRef(
       iconOnly = false, // as normal button and icon have diff styles make sure to pass it as truw when icon only button is used
       loaderText = null,
       isLucid = false, // Use Lucide icons instead of SolidIcon
+      leadingIconWidth = '24',
+      trailingIconWidth = '24',
```
devanshu052000: Would recommend not adding icon size changes in this component as it might lead to backward compatibility issues

frontend/src/components/ui/Button/Button.jsx
```diff
@@ -156,7 +158,7 @@ const Button = forwardRef(
         {isLucid ? (
           <DynamicIcon name={leadingIcon} size={iconSize} className={lucideIconClassName} />
         ) : (
-          <SolidIcon name={leadingIcon} fill={iconFillColor} />
+          <SolidIcon name={leadingIcon} fill={iconFillColor} width={leadingIconWidth} />
```
devanshu052000: same as above avoid size changes

frontend/src/components/ui/Button/Button.jsx
```diff
@@ -165,7 +167,7 @@ const Button = forwardRef(
         {isLucid ? (
           <DynamicIcon name={trailingIcon} size={iconSize} className={lucideIconClassName} />
         ) : (
-          <SolidIcon name={trailingIcon} fill={iconFillColor} />
+          <SolidIcon name={trailingIcon} fill={iconFillColor} width={trailingIconWidth} />
```
devanshu052000: same as above avoid size changes

frontend/src/AppBuilder/Header/HeaderActions.jsx
```diff
@@ -67,9 +67,11 @@ const HeaderActions = function HeaderActions({ darkMode, showFullWidth, showPrev
         <ButtonComponent
           size="default"
           variant="outline"
-          leadingIcon="play01"
+          leadingIcon="play-lucide"
           data-cy="preview-link-button"
           style={{ padding: "7px 12px" }}
+          fill='var(--icon-default)'
+          leadingIconWidth='16'
```
devanshu052000: Instead of adding a new icon, would recommend using the support for lucide icons in the ButtonComponent of design system that way you won't even have to change sizes of icon since it's already configured.
simply pass a prop `isLucid=true` and in `leadingIcon` pass the lucide icon's name

PR #14415, opened 2025-10-27
Fix : Clicking on canvas when codehinter popover is open, don't update the properties panel

nakulnagargade:
Closes https://github.com/ToolJet/tj-ee/issues/4268

### DIFF
--- a/frontend/src/AppBuilder/_stores/slices/gridSlice.js
+++ b/frontend/src/AppBuilder/_stores/slices/gridSlice.js
@@ -479,13 +479,7 @@ export const createGridSlice = (set, get) => ({
     }));
   },
   handleCanvasContainerMouseUp: (e) => {
-    const {
-      selectedComponents,
-      clearSelectedComponents,
-      setActiveRightSideBarTab,
-      isRightSidebarOpen,
-      activeRightSideBarTab,
-    } = get();
+    const { clearSelectedComponents, setActiveRightSideBarTab, isRightSidebarOpen } = get();
     const selectedText = window.getSelection().toString();
     const isClickedOnSubcontainer =
       e.target.getAttribute('component-id') !== null && e.target.getAttribute('component-id') !== 'canvas';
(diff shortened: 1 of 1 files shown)


frontend/src/AppBuilder/_stores/slices/gridSlice.js
```diff
@@ -477,4 +478,46 @@ export const createGridSlice = (set, get) => ({
...
+    // If page settings tab is active and user clicks on canvas, switch to components tab
+    if (
+      !isClickedOnSubcontainer &&
+      ['rm-container', 'real-canvas', 'modal'].includes(e.target.id) &&
+      !selectedText &&
+      isRightSidebarOpen &&
+      activeRightSideBarTab === RIGHT_SIDE_BAR_TAB.PAGES
+    ) {
+      setActiveRightSideBarTab(RIGHT_SIDE_BAR_TAB.COMPONENTS);
+    }
```
devanshu052000: Can we combine and simplify these two conditions for better readability?

devanshu052000 (changes requested):

frontend/src/AppBuilder/_stores/slices/gridSlice.js
nakulnagargade: Yes, have removed the stale code.

frontend/src/AppBuilder/_stores/slices/gridSlice.js
```diff
@@ -477,4 +478,28 @@ export const createGridSlice = (set, get) => ({
...
+      ['rm-container', 'real-canvas', 'modal'].includes(e.target.id) &&
+      !selectedText &&
+      !isCodehinterPreviewOpen()
+    ) {
+      clearSelectedComponents();
+      if (isRightSidebarOpen) {
+        setActiveRightSideBarTab(RIGHT_SIDE_BAR_TAB.COMPONENTS);
+      }
+    }
+  },
```
devanshu052000, line 504: Based on the previous two conditions, to maintain backward compatibility shouldn't we call clearSelectedComponents() conditionally instead of directly as follows?
```
    if (
      selectedComponents.length &&
      !isCodehinterPreviewOpen()
    ) {
      clearSelectedComponents();
    },
```

PR #14516, opened 2025-11-06
Fix: Unable to signup at instance level

Siddharthpl:
Resolves - https://github.com/orgs/ToolJet/projects/28/views/14?filterQuery=&sliceBy%5Bvalue%5D=_noValue&pane=issue&itemId=136589525


server/src/modules/onboarding/service.ts
```diff
@@ -133,6 +133,9 @@ export class OnboardingService implements IOnboardingService {
         );
       } else {
         if (defaultWorkspace && !signingUpOrganization) {
+          if (!defaultWorkspace.enableSignUp) {
```
shah21, line 148: does CE have default workspace or is this another bug ?

Can you explain what is the issue here ?

server/src/modules/onboarding/service.ts
```diff
@@ -727,7 +730,7 @@ export class OnboardingService implements IOnboardingService {
       // Create first organization
       const workspaceSlug = generateWorkspaceSlug(workspaceName || 'My workspace');
       const organization = await this.setupOrganizationsUtilService.create(
-        { name: workspaceName || 'My workspace', slug: workspaceSlug },
+        { name: workspaceName || 'My workspace', slug: workspaceSlug, isDefault: true },
```
shah21, line 754: Since CE doesn't have default workspace concept. Please make sure this won't cause EE behaviour in CE.

server/src/modules/onboarding/service.ts
Siddharthpl: When a user tries to sign up at the instance level without specifying a workspace, according to the code, they are supposed to be created in a default workspace. However, in CE, there is no default workspace, which caused this issue.
So, to resolve this when the first user signs up, a workspace is created for them and assigned as the default. So if no workspace is provided during signup, the user is created in that first workspace. But in CE the default workspace cannot be changed as we does not have any concept of default workspace for now.

server/src/modules/onboarding/service.ts
Siddharthpl: Yess, It is not doing that.

PR #14551, opened 2025-11-11
Fix: MySQL SSL Toggle reject unauthorised connection

Pratush613:


plugins/packages/mysql/lib/index.ts
```diff
@@ -119,39 +119,44 @@ export default class MysqlQueryService implements QueryService {
...
-    const sslObject = { rejectUnauthorized: (sourceOptions.ssl_certificate ?? 'none') != 'none' };
+  const props = sourceOptions.socket_path
+    ? { socketPath: sourceOptions.socket_path }
+    : {
+        host: sourceOptions.host,
+        port: +sourceOptions.port,
+      };
+
+  const shouldUseSSL = !!sourceOptions.ssl_enabled;
+  let sslObject: any = undefined;
```
ganesh8056: use `null` as default value instead of undefined

plugins/packages/mysql/lib/index.ts
```diff
@@ -119,39 +119,44 @@ export default class MysqlQueryService implements QueryService {
...
-
-    const sslObject = { rejectUnauthorized: (sourceOptions.ssl_certificate ?? 'none') != 'none' };
+  const props = sourceOptions.socket_path
+    ? { socketPath: sourceOptions.socket_path }
+    : {
+        host: sourceOptions.host,
+        port: +sourceOptions.port,
+      };
+
+  const shouldUseSSL = !!sourceOptions.ssl_enabled;
```
ganesh8056: `ssl_enabled` will already be boolean datatype from `sourceOptions`

ganesh8056 (approved):

PR #14559, opened 2025-11-12
fix: If openapi datasource is created automatically when it does not exist and user does not adds specs definition then block that automatic datasource selection

NishidhJain:
chore: submodule update


NishidhJain:
Not required

PR #14612, opened 2025-11-18
Feat: Query level timeout for Postgresql datasource

ganesh8056:


server/src/modules/data-queries/util.service.ts
```diff
@@ -102,9 +107,27 @@ export class DataQueriesUtilService implements IDataQueriesUtilService {
         opts
       );
 
+      // Determine whether query timeout is set, to initiate abort controller
+      queryTimeoutMs =
+        typeof parsedQueryOptions['query_timeout'] === 'string' && parsedQueryOptions['query_timeout'].trim() === ''
```
akshaysasidrn, line 109: We would have to enforce all  plugin to use this key`query_timeout` onwards?

plugins/packages/common/lib/utils.helper.ts
```diff
@@ -271,3 +271,16 @@ export const redactHeaders = (headers) => {
...
 };
+
+export function getTooljetEdition(): string {
+  if (process.env.TOOLJET_EDITION) {
+    return process.env.TOOLJET_EDITION.toLowerCase();
+  }
+  return 'ce';
+}
+
+export function checkIfSignalIsAborted(signal: AbortSignal, timeout: number) {
```
akshaysasidrn: `throwIfSignalAborted`* ?

server/src/modules/data-queries/util.service.ts
```diff
@@ -73,6 +74,10 @@ export class DataQueriesUtilService implements IDataQueriesUtilService {
     let result;
     const queryStatus = new DataQueryStatus();
     const forwardRestCookies = this.configService.get<string>('FORWARD_RESTAPI_COOKIES') === 'true';
+    let canAbortQuery = false;
```
akshaysasidrn: Do we need this variable? Can instead check if abort Controller is defined instead?

akshaysasidrn: Or better yet try and encapsulate all this onto a class instead for abort control

akshaysasidrn: ⚠️ _Claude-generated_ 🤖 

**Interface:**
```typescript
export interface IQueryAbortSignal {
  readonly canAbort: boolean;
  readonly signal: AbortSignal | null;
  readonly timeoutMs: number;
  throwIfAborted(): void;
  isAborted(): boolean;
}

export interface QueryService {
  run(
    sourceOptions: object,
    queryOptions: object,
    dataSourceId?: string,
    dataSourceUpdatedAt?: string,
    context?: { user?: User; app?: App },
    abortSignal?: IQueryAbortSignal
  ): Promise<QueryResult>;
}
```

**Implementation:**
```typescript
class QueryAbortController implements IQueryAbortSignal {
  private controller: AbortController | null = null;
  private timeoutRef: NodeJS.Timeout | null = null;
  public readonly timeoutMs: number;
  public readonly canAbort: boolean;

  constructor(timeoutMs: number) {
    this.timeoutMs = timeoutMs;
    this.canAbort = Number.isFinite(timeoutMs) && timeoutMs > 0;
  }

  start(): void {
    if (!this.canAbort) return;
    this.cleanup();
    this.controller = new AbortController();
    this.timeoutRef = setTimeout(() => this.controller?.abort(), this.timeoutMs);
  }

  cleanup(): void {
    if (this.timeoutRef) {
      clearTimeout(this.timeoutRef);
      this.timeoutRef = null;
    }
  }

  get signal(): AbortSignal | null {
    return this.controller?.signal ?? null;
  }

  isAborted(): boolean {
    return this.controller?.signal?.aborted ?? false;
  }

  throwIfAborted(): void {
    if (this.canAbort && this.signal?.aborted) {
      throw new QueryError(
        'Query timed out',
        `Defined query timeout of ${this.timeoutMs}ms exceeded.`,
        {}
      );
    }
  }

  createAbortPromise(): Promise<never> {
    return new Promise((_, reject) => {
      if (!this.controller) return;
      const onAbort = () => reject(new QueryError(
        'Query timed out',
        `Defined query timeout of ${this.timeoutMs}ms exceeded.`,
        {}
      ));
      if (this.controller.signal.aborted) return onAbort();
      this.controller.signal.addEventListener('abort', onAbort, { once: true });
    });
  }
}
```

**Usage in `runQuery`:**
```typescript
async runQuery(...): Promise<object> {
  const timeoutMs = Number(parsedQueryOptions['query_timeout']) || NaN;
  const abortCtrl = new QueryAbortController(timeoutMs);

  try {
    abortCtrl.start();
    abortCtrl.throwIfAborted();

    const promises: Promise<any>[] = [
      service.run(sourceOptions, parsedQueryOptions, dataSourceId, updatedAt, context, abortCtrl)
    ];
    if (abortCtrl.canAbort) {
      promises.push(abortCtrl.createAbortPromise());
    }
    result = await Promise.race(promises);

    // OAuth retry
    if (needsOAuthRetry) {
      abortCtrl.start();
      // ... same pattern
    }

    return result;
  } finally {
    abortCtrl.cleanup();
  }
}
```

**Plugin usage:**
```typescript
// Native signal support (fetch, axios)
const response = await fetch(url, { signal: abortSignal?.signal });

// Cooperative cancellation
abortSignal?.throwIfAborted();
const result = await connection.query(query);
abortSignal?.throwIfAborted();
```

**Coverage of all existing abort cases:**

| Original | Proposed |
|----------|----------|
| Timeout parsing + `new AbortController()` + `setTimeout` | `constructor()` + `start()` |
| 4× `if (canAbortQuery) checkIfSignalIsAborted(...)` | `throwIfAborted()` |
| 2× duplicate abort promise creation (15 lines each) | `createAbortPromise()` |
| OAuth retry: re-create controller + timeout | `start()` (cleans up + re-creates) |
| 4× scattered `clearTimeout()` | Single `cleanup()` in `finally` |
| Event listener without `{ once: true }` | Uses `{ once: true }` |

akshaysasidrn: Maybe it will be better to limit optional arguments so that the plugin signature can remain the same. Wdyt?
```
export interface QueryService {
  run(
    sourceOptions: object,
    queryOptions: object,
    dataSourceId?: string,
    dataSourceUpdatedAt?: string,
    context?: { user?: User; app?: App, abortSignal?: IQueryAbortSignal }, // we can put it here too without modifying the signature until additional cases arise for query run.
  ): Promise<QueryResult>;
}
```

server/src/modules/data-queries/util.service.ts
ganesh8056: Yes the text field component, itself will ensure this variable name.

PR #14648, opened 2025-11-25
feat(otel): Add edition-based OpenTelemetry architecture

adishM98:
- EE server - https://github.com/ToolJet/ee-server/pull/312


server/src/otel/interfaces/IListener.ts
```diff
@@ -0,0 +1,23 @@
+import { INestApplication } from '@nestjs/common';
```
gsmithun4, line 1: Move to /interfaces/IListener.ts

PR #14666, opened 2025-11-27
Enhance : ClickHouse module updated

abhijeet760:
Replaced old clickhouse package (.query().toPromise()) with new official @clickhouse/client

Updated connection logic → createClient() now used instead of new ClickHouse()

Removed unused legacy configs (usePost, gzip, raw, trimQuery, etc.)

Modified query execution to use:

.query({query, format:'JSONEachRow'}).json() for SQL

client.insert({ table, values, columns, format:'JSONEachRow' }) for insert

Updated testConnection() to match new client behavior

Reduced required datasource fields → only host, port, protocol, username, password, database

empty result handling → returns { r:1 } when no rows exist


plugins/package.json
```diff
@@ -19,6 +19,7 @@
     "test": "NODE_ENV=test jest"
   },
   "dependencies": {
+    "@clickhouse/client": "^1.14.0",
```
ganesh8056: This module must be in ClickHouse plugins package.json

plugins/packages/clickhouse/package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "@tooljet-plugins/common": "file:../common",
     "clickhouse": "^2.6.0",
```
ganesh8056: We would need to remove this module. That's the main goal of this enhancement.

plugins/packages/clickhouse/lib/index.ts
```diff
@@ -11,13 +13,37 @@ export default class Click implements QueryService {
...
+              query,
+              format: format
+            });
+            
+            let data = await resultSet.json();
+            
+              result = data;
+            
+          } else {
+            await clickhouseClient.command({
```
ganesh8056, line 30: eventhough the command flow doesn't return the value. Similar to above method `resultSet` return the resultSet

plugins/packages/clickhouse/lib/index.ts
```diff
@@ -11,13 +13,37 @@ export default class Click implements QueryService {
...
-            .insert(`INSERT INTO ${tablename} (${fields.join(',')})`, this.parseJSON(query))
-            .toPromise();
+          const rows = this.parseJSON(query);
+          await clickhouseClient.insert({
+            table: tablename,
+            values: rows,
+            columns: fields,
+            format: 'JSONEachRow'
+          });
+          result = { r:1 };
```
ganesh8056: What this expression indicates`{ r:1 }`. Number of row inserted ?
If so we insert whether the insert operation can insert multiple rows ?

Or `{ r:1 }` the expression is given only for backward compatibility if so, add a code comment

ganesh8056 (changes requested):

plugins/packages/clickhouse/lib/index.ts
abhijeet760: {r:1} is only for backward compatibility. It was the output of the queries that didnot return anything in old clickhouse plugin.
Ok I will add a comment in the code as well.

plugins/packages/clickhouse/lib/index.ts
abhijeet760: Can you please provide a bit more information on the comment. I am unable to understand it.

PR #14688, opened 2025-12-02
fix: enhance ConfigHandle to prevent rendering in read-only Kanban/Listview subcontainers

manishkushare:
Resolves [Issue](https://github.com/orgs/ToolJet/projects/26/views/4?filterQuery=sprint%3A%40current+++assignee%3Amanishkushare&pane=issue&itemId=142519885&issue=ToolJet%7CToolJet%7C14687)

### DIFF
--- a/frontend/src/AppBuilder/AppCanvas/ConfigHandle/ConfigHandle.jsx
+++ b/frontend/src/AppBuilder/AppCanvas/ConfigHandle/ConfigHandle.jsx
@@ -38,11 +38,6 @@ export const ConfigHandle = ({
   const isModulesEnabled = useStore((state) => state.license.featureAccess?.modulesEnabled, shallow);
   const shouldFreeze = useStore((state) => state.getShouldFreeze());
   const componentName = useStore((state) => state.getComponentDefinition(id, moduleId)?.component?.name || '', shallow);
-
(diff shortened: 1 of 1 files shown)


frontend/src/AppBuilder/AppCanvas/ConfigHandle/ConfigHandle.jsx
```diff
@@ -37,6 +37,21 @@ export const ConfigHandle = ({
...
   const shouldFreeze = useStore((state) => state.getShouldFreeze());
   const componentName = useStore((state) => state.getComponentDefinition(id, moduleId)?.component?.name || '', shallow);
+
+  const parentComponentType = useStore((state) => {
+    const parentId = state.getComponentDefinition(id, moduleId)?.component?.parent;
+    if (!parentId) return null;
+    return state.getComponentDefinition(parentId, moduleId)?.component?.component;
+  }, shallow);
+
+  const isInReadOnlySubcontainer =
```
nakulnagargade: We have a prop i.e readOnly already doing the same thing in WidgetWrapper, we can use it as a prop in this component. This is to avoid the logic duplication.

johnsoncherian (approved):

PR #14690, opened 2025-12-02
feat: add show currency flag option

RupaakSrinivas:


server/src/modules/apps/services/app-import-export.service.ts
```diff
@@ -2769,7 +2771,15 @@ function migrateProperties(
         styles.labelStyle = { value: 'legacy' };
       }
     }
+
+    // CurrencyInput
+    if (componentType === 'CurrencyInput') {
+      if (properties.loadingState == undefined) {
+        properties.loadingState = { value: false };
+      }
+    }
```
shaurya-sharma064, line 2780: Condition should be for showFlag

johnsoncherian (approved):

PR #14814, opened 2025-12-15
[docs]: Update SendGrid Plugin

nayanan10:


RudraPratik30 (changes requested):
delete zone identifier

PR #14872, opened 2025-12-26
Promote and Release for the app in phase 3

Pratush613:
Promote and Relsease in Licensing for Phase 2


server/src/modules/licensing/configs/LicenseBase.ts
```diff
@@ -115,6 +117,8 @@ export default class LicenseBase {
     this._permissions = licenseData?.permissions;
     this._app = licenseData?.app;
     this._isCustomGroups = this.getPermissionValue('customGroups');
+    this._canPromote = licenseData?.app?.features?.promote ?? false;   
+    this._canRelease = licenseData?.app?.features?.release ?? false;
```
rudeUltra: can fetch directly from app object, no need to declare more variables

server/src/modules/licensing/configs/LicenseBase.ts
```diff
@@ -555,4 +561,19 @@ export default class LicenseBase {
     }
     return !!this._workflows?.['enabled'];
   }
+  public get canPromote(): boolean {
```
rudeUltra, line 606: Refer **appPermissionQuery** , consider older licenses where app and app.features.promote is empty,
should return true for older licenses (if not passed)

rudeUltra (dismissed):

PR #14875, opened 2025-12-26
list tables for mongo and postgres

abhijeet760:


server/src/modules/data-queries/util.service.ts
```diff
@@ -360,6 +360,51 @@ export class DataQueriesUtilService implements IDataQueriesUtilService {
...
+    app?: App
+  ): Promise<object> {
+    let result;
+    try {
+
+      if (!(dataSource)) {
+        throw new UnauthorizedException();
+      }
+
+      const organizationId = app?.organizationId || user?.organizationId;
```
gsmithun4: can remove app
const organizationId = user?.organizationId;
is enough

server/src/modules/data-queries/util.service.ts
```diff
@@ -360,6 +360,51 @@ export class DataQueriesUtilService implements IDataQueriesUtilService {
     }
   }
 
+  async listTables(
+    user: User,
+    dataSource: DataSource,
+    response: Response,
```
gsmithun4: con remove app and response

server/src/modules/data-queries/service.ts
```diff
@@ -257,6 +257,40 @@ export class DataQueriesService implements IDataQueriesService {
...
   }
 
+  async listTablesForApp(user: User, dataSource: DataSource, response: Response, app?: App) {
+    return this.listTablesAndGetResult(user, dataSource, response, app);
+  }
+
+  protected async listTablesAndGetResult(
+    user: User,
+    dataSource: DataSource,
+    response: Response,
```
gsmithun4: remove app and response

server/src/modules/data-queries/service.ts
```diff
@@ -257,6 +257,40 @@ export class DataQueriesService implements IDataQueriesService {
     return result;
   }
 
+  async listTablesForApp(user: User, dataSource: DataSource, response: Response, app?: App) {
+    return this.listTablesAndGetResult(user, dataSource, response, app);
+  }
+
+  protected async listTablesAndGetResult(
```
gsmithun4: no need to create separate function since its not reused
write everything under
`async listTablesForApp(user: User, dataSource:`

server/src/modules/data-queries/controller.ts
```diff
@@ -146,6 +146,18 @@ export class DataQueriesController implements IDataQueriesController {
     return this.dataQueriesService.runQueryForApp(user, dataQueryId, updateDataQueryDto, response, app);
   }
 
+  @InitFeature(FEATURE_KEY.LIST_TABLES)
+  @UseGuards(JwtAuthGuard, ValidateQuerySourceGuard, DataSourceFeatureAbilityGuard)
+  @Post(':dataSourceId/list-tables')
+  async listTables(
+    @User() user: UserEntity,
```
gsmithun4, line 153: remove app and response

server/src/modules/data-queries/controller.ts
```diff
@@ -146,6 +146,18 @@ export class DataQueriesController implements IDataQueriesController {
     return this.dataQueriesService.runQueryForApp(user, dataQueryId, updateDataQueryDto, response, app);
   }
 
+  @InitFeature(FEATURE_KEY.LIST_TABLES)
+  @UseGuards(JwtAuthGuard, ValidateQuerySourceGuard, DataSourceFeatureAbilityGuard)
+  @Post(':dataSourceId/list-tables')
```
gsmithun4: should be get

plugins/packages/postgresql/lib/index.ts
```diff
@@ -104,6 +104,33 @@ export default class PostgresqlQueryService implements QueryService {
...
+    dataSourceUpdatedAt: string
+  ): Promise<QueryResult> {
+    let knexInstance;
+    try {
+      knexInstance = await this.getConnection(sourceOptions, {}, true, dataSourceId, dataSourceUpdatedAt);
+
+      const { rows } = await knexInstance.raw(`
+      SELECT table_name 
+      FROM information_schema.tables 
+      WHERE table_schema = 'public'
```
ganesh8056, line 119: Schema can't be static. It can be different from 'public' as well

ganesh8056: We can Skip this

PR #14877, opened 2025-12-30
Remove email verfication for signup in selfhosted

Siddharthpl:
Issue - https://github.com/ToolJet/tj-ee/issues/4584
We are only removing email verification from signup in ee and ce

ee-server PR - https://github.com/ToolJet/ee-server/pull/346

### DIFF
--- a/cypress-tests/cypress/e2e/happyPath/platform/commonTestcases/apps/privateAndpublicApps.cy.js
+++ b/cypress-tests/cypress/e2e/happyPath/platform/commonTestcases/apps/privateAndpublicApps.cy.js
@@ -216,9 +216,6 @@ describe("Private and Public apps", () => {
     cy.clearAndType(onboardingSelectors.loginPasswordInput, "password");
     cy.get(commonSelectors.signUpButton).click();
     cy.wait("@signup");
-    cy.get(`[data-cy="resend-verification-email-button"]`).should("be.visible");
-    cy.wait(15000); // Waiting for mailhog to receive the email
-    fetchAndVisitInviteLinkViaMH(data.email);
     verifyWidget("private");
 
     // cy.apiLogout();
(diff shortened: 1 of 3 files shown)


server/src/modules/onboarding/util.service.ts
```diff
@@ -479,15 +521,57 @@ export class OnboardingUtilService implements IOnboardingUtilService {
...
-        this.eventEmitter.emit('emailEvent', {
-          type: EMAIL_EVENTS.SEND_WELCOME_EMAIL,
-          payload: {
-            to: user.email,
-            name: user.firstName,
-            invitationtoken: user.invitationToken,
-          },
-        });
+        // Auto-activate users for non-cloud editions (skip email verification)
+        const edition = getTooljetEdition();
```
shah21, line 525: nitpick - Can define this in one place and use mulitple places

server/src/modules/onboarding/util.service.ts
```diff
@@ -479,15 +521,57 @@ export class OnboardingUtilService implements IOnboardingUtilService {
...
+        const orgIdForValidation = user.defaultOrganizationId || organizationId;
+        if (orgIdForValidation) {
+          await this.licenseUserService.validateUser(manager, orgIdForValidation);
+        }
+
+        // For non-cloud editions, return login payload to auto-login user
+        if (!isCloudEdition && response) {
+          console.log('Attempting auto-login session generation for default org');
+          const userOrg = await this.organizationRepository.get(user.defaultOrganizationId);
+          if (!userOrg) console.log('WARNING: userOrg not found for defaultOrganizationId:', user.defaultOrganizationId);
```
shah21, line 550: Will this fail Login in this case?

Siddharthpl: Signup at the instance level for CE is currently unsupported in this PR. This remains a separate issue, as how it should be handled hasn't been finalized yet. There is a sperate PR for this issue in which i will handle this issue once it is confirmed.

PR #14885, opened 2025-12-31
Automation fix for environment permission

YuktiGoyal02:


cypress-tests/cypress/constants/selectors/manageGroups.js
```diff
@@ -178,6 +178,10 @@ export const groupsSelector = {
   modalMessage: '[data-cy="modal-message"]',
   changeRoleModalHeader: '[data-cy="modal-title"]:eq(1)',
   changeRoleModalMessage: '[data-cy="modal-body"]',
+  changeRoleModalIcon: '[data-cy="modal-icon"]',
```
Mekhla-Asopa: Can you please update it inside common selector because it is common for all model.

Mekhla-Asopa (approved):

PR #14954, opened 2026-01-11
REST API SSL VERIFICATION

abhijeet760:


abhijeet760:
If ssl certificate is none then self signed certificates are allowed even if SSL is enabled.
 if not none then existing flow should work as well.

PR #14995, opened 2026-01-18
changing gaurd.ts MODULE--->APP

Pratush613:
Corrected FeatureAbilityGuard to map APP_TYPES.MODULE to MODULES.APP instead of MODULES.MODULES.
This ensures MODULE apps load APP permissions, enabling environment access and fixing the 403 “restricted-preview” error.

### DIFF
--- a/server/src/modules/apps/ability/guard.ts
+++ b/server/src/modules/apps/ability/guard.ts
@@ -17,7 +17,7 @@ export class FeatureAbilityGuard extends AbilityGuard {
         };
       case APP_TYPES.MODULE:
         return {
-          resourceType: MODULES.APP,
+          resourceType: MODULES.MODULES,
         };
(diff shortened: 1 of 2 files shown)


server/src/modules/apps/ability/guard.ts
```diff
@@ -17,7 +17,7 @@ export class FeatureAbilityGuard extends AbilityGuard {
         };
       case APP_TYPES.MODULE:
         return {
-          resourceType: MODULES.MODULES,
+          resourceType: MODULES.APP,
```
vjaris42: Lets not modify guard directly, the correct approach would be to check the validate-private-app-access controller and check the service function to completely bypass environment check when app type is of modules

vjaris42 (changes requested):

vjaris42 (dismissed):

PR #15014, opened 2026-01-20
fix: prevent password reset with only whitespace characters (#14718)

sahil7303:
### Description
This PR fixes an issue where the "Reset Password" button was enabled even if the user entered only blank spaces (e.g., 5 spaces). The button will now remain disabled unless there is at least one non-whitespace character.

### Testing
1. Log in as an Admin and open the "All Users" page.
2. Click "Reset Password" for any user.
3. Type 5 blank spaces into the password field.
   - **Expected:** The "Reset" button remains **disabled**.


sahil7303:
Closing the PR in favour of new PR #15019

PR #15061, opened 2026-01-23
normalize(folder-apps): enforce single folder per app using created_at

Pratush613:
normalize(folder-apps): enforce single folder per app using created_at

### DIFF
--- a/server/src/modules/folder-apps/util.service.ts
+++ b/server/src/modules/folder-apps/util.service.ts
@@ -10,6 +10,7 @@ import { MODULES } from '@modules/app/constants/modules';
 import { UserAppsPermissions, UserWorkflowPermissions } from '@modules/ability/types';
 import { AbilityService } from '@modules/ability/interfaces/IService';
 import { APP_TYPES } from '@modules/apps/constants';
+import { AppGitSync } from '../../entities/app_git_sync.entity'
 
 @Injectable()
 export class FolderAppsUtilService implements IFolderAppsUtilService {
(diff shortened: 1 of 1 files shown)


server/src/modules/folder-apps/service.ts
```diff
@@ -22,7 +22,7 @@ export class FolderAppsService implements IFolderAppsService {
   async create(folderId: string, appId: string): Promise<FolderApp> {
     return dbTransactionWrap(async (manager: EntityManager) => {
       const existingFolderApp = await manager.findOne(FolderApp, {
-        where: { appId, folderId },
+        where: { appId },
       });
 
       if (existingFolderApp) {
```
gsmithun4: `App has already been added to a folder`

server/data-migrations/1769151383974-NormalizeFolderAppsKeepFirstCreatedMappingPerApp.ts
```diff
@@ -0,0 +1,30 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class NormalizeFolderAppsKeepFirstCreatedMappingPerApp1769151383974 implements MigrationInterface {
+
+public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
```
gsmithun4, line 6: please check this case - it should not fail if app is not added to any folders

Pratush613: If an app is not mapped to any folder, there will be no rows for that app_id in the folder_apps table, so the DELETE FROM folder_apps ... query will simply match zero rows and act as a no-op. SQL DELETE operations on empty result sets do not throw errors, and the CREATE UNIQUE INDEX statement does not require every app to have a row in folder_apps; it only enforces uniqueness on rows that actually exist.

PR #15063, opened 2026-01-23
Fixing the instance level signup issue

Siddharthpl:


server/src/helpers/utils.helper.ts
```diff
@@ -767,3 +768,25 @@ export async function validatePasswordDomain(
     instanceSettings?.PASSWORD_RESTRICTED_DOMAINS
   );
 }
+
+export async function getDefaultOrOldestWorkspaceOfInstance(
```
gsmithun4: since this function is used only in service.ts create it on
server/src/modules/onboarding/service.ts
and make it private

PR #15108, opened 2026-01-29
[docs]: Update How to

RudraPratik30:


docs/versioned_docs/version-3.16.0-LTS/widgets/table/bulk-row-operations.md
```diff
@@ -0,0 +1,156 @@
+---
+id: bulk-row-operations
+title: Bulk Row Operations
+---
+
+This guide explains how to perform bulk operations on table rows using row selection, JavaScript queries, and SQL. It assumes that a data source is already connected. PostgreSQL is used in the examples, but the same approach applies to other SQL databases.
```
RudraPratik30: Can repharse this para a bit, sounds to blunt, this part - `It assumes that a data source is already connected. PostgreSQL is used in the examples, but the same approach applies to other SQL databases.`

can do something like -
It is assumed that you have already configured a data source. We are going to use PostgreSQL.....

RudraPratik30: If SQL is necessary then do it like

It is assumed that you have already configured a SQL data source.

docs/versioned_docs/version-3.16.0-LTS/widgets/table/bulk-row-operations.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Prerequisites
```
RudraPratik30, line 9: can remove this part, just move SQL data source in the intro para

PR #15112, opened 2026-01-29
Rendered RowCount on the Footer of the Table Component

rahulprasad00:
Closes: https://github.com/ToolJet/tj-ee/issues/4677

In the Table Component Footer, added the display of number of records of the table.

### DIFF
--- a/frontend/src/AppBuilder/Widgets/NewTable/Table.jsx
+++ b/frontend/src/AppBuilder/Widgets/NewTable/Table.jsx
@@ -62,7 +62,6 @@ const Table = memo(
     const loadingState = useTableStore((state) => state.getLoadingState(id), shallow);
     const colorMode = getColorModeFromLuminance(containerBackgroundColor);
     const iconColor = getCssVarValue(document.documentElement, `var(--cc-default-icon-${colorMode})`);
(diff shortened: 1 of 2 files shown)


frontend/src/AppBuilder/Widgets/NewTable/_components/Footer/Footer.jsx
```diff
@@ -98,6 +98,7 @@ export const Footer = memo(
             {editedRows.size > 0 && showBulkUpdateActions && (
               <div className="d-flex custom-gap-4">{renderChangeSetUI()}</div>
             )}
+            {renderRowCount()}
```
devanshu052000: rowCount should only be visible if change set is not rendered
`<div className="d-flex custom-gap-4">{editedRows.size > 0 && showBulkUpdateActions ? renderChangeSetUI() : renderRowCount()}</div>`

devanshu052000 (changes requested):

frontend/src/AppBuilder/Widgets/NewTable/_components/Footer/Footer.jsx
rahulprasad00: resolved

PR #15139, opened 2026-02-02
Adding the password check in the invitaiton redirect sigin flow

Siddharthpl:
Adding the password check for the flow when
User already exists in an otganization.
User trying to signin through organization invite in another organization


server/src/modules/auth/service.ts
```diff
@@ -58,7 +59,8 @@ export class AuthService implements IAuthService {
       invitingOrganizationId = organizationId;
       /* give access to the default organization */
       user = await this.userRepository.findByEmail(email, organizationId, [WORKSPACE_USER_STATUS.INVITED]);
-      if (!user) {
+      const isPasswordMatching = await bcrypt.compare(password, user.password);
```
shah21:
```suggestion
      const isPasswordMatching = await bcrypt.compare(password, user?.password);
```

user.password might throw an error if the user is null ?

shah21 (changes requested):
Added a minor change @Siddharthpl. please check

shah21 (approved):

PR #15174, opened 2026-02-05
DO NOT MERGE: ssh tunneling

abhijeet760:


plugins/packages/mysql/lib/index.ts
```diff
@@ -10,6 +11,75 @@ import {
...
+
+ 
+    sshClient.connect({
+      host: sourceOptions.ssh_host,
+      port: sourceOptions.ssh_port || 22,
+      username: sourceOptions.ssh_username,
+      ...(sourceOptions.ssh_auth_type === 'password'
+        ? { password: sourceOptions.ssh_password }
+        : {
+            privateKey: fs.readFileSync('/home/abhijeet/projects/tooljet/plugins/packages/mysql/lib/testing_key.pem'),
```
rudeUltra, line 74: move this to .env or UI if it's in the Design

abhijeet760: will move it to ui, was only for testing.

Issue #14739, opened 2025-12-09
TJDB: Selecting table name creates three App History entries for a single action


ashmithamelita:
### Version Information
Version 3.20.50-ee-lts

### What is the expected behaviour?
Only one App History entry should be created when the user selects or changes a table name in a ToolJet DB query manager

### What is the current behaviour?
Three App History entries are created for a single table selection in the query manager editor.

### How to reproduce the issue?

1.Add the ToolJet database in the queries
2.Select the table name in the query manager
3.Open the network tab and check the entries 

### Screenshots or Screencast
https://jam.dev/c/5ff5c006-78b7-479e-9c3d-8aa7a9cd2be8

moyaqoob:
can I work on this issue?

Issue #14736, opened 2025-12-09
TJDB: In ToolJet DB List Query, Offset fails with Number without throwing error but works with String


ashmithamelita:
### Version Information
Version 3.20.50-cloud-lts

### What is the expected behaviour?
Offset should work identically regardless of input as Number or String

### What is the current behaviour?
Offset doesn't work with Number but doesn't show any error on entering a Number either
### How to reproduce the issue?

1.Add ToolJet DB as a data source and enter table name, operations and limit in the query manager
2.Set Offset to a number by passing in syntax format {{1}} 
3.Run the query and observe the table and preview in the query manager

### Screenshots or Screencast
https://jam.dev/c/d8eab6f8-686b-4266-abf5-57d03ebcb2d1

PR #15199, opened 2026-02-09
TJDB: In ToolJet DB List Query, Offset fails with Number without throwing error but works with String

abhijeet760:
Fix for #14736 and #14739

Issue:
https://github.com/ToolJet/ToolJet/issues/14736
https://github.com/ToolJet/ToolJet/issues/14739

Solution
Numbers have no enumerable properties, so isEmpty treats all numbers as empty.


frontend/src/AppBuilder/QueryManager/QueryEditors/TooljetDatabase/ToolJetDbOperations.jsx
```diff
@@ -497,6 +502,7 @@ const ToolJetDbOperations = ({
   };
 
   const handleTableNameSelect = (tableId) => {
+    isSelectingTable.current = true;
     setSelectedTableId(tableId);
     fetchTableInformation(tableId, true, tables);
```
rudeUltra: need to remove this extra call

abhijeet760: Removed the extra fetchTableInformation calls in handleNameSelect and fetchtables

PR #15211, opened 2026-02-10
add reload action to iframe widget configuration

manishkushare:
Resolves https://github.com/ToolJet/tj-ee/issues/4459


frontend/src/AppBuilder/Widgets/IFrame.jsx
```diff
@@ -108,7 +112,7 @@ export const IFrame = function IFrame({
         </div>
       ) : (
         <iframe
```
johnsoncherian, line 124: Can we use js native reload api instead of this? We can use ref to get the reference to the iframe element https://developer.mozilla.org/en-US/docs/Web/API/Location/reload

PR #15234, opened 2026-02-12
Fix: support custom primary keys in tooljet-db delete rows execution

### LINKED ISSUE #14799: TJDB: ToolJet DB Delete operation fails with custom primary key column, caching issue persists after rename without refresh
### Version Information
Version 3.20.52-ee-lts

Delete operation uses custom primary key column name from table creation, without defaulting to "id".

Delete operation defaults to non-existent "id" column instead of custom primary key name. Renaming column to "id" works correctly, but cached reference persists until app refresh.

1.Create a new TJDB table
2.Edit the id column name
3.Open the app and add TJDB in queries
4.Select the delete operation in GUI mode

### Screenshots or Screencast
https://jam.dev/c/78b5bed6-ccf0-492c-8c3b-53e0723552a4

sahil7303:
### Description

- Fixes ToolJet DB delete operation failing for tables with custom primary keys.

Fixes #14799


server/src/modules/tooljet-db/services/tooljet-db-data-operations.service.ts
```diff
@@ -254,7 +254,19 @@ export class TooljetDbDataOperationsService implements QueryService {
     }
 
     if (!isEmpty(whereQuery)) query.push(whereQuery);
-    if (limit && limit !== '') query.push(`limit=${limit}&order=id`);
+
+    if (limit && limit !== '') {
+      query.push(`limit=${limit}`);
+
+      const tableInfo = await this.tableOperationsService.perform(organizationId, 'view_table', { id: tableId });
```
rudeUltra: we are fetching it again, aren't we already storing the columns

PR #15240, opened 2026-02-12
Feat: App Version Deletion Audit Logs

RndmCodeGuy20:


server/src/modules/versions/service.ts
```diff
@@ -125,6 +125,18 @@ export class VersionService implements IVersionService {
   }
 
   async deleteVersion(app: App, user: User, manager?: EntityManager): Promise<void> {
+    RequestContext.setLocals(AUDIT_LOGS_REQUEST_CONTEXT_KEY, {
```
gsmithun4, line 130: should be added after the operation completes, it may throw error.

PR #15243, opened 2026-02-12
feat: add 'scroll to component' action with configurable behavior and alignment options.

nakulnagargade:
Closes https://github.com/ToolJet/tj-ee/issues/4686


frontend/src/AppBuilder/_stores/slices/eventsSlice.js
```diff
@@ -1281,6 +1301,22 @@ export const createEventsSlice = (set, get) => ({
         return executeAction(event, mode, {});
       };
 
+      const scrollComponentInToView = (componentName, behaviour, block, moduleId = 'canvas') => {
```
johnsoncherian: Should we club behaviour and block into single object, so the signature is 
```
const scrollComponentInToView = (
  componentName: string, 
  {behaviour: string, block: string}, 
  moduleId: string
)
```

nakulnagargade: Done

johnsoncherian (dismissed):

Issue #15266, opened 2026-02-16
Add require data-cy for Kanban component


Mekhla-Asopa:
Add data-cy for kanban component

- Button
- Fields
- Containers

PR #15267, opened 2026-02-16
Added data-cy for Kanban component

Mekhla-Asopa:
Fixes: https://github.com/ToolJet/ToolJet/issues/15266

### DIFF
--- a/frontend/src/AppBuilder/Widgets/Kanban/Components/Container.jsx
+++ b/frontend/src/AppBuilder/Widgets/Kanban/Components/Container.jsx
@@ -9,7 +9,7 @@ export const Container = ({ children, id, disabled, ...props }) => {
     id,
   });
 
-  const { kanbanProps, style, label, componentName } = props;
+  const { kanbanProps, style, label, dataCy } = props;
   const {
     styles: { accentColor },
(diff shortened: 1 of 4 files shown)


frontend/src/AppBuilder/Widgets/Kanban/Components/Container.jsx
```diff
@@ -36,15 +37,17 @@ export const Container = ({ children, id, disabled, ...props }) => {
       }}
       className={cx('kanban-container', 'scrollable', darkMode && 'dark')}
       onMouseDown={(e) => e.stopPropagation()}
+      data-cy={componentName ? `${componentName}-container-${generateCypressDataCy(label)}` : `${generateCypressDataCy(label)}-container`}
```
emidhun: I think if we are using dataCy we can skip these conditions.

emidhun (changes requested):

frontend/src/AppBuilder/Widgets/Kanban/Components/Container.jsx
Mekhla-Asopa: Yes, updated the same

PR #15284, opened 2026-02-18
Added Placeholder Text Color Style for Text input, dropdown,  number, Password components

rahulprasad00:
Solves: [Issue](https://github.com/ToolJet/tj-ee/issues/4706)

Added Placeholder Text color property for:
- Text input
- dropdown
- number
- Password

Also created the migration for it with progress counter for each iteration/batch. And updated the import-export service


frontend/src/AppBuilder/Widgets/BaseComponents/BaseInput.jsx
```diff
@@ -88,6 +89,7 @@ export const BaseInput = ({
         : 'var(--text-primary)',
     textOverflow: 'ellipsis',
     backgroundColor: 'inherit',
+    '--cc-placeholder-text': placeholderTextColor || 'var(--cc-placeholder-text)',
```
shaurya-sharma064: If placeholderTextColor is missing there is no need to set --cc-placeholder-text to itself again. We can simple remove it from styles.

frontend/src/AppBuilder/Widgets/DropdownV2/CustomMenuList.jsx
```diff
@@ -84,6 +84,7 @@ const CustomMenuList = ({ selectProps, ...props }) => {
             onFocus={onMenuInputFocus}
             placeholder="Search"
             className="dropdown-multiselect-widget-search-box"
+            style={{ '--cc-placeholder-text': placeholderTextColor || 'var(--cc-placeholder-text)' }}
```
shaurya-sharma064: Same here we can avoid setting the css variable back to itself.

frontend/src/AppBuilder/Widgets/BaseComponents/BaseInput.jsx
rahulprasad00: Resolved

shaurya-sharma064 (approved):

PR #15287, opened 2026-02-18
fix: prevent duplicate API calls on TJDB

sahil7303:
### Description
- Added  skipJoinTableUpdateRef to prevent secondary update loop..
- Merged table_id and join_table updates into a single call within handleTableNameSelect to prevent multiple parent re-renders.
- Added a  useEffect to sync tableInfo with the columns state, ensuring filter dropdowns populate correctly after metadata retrieval(removed fetchTableInformation logic and useEffect).

Review Changes:
- Used creatingQueryInProcessId as stable key to prevent component remount during tempId to realId transition.
- Removed unused functions.

### DIFF
--- a/frontend/src/AppBuilder/QueryManager/QueryEditors/TooljetDatabase/ToolJetDbOperations.jsx
+++ b/frontend/src/AppBuilder/QueryManager/QueryEditors/TooljetDatabase/ToolJetDbOperations.jsx
@@ -23,11 +23,6 @@ import config from 'config';
 import './styles.scss';
 import CodeHinter from '@/AppBuilder/CodeEditor';
 
-// Cache to prevent duplicate API calls during query creation
-let fetchTablesPromiseCache = null;
-let fetchTablesCacheTimestamp = null;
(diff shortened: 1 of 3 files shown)


frontend/src/AppBuilder/QueryManager/QueryEditors/TooljetDatabase/ToolJetDbOperations.jsx
```diff
@@ -191,12 +197,22 @@ const ToolJetDbOperations = ({
...
+    if (selectedTableId && tables.length > 0) {
+       const tableDetails = findTableDetails(selectedTableId);
+       if (tableDetails?.table_name && tableInfo[tableDetails.table_name]) {
+         setColumns(tableInfo[tableDetails.table_name]);
+       }
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedTableId, tableInfo, tables]);
+
+  // useEffect(() => {
```
rudeUltra: remove the comments

PR #15350, opened 2026-02-24
fix: Add missing guards to ToolJet DB endpoints

sahil7303:
### Changes:
Added authentication guards to :
1. POST /organizations/:organizationId/table/:tableName/bulk-upload
2. POST /organizations/:organizationId/join

### DIFF
--- a/server/src/modules/tooljet-db/controller.ts
+++ b/server/src/modules/tooljet-db/controller.ts
@@ -165,7 +165,7 @@ export class TooljetDbController {
   @InitFeature(FEATURE_KEY.JOIN_TABLES)
   @Post('/organizations/:organizationId/join')
   @UseFilters(new TooljetDbJoinExceptionFilter())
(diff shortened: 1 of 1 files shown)


server/src/modules/tooljet-db/controller.ts
```diff
@@ -164,7 +165,7 @@ export class TooljetDbController {
   @InitFeature(FEATURE_KEY.JOIN_TABLES)
   @Post('/organizations/:organizationId/join')
   @UseFilters(new TooljetDbJoinExceptionFilter())
-  @UseGuards(OrganizationAuthGuard, FeatureAbilityGuard)
+  @UseGuards(JwtAuthGuard, OrganizationAuthGuard, FeatureAbilityGuard)
```
gsmithun4: OrganizationAuthGuard is enough same extends to JwtAuthGuard

PR #15374, opened 2026-02-26
Query preview shifted to the bottom 

shaurya-sharma064:
closes: https://github.com/ToolJet/tj-ee/issues/4757

### DIFF
--- a/frontend/src/AppBuilder/CodeEditor/PreviewBox.jsx
+++ b/frontend/src/AppBuilder/CodeEditor/PreviewBox.jsx
@@ -593,7 +593,7 @@ const PreviewContainer = ({
           <Card
             className={cx({
               'bg-slate2': darkMode,
-              'query-manger-input-preview-popover-card': isInsideQueryManager,
+              'query-manager-input-preview-popover-card': isInsideQueryManager,
             })}
             style={{
               borderColor: errorStateActive ? 'var(--tomato8)' : 'var(--slate6)',
--- a/frontend/src/AppBuilder/CodeEditor/SingleLineCodeEditor.jsx
+++ b/frontend/src/AppBuilder/CodeEditor/SingleLineCodeEditor.jsx
@@ -84,7 +84,7 @@ const SingleLineCodeEditor = ({ componentName, fieldMeta = {}, componentId, ...r
   );
 
   const [wrapperWidth, setWrapperWidth] = useState(0);
-  const [wrapperHeight, setWrapperHeight] = useState(0);
+  const [_wrapperHeight, setWrapperHeight] = useState(0);
   const [overlayKey, setOverlayKey] = useState(0);
 
   useEffect(() => {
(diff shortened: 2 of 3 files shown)


frontend/src/AppBuilder/CodeEditor/SingleLineCodeEditor.jsx
```diff
@@ -78,6 +78,31 @@ const SingleLineCodeEditor = ({ componentName, fieldMeta = {}, componentId, ...r
...
+      setWrapperWidth(wrapperRef.current?.clientWidth || 0);
+      const newHeight = wrapperRef.current?.clientHeight || 0;
+      if (newHeight !== wrapperHeight) {
+        setWrapperHeight(newHeight);
+        setOverlayKey((prev) => prev + 1); // force remount Overlay
+      }
+    });
+    observer.observe(wrapperRef.current);
+    return () => observer.disconnect();
+  }, [wrapperRef, wrapperHeight, isInsideQueryManager]);
```
kavinvenkatachalam:
```suggestion
  useEffect(() => {
      if (!wrapperRef.current || !isInsideQueryManager) return;
      setWrapperWidth(wrapperRef.current.clientWidth);
      const observer = new ResizeObserver((entries) => {
        const el = entries[0]?.target;
        if (!el) return;
        setWrapperWidth(el.clientWidth);
        setWrapperHeight((prev) => {
          if (el.clientHeight !== prev) {
            setOverlayKey((k) => k + 1);
            return el.clientHeight;
          }
          return prev;
        });
      });
      observer.observe(wrapperRef.current);
      return () => observer.disconnect();
    }, [isInsideQueryManager]);
```

The ResizeObserver callback captures `wrapperHeight` from the closure, but `wrapperHeight` is in the dependency array, causing the effect to tear down and re-create the observer on every height change. This defeats the purpose of using a ResizeObserver.

frontend/src/AppBuilder/CodeEditor/styles.scss
```diff
@@ -1,5 +1,12 @@
 @import "../../_styles/colors.scss";
 
+.popover-body {
+    .query-manger-input-preview-popover-card{
+        border: 1px solid var(--border-success-weak) !important;
+        box-shadow: var(--elevation-300-box-shadow) !important;
```
kavinvenkatachalam, line 6: Can we increase the specificity if possible instead of `!important`?

frontend/src/AppBuilder/CodeEditor/styles.scss
```diff
@@ -1,5 +1,12 @@
 @import "../../_styles/colors.scss";
 
+.popover-body {
+    .query-manger-input-preview-popover-card{
```
kavinvenkatachalam: Typo in the class name. It should be `manager`

frontend/src/AppBuilder/CodeEditor/PreviewBox.jsx
```diff
@@ -616,6 +633,7 @@ const PreviewContainer = ({
             modifiers: [
               {
                 name: 'flip',
+                enabled: isInsideQueryManager,
```
kavinvenkatachalam, line 636: Can you check this? Earlier it was `true` by default. Now, it will be true only if is inside query manager and rest will be disabled

shaurya-sharma064: Wasnt working in any scenario previously from what i saw.

frontend/src/AppBuilder/CodeEditor/styles.scss
shaurya-sharma064: card.scss has .card border set to 0px as !important. Cannot override this with specificity.

kavinvenkatachalam (approved):

PR #15396, opened 2026-03-02
Helper text for required fields

abhijeet760:


frontend/src/modules/dataSources/components/DataSourceManager/DataSourceManager.jsx
```diff
@@ -962,6 +984,11 @@ class DataSourceManagerComponent extends React.Component {
       !isEmpty(validationMessages)
       : true;
     this.props.setGlobalDataSourceStatus({ isEditing: !isSaveDisabled });
+    const hintKind = selectedDataSource?.kind;
+    const hintManifest = hintKind ? getManifestByKind(hintKind) : null;
+    const showSaveHint = !isSaving && !this.props.isVersionReleased &&
```
rudeUltra, line 989: isVersionReleased takes care of already saved Datasources right ? so if i change just 1 field the error should not come up and only if the user has entered a value then the error should come up not on load

PR #15440, opened 2026-03-05
list-tables for OPENAPI and pagination and search with list tables for Postgres, mongodb,mssql,mysql

abhijeet760:
Feature reuested in #15447
OpenAPI: Added listTables() to return all endpoints (path, method, summary, operationId) parsed from the stored spec.

PostgreSQL: Extended listTables() to support schema (default public), search (ILIKE filter), and optional page/limit pagination. Response now includes table_schema alongside table_name.

MongoDb: listTables extended with pagination and search. ListCollectionsCursor does not support .skip()/.limit(), fetch all matching then slice

mssql and mysql: Extended listTables() with optional page/limit pagination


plugins/packages/mongodb/lib/index.ts
```diff
@@ -271,15 +271,31 @@ export default class MongodbService implements QueryService {
     };
   }
 
-  async listCollections(
+  async listTables(
```
rudeUltra, line 275: Check for backward compatibility if we have to rename the function

plugins/packages/postgresql/lib/index.ts
```diff
@@ -290,10 +322,79 @@ export default class PostgresqlQueryService implements QueryService {
...
+          SELECT COUNT(*) AS total
+          FROM information_schema.tables
+          WHERE table_type = 'BASE TABLE'
+          AND table_schema NOT IN ('information_schema', 'pg_catalog', 'pg_toast')
+          ${allSchemas ? '' : 'AND table_schema = ?'}
+          AND table_name ILIKE ?
+        `;
+        const countParams: any[] = allSchemas ? [searchPattern] : [schema, searchPattern];
+
+        const { rows } = await knexInstance.raw(query, params);
```
rudeUltra: try to Use promise .all same as mysql

plugins/packages/postgresql/lib/index.ts
```diff
@@ -349,14 +453,34 @@ export default class PostgresqlQueryService implements QueryService {
...
-      );
+      const searchPattern = `%${search}%`;
+      const params: any[] = [schema, searchPattern];
+
+      let query = `SELECT table_name FROM information_schema.tables
+                  WHERE table_schema = ? AND table_type = 'BASE TABLE'
+                  AND table_name ILIKE ?
+                  ORDER BY table_name`;
+
+      if (limit) {
```
rudeUltra, line 466: postgres supports limit and OFFset , directly use

server/src/modules/data-queries/service.ts
```diff
@@ -308,11 +308,10 @@ export class DataQueriesService implements IDataQueriesService {
     return result;
   }
 
-  async listTablesForApp(user: User, dataSource: DataSource, environmentId: string) {
+  async listTablesForApp(user: User, dataSource: DataSource, environmentId: string, schema?: string, search?: string, page?: number, limit?: number) {
```
rudeUltra: pass the extra fields in an optional object param , it's getting crowded, use DTO for validation as well

plugins/packages/postgresql/lib/index.ts
abhijeet760: Using them directly only just adding the limits and offset if available in queryoptions

server/src/modules/data-queries/service.ts
abhijeet760: Ok

plugins/packages/mongodb/lib/index.ts
abhijeet760: Its name was earlier listtables but was changed toh listcollections in db improvements pr.So i changed it again.

plugins/packages/postgresql/lib/index.ts
abhijeet760: Ok

Ammynaka1212:
> Feature reuested in #15447
> OpenAPI: Added listTables() to return all endpoints (path, method, summary, operationId) parsed from the stored spec.
> 
> PostgreSQL: Extended listTables() to support schema (default public), search (ILIKE filter), and optional page/limit pagination. Response now includes table_schema alongside table_name.
> 
> MongoDb: listTables extended with pagination and search. ListCollectionsCursor does not support .skip()/.limit(), fetch all matching then slice
> 
> mssql: Extended listTables() with optional page/limit pagination

Ammynaka1212:
> Feature reuested in #15447
> OpenAPI: Added listTables() to return all endpoints (path, method, summary, operationId) parsed from the stored spec.
> 
> PostgreSQL: Extended listTables() to support schema (default public), search (ILIKE filter), and optional page/limit pagination. Response now includes table_schema alongside table_name.
> 
> MongoDb: listTables extended with pagination and search. ListCollectionsCursor does not support .skip()/.limit(), fetch all matching then slice
> 
> mssql: Extended listTables() with optional page/limit pagination

PR #15456, opened 2026-03-06
fix:Uniqueness-of-data-source

Pratush613:
Prevents duplicate data source names within the same branch
Ensures database-level consistency via a unique constraint
Safely migrates existing duplicate records without data loss


server/src/modules/data-sources/util.service.ts
```diff
@@ -1148,4 +1154,85 @@ export class DataSourcesUtilService implements IDataSourcesUtilService {
...
+     manager: EntityManager
+ ): Promise<void> {
+   const qb = manager
+     .createQueryBuilder(DataSourceVersion, 'dsv')
+     .where('LOWER(dsv.name) = LOWER(:name)', { name })
+     .andWhere('dsv.isActive = true')
+     .andWhere('dsv.id != :currentDsvId', { currentDsvId });
+
+
+   if (branchId) {
```
gsmithun4: for update we have currentDsvId, so no need to add where clause

server/src/modules/data-sources/util.service.ts
```diff
@@ -1148,4 +1154,85 @@ export class DataSourcesUtilService implements IDataSourcesUtilService {
...
+   const qb = manager
+     .createQueryBuilder(DataSourceVersion, 'dsv')
+     .where('dsv.isActive = true')
+     .andWhere('dsv.name LIKE :name', { name: `${escapedBase}%` });
+
+
+   if (branchId) {
+     qb.andWhere('dsv.branchId = :branchId', { branchId });
+   } else {
+     qb.andWhere('dsv.branchId IS NULL');
```
gsmithun4: If branch id is not there, then git is disabled.

so can add this
dsv.is_default should be true

vjaris42 (approved):

PR #15506, opened 2026-03-10
feat: Creates a file button widget for appbuilder

nithindavid:
fixes - https://github.com/ToolJet/tj-ee/issues/4723
Figma - https://www.figma.com/design/UUIFeymC32fkJZiY5F2VTe/02-Input-components?node-id=1620-15247&t=r6JRodzCIoVGx5NA-0

image

image
image


server/src/modules/apps/services/app-import-export.service.ts
```diff
@@ -99,6 +99,7 @@ type NewRevampedComponent =
   | 'PhoneInput'
   | 'IFrame'
   | 'DropdownV2'
+  | 'FileButton'
```
shaurya-sharma064: New component no need to add in import export.

johnsoncherian (approved):

PR #15599, opened 2026-03-18
Oauth Token encryption

sahil7303:
#### Description
- Encrypt and save OAuth credentials in the new `datasource_user_token_data` entity.
- Inject decrypted tokens into data source options during query execution.
- Clear all associated token records when toggling between single and multi-auth.

ee-server PR: https://github.com/ToolJet/ee-server/pull/415

### DIFF
--- a/server/src/modules/data-sources/util.service.ts
+++ b/server/src/modules/data-sources/util.service.ts
@@ -678,24 +678,11 @@ export class DataSourcesUtilService implements IDataSourcesUtilService {
       refreshToken = newToken['refresh_token'] ?? null;
     }
 
-    // Persist tokens to datasource_user_token_data for all relevant environments
+    // Persist tokens to datasource_user_token_data
     await dbTransactionWrap(async (manager: EntityManager) => {
-      const isMultiEnvEnabled = await this.licenseTermsService.getLicenseTerms(
-        LICENSE_FIELD.MULTI_ENVIRONMENT,
-        organizationId
-      );
       const tokenUserId = isMultiAuthEnabled ? userId : null;
-
-      if (isMultiEnvEnabled) {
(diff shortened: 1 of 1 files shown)


server/src/modules/data-sources/util.service.ts
```diff
@@ -907,33 +999,103 @@ export class DataSourcesUtilService implements IDataSourcesUtilService {
...
+    manager: EntityManager
+  ): Promise<void> {
+    const encryptedAccessToken = accessToken
+      ? await this.encryptionService.encryptColumnValue('credentials', 'value', accessToken)
+      : null;
+    const encryptedRefreshToken = refreshToken
+      ? await this.encryptionService.encryptColumnValue('credentials', 'value', refreshToken)
+      : null;
+
+    // DELETE + INSERT pattern to handle upsert with nullable user_id
```
rudeUltra: No need to delete and Insert, If row is there then update or Insert based on single/multi Auth

server/src/modules/data-sources/util.service.ts
```diff
@@ -585,41 +660,43 @@ export class DataSourcesUtilService implements IDataSourcesUtilService {
...
-          key: 'tokenData',
-          value: tokenData,
-          encrypted: false,
-        },
-      ];
-    }
-    await this.updateOptions(dataSource.id, tokenOptions, organizationId, environmentId);
-    return;
+      if (isMultiEnvEnabled) {
+        const dso = await this.appEnvironmentUtilService.getOptions(dataSource.id, organizationId, environmentId);
```
rudeUltra: Was there a license check before ?
Just keep existing logic, no need to handle each env, i think that is already handled

PR #15626, opened 2026-03-19
Fix: resolve double encoding in connection strings

### LINKED ISSUE #15397: [MongoDB]- Need factor for encoding the values passed in connection string
### Environment
LTS-3.16

Need to encode the values passed in connection string 

[Connection string]When using %40 (URL-encoded value for @) in the MongoDB password field, it is not being decoded back to @ before authentication. As a result, the system is attempting authentication with the literal value Test%40123 instead of Test@123, which causes an authentication failure.
The password field should either:

Automatically decode URL-encoded characters before use, or
Treat the password input as raw text and avoid double encoding when constructing the connection string.
Currently, %40 is not being converted back to @, leading to incorrect credential handling.

### Screenshots or Screencast

Image

sahil7303:
### Description:
- Safely decode MongoDB credentials to handle edge cases.
- Decode connection string credentials before populating frontend UI fields.

Related issue:
Fixes #15397

### DIFF
--- a/plugins/packages/mongodb/lib/index.ts
+++ b/plugins/packages/mongodb/lib/index.ts
@@ -117,11 +117,11 @@ function validateSSHOptions(sourceOptions: SourceOptions): void {
   }
 }
 
-function safeDecodeAndEncode(value: string): string {
+function safeDecode(value: string): string {
   try {
-    return encodeURIComponent(decodeURIComponent(value));
+    return decodeURIComponent(value);
   } catch {
-    return encodeURIComponent(value);
+    return value;
   }
 }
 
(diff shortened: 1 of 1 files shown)


plugins/packages/mongodb/lib/index.ts
```diff
@@ -484,8 +492,8 @@ async getConnection(sourceOptions: SourceOptions): Promise<any> {
 
   if (authPart.includes(":")) {
     const colonIdx = authPart.indexOf(":");
-    connUser = decodeURIComponent(authPart.slice(0, colonIdx));
-    connPass = decodeURIComponent(authPart.slice(colonIdx + 1));
+    connUser = authPart.slice(0, colonIdx);
```
rudeUltra: decode not required here ?

PR #15900, opened 2026-04-13
Fix: bundle Redis in CE Docker image and fix oracledb plugin lockfile

adishM98:
Problem

  Two issues were blocking CE single-instance deployments:

  1. CE image failed to boot with ECONNREFUSED 127.0.0.1:6379 — BullMQ and CustomDomainCacheService both connect to Redis on startup. CE docs only mention external
  Redis for multi-pod setups, but single-instance CE also requires it. Rather than requiring operators to run a Redis sidecar, we now bundle it in the image.
  2. Docker build failed during lerna run build — the oracle wallet PR (#15771) added unzipper as a dependency to @tooljet-plugins/oracledb but didn't update
  plugins/package-lock.json. The Dockerfile installs from the lockfile, so unzipper was never installed and tsc failed with a missing module error.

  Changes

  docker/ce-production.Dockerfile

  - Added redis-server to the apt-get install step
  - Created /var/lib/redis and /var/log/redis dirs with correct appuser ownership — merged into the useradd block so appuser exists before chown runs
  - Wrote /app/redis.conf (localhost-only, daemonized) under the appuser-owned /app path to avoid /etc/redis permission issues

  docker/ce-entrypoint.sh

  - Starts redis-server (daemonized via conf) at container boot before DB wait/migrations

  plugins/package-lock.json

  - Regenerated to include unzipper@0.12.3 and its transitive deps (duplexer2, bluebird)


ganesh8056 (dismissed):
Approved for Plugins changes

PR #15918, opened 2026-04-14
Feature : Snowflake plugin improvements

### LINKED ISSUE #12765: Snowflakes: Forward driver errors in standard error structure's data key

### LINKED ISSUE #14454: Snowflake: In the multi-user refresh token flow, when the toggle is enabled and the token expires in the default workspace, the query manager throws an error. However, for the same data source, the token for the invited user does not expire.
### Version Information
Version 3.20.26-ee-lts

When the token expires in the default workspace, it should also expire for the invited user to maintain consistent behaviour across all users.

When the token expires in the default workspace, the query manager throws an error. However, for the same data source, the invited user’s token remains active and does not expire.

1.Navigate the datasource, add the valid credential to the datasource
2.Enable the toggle, once authentication succeeds save the data source
3.Add the datasource in app and run the query in query manager
4.Invite user and accept the invite and open the same workspace
5.Click the app and run the same query once the authentication succeeds run the query once again in query manager
6. Wait for 1 hour for the token to expire

### Screenshots or Screencast
https://jam.dev/c/f49b5fcd-7edf-4516-9d8e-a7ea018cf3c5

### LINKED ISSUE #15457: Snowflake implementing the depreciation of user name and password for Service Accounts
1. bearer token flow
2. key pair

### LINKED ISSUE #16006: Snowflake datasource : Dynamic data connection
While running a Snowflake query, the following options can be added dynamically.

1. Database
2. Warehouse
3. role

### LINKED ISSUE #16007: Snowflake datasource : GUI mode for Querying
**Supported Operations in GUI Mode**

1. List rows.
2. Insert record.
3. Update.
4. Upsert.
5. Delete.
6. Bulk Insert.
7. Bulk update using primary key.
8. Bulk upsert using primary key.

ganesh8056:
Enhancements 

1. Refresh token flow for Oauth.
2. Dynamic connection parameters.
3. Snowflake as most commonly used plugin.
4. Error standardisation for run query and test connection.
5. GUI mode ( Insert, Update, Upsert, Delete, Bulk Insert, Bulk Update, Bulk Upsert ).
6. New Authentication mechanism ( Key pair, Bearer token )


plugins/packages/common/lib/queryBuilder.ts
```diff
@@ -763,6 +816,65 @@ export class QueryBuilder {
...
+    this._reset();
+
+    const primaryKeySet = new Set(primaryKey);
+    const allPkValuesAbsent = primaryKey.every((pk) => {
+      const value = row[pk];
+      return value === undefined || value === null || value === '';
+    });
+
+    // Rule 3: PK column values not provided in input → INSERT the records
+    if (allPkValuesAbsent) {
```
rudeUltra, line 843: Scenario A: Both PKs are empty strings
Scenario B: One PK is provided, one is empty string

if your table has a row with org_id='ORG_1', user_id='', Scenario B would match and update it. But if you then clear org_id too, Scenario A would insert a brand new row instead of matching. The behavior flips based on an unrelated column.

PR #15993, opened 2026-04-20
Dynamic height 

shaurya-sharma064:


frontend/src/AppBuilder/Widgets/TextArea.jsx
```diff
@@ -28,8 +28,12 @@ export const TextArea = (props) => {
       return;
     }
     inputRef.current.style.height = 'auto';
+    // `scrollHeight` = content + padding. We add just the 2px top/bottom
+    // border so the outer box doesn't clip and trigger an unwanted scrollbar.
+    // Previously +20 was used here, which padded every grown textarea with
+    // an extra ~20px of empty space — visible as unexplained vertical slack.
     inputRef.current.style.height =
-      height >= inputRef.current.scrollHeight ? `${height}px` : `${inputRef.current.scrollHeight + 20}px`;
+      height >= inputRef.current.scrollHeight ? `${height}px` : `${inputRef.current.scrollHeight + 2}px`;
```
kavinvenkatachalam: Won't this cause an existing components to shrink?

frontend/src/AppBuilder/AppCanvas/RenderWidget.jsx
```diff
@@ -244,6 +245,25 @@ const RenderWidget = ({
...
+    if (exposed !== undefined) return exposed;
+    return resolved?.properties?.visibility ?? resolved?.styles?.visibility ?? true;
+  }, shallow);
+  useEffect(() => {
+    if (currentMode !== 'view') return;
+    if (!collapseWhenHidden) return;
+    const contextIndices = normalizeLayoutContext(resolveIndex);
+    const handle = requestAnimationFrame(() => {
+      adjustComponentPositions(id, currentLayout, false, contextIndices);
+    });
```
kavinvenkatachalam, line 262: This is called for each component. Can you check whether this can be debounced and update the store in a single go?

kavinvenkatachalam (changes requested):

frontend/src/AppBuilder/Widgets/TextArea.jsx
shaurya-sharma064: Have removed the +2 as well now. It wasn't working properly before since even with no content min height was 20px larger than what it should have been.

frontend/src/AppBuilder/AppCanvas/RenderWidget.jsx
shaurya-sharma064: 1) Updating store in single go would not be possible as each state is dependent on the other.
2) It is called for each component as long as the toggle is true

johnsoncherian (approved):

PR #16034, opened 2026-04-22
Fix/module version rename ref update

Pratush613:


vjaris42 (approved):

server/src/modules/versions/service.ts
```diff
@@ -355,6 +355,22 @@ export class VersionService implements IVersionService {
           );
         }
       }
+      // When a module version is renamed, update all consumer ModuleViewer refs that were
+      // pinned to the old name so they continue to resolve to the correct version instead
+      // of falling through to the active-draft fallback.
+      if (
+        app.type === APP_TYPES.MODULE &&
+        appVersionUpdateDto?.name &&
+        appVersion.name !== appVersionUpdateDto.name
```
akshaysasidrn, line 365: should we also add check? `appVersion.versionType === AppVersionType.VERSION`

akshaysasidrn (approved):

PR #16038, opened 2026-04-22
Fix: use sessionStorage for active branch to enable per-tab branch isolation

RndmCodeGuy20:
## Summary
- Migrates active branch persistence from `localStorage` to `sessionStorage` so each tab maintains its own isolated branch state, enabling users to work on different branches simultaneously across tabs
- Adds `localStorage` as a read fallback with sessionStorage hydration — new tabs opened from a branch-specific context (e.g. opening an app link in a new tab) inherit the active branch instead of falling to main/master and hard-failing with a home page redirect
- Dual-writes to both storages on every `setActiveBranch` call so `localStorage` stays current as the cross-tab seed
- Clears `sessionStorage` on logout alongside the existing `localStorage.clear()` to prevent stale branch resumption on re-login within the same tab

## Files changed
- `frontend/src/_helpers/active-branch.js` — core storage logic
- `frontend/src/AppBuilder/_utils/auth.js` — logout clears sessionStorage
- `frontend/src/_stores/workspaceBranchesStore.js` — stale comments updated

## Test plan
- [ ] Single tab: switch to non-default branch, F5 — should stay on same branch
- [ ] New tab: open an app link while on a non-default branch — should open on that branch, not redirect to home
- [ ] Multi-tab isolation: Tab A on branchA, Tab B on branchB — switching in Tab B must not affect Tab A
- [ ] New tab after multi-tab: Tab A on branchA, Tab B switches to branchB, open new Tab C — should land on a real branch, no hard failure
- [ ] Logout: be on a branch, logout, re-login in same tab — should land on default branch, not previous branch
- [ ] Corrupted localStorage: manually set `tj_active_branch_{orgId}` to `{"garbage":true}`, open new tab — should fall to default branch
- [ ] Incognito / no storage: open app URL — should land on server default with no errors
- [ ] Deleted branch: be on a branch, delete it from another tab, refresh — should self-heal to default branch

### DIFF
--- a/frontend/src/_helpers/active-branch.js
+++ b/frontend/src/_helpers/active-branch.js
@@ -10,8 +10,16 @@ export function getActiveBranch(orgId) {
   const id = orgId || getOrgId();
   if (!id) return null;
   try {
-    const stored = localStorage.getItem(`${BRANCH_KEY_PREFIX}${id}`);
-    return stored ? JSON.parse(stored) : null;
+    const key = `${BRANCH_KEY_PREFIX}${id}`;
+    const sessionStored = sessionStorage.getItem(key);
+    if (sessionStored) return JSON.parse(sessionStored);
+    // Seed sessionStorage from localStorage so new tabs inherit the active branch
+    const localStored = localStorage.getItem(key);
+    if (localStored) {
+      sessionStorage.setItem(key, localStored);
+      return JSON.parse(localStored);
+    }
+    return null;
   } catch {
     return null;
   }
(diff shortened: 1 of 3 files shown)


frontend/src/_helpers/active-branch.js
NishidhJain: Please check for this case as well, might need to removeItem from sessionStorage

NishidhJain (approved):

kavinvenkatachalam (approved):

frontend/src/_helpers/active-branch.js
```diff
@@ -36,6 +48,45 @@ export function getActiveBranchId(orgId) {
...
+// Module-level ref ensures only one listener is registered at a time
+let _focusSyncListener = null;
+
+/**
+ * Registers a visibilitychange listener that writes the current tab's branch
+ * from sessionStorage to localStorage whenever the tab becomes visible.
+ * This ensures new tabs opened from this tab inherit its active branch.
+ * Safe to call multiple times — removes any existing listener before re-registering.
+ */
+export function registerBranchFocusSync() {
```
NishidhJain, line 60: This approach also works, but in my opinion, better approach would have been the reactive (declarative) approach like using useEffect in a component which appears whenever branching is enabled, so we don't have to manually add & remove listener at multiple places. What do you think @kavinvenkatachalam?

RndmCodeGuy20: which component should we add this useEffect in? is there any shell/provider component where we can add this?

NishidhJain: I think the component which shows the branch button, because it should be visible whenever branching is enabled

kavinvenkatachalam (approved):

PR #16324, opened 2026-05-07
Fix/listtables

abhijeet760:


plugins/packages/mssql/lib/index.ts
```diff
@@ -696,13 +696,15 @@ export default class MssqlQueryService implements QueryService {
 
   async listTables(
     sourceOptions: SourceOptions,
+    dataSourceId?: string,
```
rudeUltra, line 699: Are these being used ?

plugins/packages/mysql/lib/index.ts
```diff
@@ -773,14 +773,15 @@ export default class MysqlQueryService implements QueryService {
   }
   async listTables(
     sourceOptions: SourceOptions,
+    dataSourceId?: string,
```
rudeUltra, line 776: same

plugins/packages/mssql/lib/index.ts
abhijeet760: no

plugins/packages/mysql/lib/index.ts
abhijeet760: no

plugins/packages/mssql/lib/index.ts
abhijeet760: but are being sent from util.service

PR #16371, opened 2026-05-11
[feat]: Add Flex component

nakulnagargade:
This pull request introduces significant improvements to support FlexContainer widgets in the app builder, focusing on drag-and-drop, resizing, and rendering behaviors. The changes ensure that FlexContainer children follow flexbox sizing and insertion logic, and add custom handling for drag-and-drop and resizing, making the UI more intuitive and consistent with modern layout paradigms.

**FlexContainer widget enhancements:**

* Added detection of FlexContainer widgets and their children throughout `Container.jsx` and `Grid.jsx`, enabling special handling for drag-and-drop, resizing, and rendering. [[1]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR76-R83) [[2]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR252) [[3]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR261-R283) [[4]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612R647-R677)
* Integrated new flex-specific drop logic: computes and publishes insertion index during drag-over using `computeFlexInsertIndex`, and visually indicates drop targets. [[1]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR122-R132) [[2]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR96-R109) [[3]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fL184-R213) [[4]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR312-R313)

**Resizing behavior improvements:**

* Implemented resizing for FlexContainer children: resizing adjusts only the main flex axis (width for row, height for column), disables translation, and updates styles directly, rather than using grid coordinates. [[1]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612R647-R677) [[2]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612L654-R693) [[3]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612R762-R785) [[4]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612L741-R806)
* Ensured that resizing a FlexContainer widget itself keeps the left/top edge fixed, matching live preview and improving user experience. [[1]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612L654-R693) [[2]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612L741-R806)

**Drag-and-drop logic changes:**

* Added a dedicated handler (`handleFlexContainerDragEnd`) for drag-end events on FlexContainer children, separating flex logic from grid logic and ensuring correct state updates. [[1]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612R44-R45) [[2]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612R1117-R1139)
* Throttled drop-target updates for FlexContainer using `requestAnimationFrame` to optimize performance during drag-over. [[1]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR96-R109) [[2]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR122-R132)

**Rendering and code structure:**

* Refactored `Container.jsx` to render FlexContainer children using `FlexContainerChildWrapper` and display a drop indicator, while keeping grid-based rendering for other containers. [[1]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR261-R283) [[2]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR312-R313)
* Updated imports and utility usage to support new flex behaviors and maintain code clarity. [[1]](diffhunk://#diff-9fd925fa70af691f89f536b8a06db0c898c0b165b81a09937665b62c6599e94fR26-R28) [[2]](diffhunk://#diff-d7c705dd712f721b76e812ef7ec5a4d06b69312c6b30b2af776aaa476c3cc612R30-R31)

These changes collectively modernize the app builder's layout engine, providing a more robust and flexible foundation for building responsive UIs with FlexContainer widgets.


kavinvenkatachalam:
### Code review

Found 3 issues:

1. **Debug `console.log` left in production code** - `Grid.jsx` logs `batchedLayouts` on every drag-move event, which will flood the browser console for all users.

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx#L280-L283

2. **Migration class name timestamp does not match filename** - The filename is `1775040800000-AddFlexOrderToLayoutsTable.ts` but the class is named `AddFlexSizingToLayoutsTable1775040800001`. TypeORM records migrations by class name, so the effective timestamp it registers (`1775040800001`) differs from the filename convention (`1775040800000`). The class description ("FlexSizing") also diverges from the filename ("FlexOrder"). Every other migration in this repo (e.g. `1778760592536-AddBannerImageToWhiteLabelling.ts` → `AddBannerImageToWhiteLabelling1778760592536`) follows the pattern of matching filename timestamp to class name suffix exactly.

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/server/migrations/1775040800000-AddFlexOrderToLayoutsTable.ts#L1-L5

3. **`height` is unintentionally deleted from the frontend store when reparenting between two FlexContainers** - In `moveFlexContainerChild`, when `layoutPatch` (which carries `{ fillWidth, widthPx, height }` from `carriedSizing` in `flexContainerDragEnd.js`) is applied, the code then deletes `height` from the layout along with the grid-only fields `top`, `left`, and `width`. The intent is clearly to strip absolute grid fields, but `height` is a valid flex layout field and should be preserved. After the reparent, the frontend store has `height: undefined`, causing the widget to fall back to a 100px default in `useFlexWidgetLayout` until the next page reload (when the server's persisted value is loaded). The `saveComponentChanges` call does still include `height` in its diff, so the server is correct — only the in-memory store is wrong.

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/_stores/slices/flexContainerSlice.js#L163-L175

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kavinvenkatachalam:
### Code review

Found 2 issues:

1. Debug `console.log` left in production path (inside the `alignGuidesUpdate` batch loop in Grid.jsx)

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx#L279-L284

2. `moveFlexContainerChild` unconditionally deletes `height` from `child.layouts[currentLayout]` after applying `layoutPatch` — but `layoutPatch` (built as `carriedSizing` in Case 2 of `flexContainerDragEnd.js`) deliberately includes `height` so the widget preserves its size when reparented from one FlexContainer to another. The `delete` on line 2872 throws that value away from client state immediately after the spread, leaving the in-memory layout without a height. `useFlexWidgetLayout` reads `layoutData.height` to size the wrapper, so the widget collapses to the 100 px fallback after an inter-container drag. The server receives the correct value (it is included in `layoutDiff`), creating a client/server divergence until the next page reload.

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/_stores/slices/flexContainerSlice.js#L164-L173

The fix is to remove `height` from the unconditional delete block (lines that delete `top`, `left`, `width`, `height`) — only grid-position fields should be stripped when moving into a FlexContainer, not `height`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kavinvenkatachalam:
### Code review

Found 3 issues:

1. **Debug `console.log` left in `Grid.jsx`** — fires on every group-resize batch update, polluting production console.

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx#L230-L234

2. **`moveFlexContainerChild` deletes `height` after spreading `layoutPatch` (which contains `height`)** — when dragging a widget from one `FlexContainer` into another, `flexContainerDragEnd.js` builds `carriedSizing` with `height` from the source layout and passes it as `layoutPatch`. Inside `moveFlexContainerChild`, the layout is first merged with `...layoutPatch` (so `height` is written) and then unconditionally `delete`d on the next line. The widget ends up with no `height` in its layout after a cross-container drop.

```js
child.layouts[state.currentLayout] = {
  ...child.layouts[state.currentLayout],
  ...layoutPatch,                          // height: 100 written here
};
delete child.layouts[state.currentLayout].height;  // immediately removed
```

Fix: remove `height` from the delete list (grid position fields `top`, `left`, `width` should be stripped, but `height` is a valid flex-child layout field and is carried forward intentionally).

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/_stores/slices/flexContainerSlice.js#L155-L165

3. **`scheduleFlexContainerDropTargetUpdate` is called unconditionally for every drag event of any widget** — the call at line 1277 of `Grid.jsx` is outside the "is this a FlexContainer child?" early-return block, so it fires ~60 fps for every widget drag on the canvas. When the cursor is not near a `FlexContainer`, the rAF callback resolves `flexContainerId` as `null` and calls `setFlexContainerDropTarget(null)` — even when `flexContainerDropTarget` is already `null`. Each call triggers a Zustand `set()`, which notifies all subscribers regardless of whether the value changed. Add a guard in the rAF callback (or move the call inside the flex-child branch) to avoid the no-op store writes on non-flex drags.

https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx#L1274-L1283

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx
```diff
@@ -1165,6 +1274,12 @@ export default function Grid({ gridWidth, currentLayout, mainCanvasWidth }) {
             e.target.style.width = `${e.target.clientWidth}px`;
           }
 
+          scheduleFlexContainerDropTargetUpdate({
```
kavinvenkatachalam, line 1276: Can we add the condition to call this only for `FlexContainer`?

frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx
```diff
@@ -1134,6 +1224,25 @@ export default function Grid({ gridWidth, currentLayout, mainCanvasWidth }) {
...
+            currentWidget?.component?.parent &&
+            getComponentTypeFromId(currentWidget.component.parent) === 'FlexContainer'
+          ) {
+            const scrollDelta = getScrollDelta();
+            e.target.style.transform = `translate(${e.translate[0] + scrollDelta.x}px, ${
+              e.translate[1] + scrollDelta.y
+            }px)`;
+            positionGhostElement(e.target, 'moveable-ghost-widget');
+            updateMousePosition(e.clientX, e.clientY, e.target);
+            scheduleFlexContainerDropTargetUpdate({
```
kavinvenkatachalam, line 1237: Can we add the condition to call this only for `FlexContainer`?

kavinvenkatachalam (changes requested):
@nakulnagargade Suggested one code change. There are few issues when I drop the subcontainers - 

1. Config handle extends outside the component

Image

2. Drop position is incorrect
Image

3. I tried to move the component outside the sub container and into the flex container but it got stuck somewhere in between

Image

4. If I place container inside the flex and made the direction as column, I'm not able to view other components

Image

5. Tooltip and loading state are not working

frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx
nakulnagargade: Yes, This is currently is under this condition.
```
if (
            currentWidget?.component?.parent &&
            getComponentTypeFromId(currentWidget.component.parent) === 'FlexContainer'
          ) 
```

frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx
nakulnagargade: This is required without the condition since if we are moving from any Subcontainer to flexContainer, we need to show the drop indicator during drag.

nakulnagargade:
> ### Code review
> Found 3 issues:
> 
> 1. **Debug `console.log` left in `Grid.jsx`** — fires on every group-resize batch update, polluting production console.
> 
> https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx#L230-L234
> 
> 2. **`moveFlexContainerChild` deletes `height` after spreading `layoutPatch` (which contains `height`)** — when dragging a widget from one `FlexContainer` into another, `flexContainerDragEnd.js` builds `carriedSizing` with `height` from the source layout and passes it as `layoutPatch`. Inside `moveFlexContainerChild`, the layout is first merged with `...layoutPatch` (so `height` is written) and then unconditionally `delete`d on the next line. The widget ends up with no `height` in its layout after a cross-container drop.
> 
> ```js
> child.layouts[state.currentLayout] = {
>   ...child.layouts[state.currentLayout],
>   ...layoutPatch,                          // height: 100 written here
> };
> delete child.layouts[state.currentLayout].height;  // immediately removed
> ```
> 
> Fix: remove `height` from the delete list (grid position fields `top`, `left`, `width` should be stripped, but `height` is a valid flex-child layout field and is carried forward intentionally).
> 
> https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/_stores/slices/flexContainerSlice.js#L155-L165
> 
> 3. **`scheduleFlexContainerDropTargetUpdate` is called unconditionally for every drag event of any widget** — the call at line 1277 of `Grid.jsx` is outside the "is this a FlexContainer child?" early-return block, so it fires ~60 fps for every widget drag on the canvas. When the cursor is not near a `FlexContainer`, the rAF callback resolves `flexContainerId` as `null` and calls `setFlexContainerDropTarget(null)` — even when `flexContainerDropTarget` is already `null`. Each call triggers a Zustand `set()`, which notifies all subscribers regardless of whether the value changed. Add a guard in the rAF callback (or move the call inside the flex-child branch) to avoid the no-op store writes on non-flex drags.
> 
> https://github.com/ToolJet/ToolJet/blob/d03942626ecf58bccd74d1df916535486a09a984/frontend/src/AppBuilder/AppCanvas/Grid/Grid.jsx#L1274-L1283
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Resolved!

shaurya-sharma064 (dismissed):

PR #16411, opened 2026-05-14
fix: prevent clone/export crash for missing ToolJet DB tables

Pratush613:
# PR Title

fix: prevent clone/export failures caused by stale ToolJet DB table references

---

# Files Changed

```txt
server/src/modules/import-export-resources/service.ts

server/src/modules/tooljet-db/services/tooljet-db-import-export.service.ts

server/src/modules/tooljet-db/services/tooljet-db-table-operations.service.ts
```

---

# Issue

Apps can contain stale `table_id` references inside `data_queries.options` for ToolJet DB queries.

This happens when a ToolJet DB table is deleted after being used inside an app query.

## Failure Flow

```text
ToolJet DB table created
        │
        ▼
Query created using the table
        │
        ▼
App saved/published
        │
        ▼
ToolJet DB table deleted
        │
        ▼
stale data_queries.options.table_id remains
        │
        ▼
Clone / Export triggered
        │
        ▼
findTables() picks stale table_id
        │
        ▼
InternalTable lookup returns null
        │
        ▼
"Tooljet database table not found"
        │
        ▼
Entire operation fails
```

---

# Approach to Fix

This PR introduces 2 protections.

## 1. Clone/Export Resiliency

Before exporting ToolJet DB resources, stale/missing ToolJet DB table references are filtered out safely.

This prevents clone/export flows from crashing when apps contain deleted ToolJet DB references.

---

## 2. Delete Guard for ToolJet DB Tables

Added a delete guard similar to Data Sources.

A ToolJet DB table can no longer be deleted if it is being used in any app query across any app version.

This prevents future stale references from being created.

---

# What is Fixed

## Before

- ToolJet DB tables could be deleted even if referenced in queries
- Stale `table_id` references remained in app JSON
- Clone/export failed with:

```text
Tooljet database table not found
```

---

## After

- Clone/export succeeds even if stale references exist
- Missing ToolJet DB references are skipped safely
- ToolJet DB tables cannot be deleted if used in queries
- Guard checks usage across all versions

---

# How to Test

### Steps

1. Create a ToolJet DB table

Example:

```text
users_test
```

2. Create an app

3. Create a query using that ToolJet DB table

4. Save/publish the app

5. Delete the ToolJet DB table from ToolJet Database UI

6. Try:
   - Export app
   - Clone app

---

### Export

Fails with:

```text
Tooljet database table not found
```

### Clone

Fails with:

```text
Tooljet database table not found
```

---

## On This PR Branch

Using the same broken app:

### Export

✅ Export succeeds

### Clone

✅ Clone succeeds

---

# Scenario 2 — Delete Guard Validation

## Steps

1. Create a new ToolJet DB table

2. Create a new app

3. Use the table inside a query

4. Save/publish the app

5. Try deleting the ToolJet DB table

---

## Expected Behavior

❌ Table deletion is blocked

Behavior is similar to the existing Data Source delete guard.

If the table is referenced in any app version/query, it cannot be deleted.

---

# Flow After Fix

```text
ToolJet DB table created
        │
        ▼
Used inside app query
        │
        ▼
Delete attempted
        │
        ▼
Usage check across all app versions
        │
        ├── Table in use
        │       ▼
        │   Delete blocked
        │
        └── Not in use
                ▼
           Delete allowed
```

---

# What's Included in This Branch

- Guard against stale ToolJet DB references during clone/export
- Delete protection for ToolJet DB tables
- Usage validation across all app versions
- Prevent hard crashes caused by deleted ToolJet DB references
- Behavior aligned with existing Data Source delete guard

### DIFF
--- a/server/src/modules/tooljet-db/services/tooljet-db-import-export.service.ts
+++ b/server/src/modules/tooljet-db/services/tooljet-db-import-export.service.ts
@@ -24,7 +24,12 @@ export class TooljetDbImportExportService {
     const internalTable = await this.manager.findOne(InternalTable, {
       where: { organizationId, id: tjDbDto.table_id },
     });
-
+    
+    // Backward compatibility:
+    // Older apps may contain stale ToolJet DB table references because
+    // table deletion was historically allowed even when queries still referenced them.
+    // We intentionally skip missing tables here to avoid hard failures during export/clone.
+    // New stale references are prevented by delete guards added in this PR.
     if (!internalTable) {
       return null;
     }
--- a/server/src/modules/tooljet-db/services/tooljet-db-import-export.service.ts
+++ b/server/src/modules/tooljet-db/services/tooljet-db-import-export.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, NotFoundException } from '@nestjs/common';
+import { Injectable } from '@nestjs/common';
 import { ExportTooljetDatabaseDto } from '@dto/export-resources.dto';
 import { ImportResourcesDto, ImportTooljetDatabaseDto } from '@dto/import-resources.dto';
 import { EntityManager } from 'typeorm';
@@ -25,7 +25,9 @@ export class TooljetDbImportExportService {
       where: { organizationId, id: tjDbDto.table_id },
     });
 
-    if (!internalTable) throw new NotFoundException('Tooljet database table not found');
+    if (!internalTable) {
+      return null;
+    }
 
     const { configurations = {} } = internalTable;
     const { columns, foreign_keys } = await this.tableOperationsService.perform(organizationId, 'view_table', {
--- a/server/src/modules/tooljet-db/services/tooljet-db-table-operations.service.ts
+++ b/server/src/modules/tooljet-db/services/tooljet-db-table-operations.service.ts
@@ -470,23 +470,28 @@ export class TooljetDbTableOperationsService {
     }
   }
 
-private async findQueriesLinkedToTable(tableId: string): Promise<boolean> {
-  const result = await this.manager.query(
-    `
-      SELECT EXISTS (
-        SELECT 1
-        FROM data_queries dq
-        INNER JOIN data_sources ds
-          ON dq.data_source_id = ds.id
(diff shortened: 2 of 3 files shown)


server/src/modules/tooljet-db/services/tooljet-db-import-export.service.ts
```diff
@@ -25,7 +25,9 @@ export class TooljetDbImportExportService {
       where: { organizationId, id: tjDbDto.table_id },
     });
 
-    if (!internalTable) throw new NotFoundException('Tooljet database table not found');
+    if (!internalTable) {
```
akshaysasidrn, line 33: nit: Since we are not doing data migration to fix the apps that refer stale internal tables, should we contextually still throw the error instead of silently supressing it?

As we now added a check to disallow deletion of tables if queries use it, maybe something like:
```js

if (!internalTable) {
  if  (isVersionLessThan("x.x.x")) return null // release version of this PR
  throw new NotFoundException('Tooljet database table not found');
}
```

Or we can have an outer `catch` block for exports where it checks version and contextually suppressing this error. 

Reasoning here is that export logic always works with concrete data. Which enforces we store data properly

server/src/modules/tooljet-db/services/tooljet-db-table-operations.service.ts
```diff
@@ -462,6 +470,24 @@ export class TooljetDbTableOperationsService {
...
+private async findQueriesLinkedToTable(tableId: string): Promise<boolean> {
+  const result = await this.manager.query(
+    `
+      SELECT EXISTS (
+        SELECT 1
+        FROM data_queries dq
+        INNER JOIN data_sources ds
+          ON dq.data_source_id = ds.id
+        WHERE ds.kind = 'tooljetdb'
+          AND dq.options->>'table_id' = $1
```
ganesh8056: For joins query, only verifying `dq.options->>'table_id'` won't work, as it will have nested options for selecting table.

ganesh8056 (changes requested):

johnsoncherian (approved):

PR #16413, opened 2026-05-14
Add invitation token and invite URL generation for external API users and expand test coverage

RndmCodeGuy20:


RndmCodeGuy20:
Two things here:

**Seed fix** — `maybeCreateDefaultGroupPermissions` was only seeding `admin` and `end-user` default groups for test orgs. `builder` was missing, which caused a null constraint on `GroupUsers` whenever a test tried to create a user with `role: 'builder'` via the external API. Added with `isAdmin: true`, consistent with `DEFAULT_GROUP_PERMISSIONS.BUILDER` (appCreate, appDelete, folderCRUD etc. are all true for builder).

**New test file** — `external-apis/e2e/users.spec.ts`, 12 tests covering:
- `inviteUrl` present and well-formed per workspace
- Multi-workspace user creation (regression for the early-return bug fixed in ee-server)
- Role assignment (builder vs end-user) and per-workspace role differences
- Multiple custom groups in one workspace
- Failure paths: duplicate email, nonexistent workspace, default group name in `groups` field, nonexistent group, end-user added to a builder-level custom group

Implementation: ToolJet/ee-server#542

PR #16427, opened 2026-05-15
Feature: Expose control component functions for pagination

parthy007:
[Issue](https://github.com/ToolJet/tj-ee/issues/4754)


server/src/modules/apps/services/widget-config/pagination.js
```diff
@@ -62,7 +62,26 @@ export const paginationConfig = {
...
+      handle: 'setVisibility',
+      displayName: 'Set visibility',
+      params: [{ handle: 'visible', displayName: 'Value', defaultValue: '{{false}}', type: 'toggle' }],
+    },
+    {
+      handle: 'setDisable',
+      displayName: 'Set disable',
+      params: [{ handle: 'disable', displayName: 'Value', defaultValue: '{{false}}', type: 'toggle' }],
+    },
+  ],
```
shaurya-sharma064, line 106: Frontend/server config drift introduced by this PR. The frontend `widgets/pagination.js` adds `setLoading` to `actions` and `isLoading: false` to `exposedVariables`, but the server config here is missing both. Since server config drives validation/seeding for widget properties, `setLoading` calls and the `isLoading` exposed variable may not validate or seed correctly on existing apps. Please mirror both additions on the server side.

shaurya-sharma064 (changes requested):

PR #16433, opened 2026-05-15
fix: JSONSchema form import now generates real editable canvas widgets

### LINKED ISSUE #16432: JSONSchema form import fails to create child fields
### Version Information
tooljet-ce:v3.20.160-lts

### Environment

Docker

Imported forms from JSONSchema should behave the same as those manually constructed or imported from raw data

JSONSchema imported forms don't have any children in the inspector and don't have any formData. They only have a .data property containing the field values. Fields cannot be edited in the editor.

1. Drag a form onto an app
2. Select import from jsonschema and accept the example schema
3. Observe that the fields do not show up in the inspector and cannot be individually edited

### Screenshots or Screencast

Image

None relevant that I can see

### Any possible solutions?
None, other than manually building forms or building from json data and editing

No

Maybe

pranavchoudhary-tech:
## Description

Closes #16432 
When a form's **Generate form from** is set to **JSON schema**, the generated child fields were not appearing in the Inspector and `formData` / `children` were missing from exposed variables.

## Root Cause

The JSONSchema path activated an internal `advanced` mode which dynamically rendered fields via a `RenderSchema` component instead of adding real widgets to the Redux store. Because the children were never added to the store, the Inspector couldn't detect them - making them non-selectable and non-editable.

## Changes

- **`jsonSchemaHandlers.js`** (new file): Parses `newJsonSchema` using `generateUIComponents` and dispatches `performBatchComponentOperations` to create real ToolJet widgets inside the Form's `SubContainer`
- - **`parameterHandlers.js`**: Intercepts `newJsonSchema` param changes and triggers the schema blur handler when source is `jsonSchema`
- - **`useFormLogic.js`**: Routes blur events to the correct handler (`jsonSchema` vs `rawJson`)
- - **`DataSection.jsx`**: Removed the `source.value !== 'jsonSchema'` condition that was hiding the Generate Form button for JSON Schema
- - **`formComponentSlice.js`**: Added `newJsonSchema` to `getFormDataSectionData`, `saveFormDataSectionData`, and `updateFormDataSectionDataLocally` so the schema value is correctly persisted
- - **`Form.jsx`**: Removed `isJSONSchema` from the `advanced` flag so JSONSchema forms now render via `SubContainer` just like all other forms
## Behavior After Fix

1. Select a Form widget and set **Generate form from** -> **JSON schema**
2. 2. Enter a valid schema in the code editor and blur
3. 3. Real canvas widgets are created inside the Form
4. 4. Child fields appear in the Inspector's Fields list
5. 5. Each child widget can be selected and its properties edited
6. 6. `formData` correctly maps to the child widget values
## Related Issue

Resolves #16432

### DIFF
--- a/frontend/src/AppBuilder/RightSideBar/Inspector/Components/Form/_components/DataSection.jsx
+++ b/frontend/src/AppBuilder/RightSideBar/Inspector/Components/Form/_components/DataSection.jsx
@@ -50,23 +50,21 @@ export const DataSection = ({
             focusCodeEditor
           );
         })}
-      {source.value !== 'jsonSchema' && (
-        <DataSectionWrapper
-          currentStatusRef={currentStatusRef}
-          source={source}
-          JSONData={JSONData}
-          component={component}
-          darkMode={darkMode}
-          saveDataSection={saveDataSection}
-          openModalFromParent={openModal}
-          setParentModalState={setParentModalState}
-          performColumnMapping={performColumnMapping}
-          newResolvedJsonData={resolveReferences('canvas', JSONData.value)}
-          existingResolvedJsonData={existingResolvedJsonData}
-          savedSourceValue={savedSourceValue}
-          isLoading={isLoading}
-        />
-      )}
+      <DataSectionWrapper
+        currentStatusRef={currentStatusRef}
+        source={source}
+        JSONData={JSONData}
+        component={component}
+        darkMode={darkMode}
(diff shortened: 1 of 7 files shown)


pranavchoudhary-tech:
Checking in to see if there are any updates on this PR. All CI checks have passed successfully. Let me know if any further changes are required before review and merging.

PR #16486, opened 2026-05-19
fix: pro plan workspace admin blocked from app preview

RndmCodeGuy20:
## Summary

Pro plan has `multiEnvironment: false`. `validateVersionEnvironment` was throwing on **any** `environment_name` param — including `development` — causing a 403 that the frontend mapped to the `restricted-preview` error page for workspace admins.

- **Backend**: allow `development` env through when multi-env is disabled; only block `staging`/`production` to prevent URL manipulation on restricted plans.
- **Frontend** (`AppsRoute.jsx`, `useAppPreviewLink.js`): `isBasicPlan` check was hardcoded to `plan === 'starter'`, missing `pro` and `basic`. Changed to `featureAccess?.multiEnvironment === false` — strips `env` from the URL only for plans without multi-env access, preserves it for team/enterprise (needed for per-env per-version preview sharing). Strict equality used so `undefined` (featureAccess not yet loaded) defaults to preserving env.

## Root Cause

```
GET /api/apps/validate-private-app-access/<id>?environment_name=development&version_name=v6
→ ForbiddenException: URL is not accessible. Multi-environment is not enabled
  at AppsUtilService.validateVersionEnvironment (util.service.ts:202)
```

Pro plan preview URLs include `?env=development` → `handleAppAccess.js` forwards it as `environment_name=development` → backend rejected it because `isMultiEnvironmentEnabled=false`, even though `development` is the only environment pro plan has.

## Files Changed

- `server/src/modules/apps/util.service.ts` — guard in `validateVersionEnvironment`: allow `development` env through when multi-env is disabled
- `frontend/src/Routes/AppsRoute.jsx` — `isBasicPlan` condition
- `frontend/src/_hooks/useAppPreviewLink.js` — same `isBasicPlan` pattern aligned

## Test Plan

- [ ] Pro plan workspace admin → Editor → Preview button → opens without error
- [ ] Pro plan user manually navigates to `?env=staging` or `?env=production` → still blocked (403)
- [ ] Team plan → preview with `?env=staging` → `env` param preserved in URL, opens staging correctly
- [ ] Team plan → share staging preview link → recipient gets correct env
- [ ] Expired/invalid license → preview → still shows restricted page
- [ ] Starter plan → preview → opens without error

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -293,7 +293,8 @@
         "@tooljet-plugins/twilio": "file:packages/twilio",
         "@tooljet-plugins/typesense": "file:packages/typesense",
         "@tooljet-plugins/woocommerce": "file:packages/woocommerce",
-        "@tooljet-plugins/zendesk": "file:packages/zendesk"
+        "@tooljet-plugins/zendesk": "file:packages/zendesk",
+        "snowflake-sdk": "^2.4.0"
       },
       "devDependencies": {
         "@types/jest": "^27.5.2",
@@ -20101,7 +20102,7 @@
         "@tooljet-plugins/common": "file:../common",
         "@types/snowflake-sdk": "^1.6.17",
         "react": "^17.0.2",
-        "snowflake-sdk": "^1.9.1"
+        "snowflake-sdk": "^2.4.0"
       }
     },
     "../plugins/packages/stripe": {
--- a/frontend/src/AppBuilder/_utils/auth.js
+++ b/frontend/src/AppBuilder/_utils/auth.js
@@ -12,7 +12,7 @@ export function fetchOAuthToken(authUrl, dataSourceId) {
 
 export function logoutAction() {
   localStorage.clear();
-  sessionService.logout(true);
+  sessionService.logout(false);
 
   return Promise.resolve();
 }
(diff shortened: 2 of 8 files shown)


RndmCodeGuy20 (commented):
**Root cause trace**

The bug manifests as two compounding issues:

1. **`frontend/src/Routes/AppsRoute.jsx` (line 91, before fix)**
   `isBasicPlan` checked `featureAccess?.plan === 'starter'` — `'pro'` was never included. So for pro plan users `shouldIncludeEnv = true`, meaning `?env=development` was preserved in the redirect URL even though pro doesn't have multi-environment.

2. **`server/src/modules/apps/util.service.ts` (line 201, before fix)**
   `validateVersionEnvironment` threw on **any** `environmentName` when `isMultiEnvironmentEnabled = false` — including `'development'`. Pro plan only has one environment (`development`), so this blanket guard was too strict.

The server log that confirmed it:
```
ForbiddenException: URL is not accessible. Multi-environment is not enabled
at AppsUtilService.validateVersionEnvironment (util.service.ts:202)
```
Request: `GET /api/apps/validate-private-app-access/<id>?environment_name=development&version_name=v6`

**Fix rationale**

- Backend guard intent is to prevent URL manipulation (e.g. a pro user crafting `?env=production`). `development` is the only valid env for non-multi-env plans, so passing it through is safe.
- Frontend fix ensures pro plan never constructs a URL with `?env=` in the first place — using the feature flag directly (`!featureAccess?.multiEnvironment`) instead of a hardcoded plan name so future plan changes don't re-introduce this.

Issue #6655, opened 2023-06-06
incorrect code-hinter header text


withshubh:
### What is the issue?

Currently, when the code hinter is expanded, the header text is shown as `Editor` instead it should display the name of the property from where the code hinter is expanded.

### How to reproduce the issue?

1. Open ToolJet app builder
2. Create a RunJS query
3. Scroll down the query editor and toggle on the `show notification on success?`
4. On `Success Message` input box, expand the code hinter by clicking on the icon on the right of the input box
5. Once the code-hinter pops-up, check the header text

### Screenshots or Screencast
![Untitled](https://github.com/ToolJet/ToolJet/assets/25361949/c22dba8e-3abe-402e-a873-979aba939756)

Avi98:
Hi @withshubh, can I take this up?

withshubh:
Hey @Avi98  
Since the issue holds the good first issue tag which is for welcoming new contributors let's give someone new the opportunity to work on this issue.
I hope you understand.

kisharnath:
Can i work on this?I am new

withshubh:
Hey @kisharnath - Thanks for showing interest in contributing to ToolJet. Assigned it to you.

withshubh:
Hey @kisharnath - Any update on the PR?

sebabratakundu:
@withshubh since there is no activity, can you assign me the issue?

fakhruddinkw:
@withshubh is this still available? I see no updates
If it is, i would like to work on this

tejakummarikuntla:
Since we've not heard back from @kisharnath, we'd need to unassing.

@sebabratakundu given this issue holds the label `good first issue`, let's welcome @fakhruddinkw, who is a first-time contributor to the project; I hope you understand.

Thanks for your interest @fakhruddinkw , assigned this to you.

fakhruddinkw:
@tejakummarikuntla THANK YOU!
Let me go through this issue and get back to you ASAP.

fakhruddinkw:
@tejakummarikuntla unfortunately no. I ran into problem setting up tooljet locally using WSL-2 . @ShazanRizvi Have also tagged you in slack channel regarding this problem. Can you please check?

tejakummarikuntla:
Hey @fakhruddinkw , I learned that your issues were resolved, Unfortunately, I'd need to unassing due to inactivity for a longer duration.

tejeshreddy:
[...]

PR #16520, opened 2026-05-21
fix: show correct property name in code hinter header (#6655)

liangqiang2025-stack:
## Summary

When the code hinter popup is opened, the header text displays Editor instead of the actual property name.

## Fix

Pass paramLabel as componentName to SingleLineCodeEditor.

Fixes #6655

Bounty: $45 (Opire issue 01J8898W2WES9F4FJ57GDG0BYS)


RudraPratik30:
Hey @liangqiang2025-stack, thanks for raising this PR, but this issue was marked closed in Dec 2024 itself, hence we are closing this PR too

PR #16545, opened 2026-05-22
Fix queries not running inside modules successfully on released apps

### LINKED ISSUE #16562: Private app having query inside module is not running successfully without login
Steps To Reproduce

  1. Create a module with a query.
  2. Create another app.
  3. Add the module to that app.
  4. Make that app public.
  5. Open/run the public app without logging in.
  6. Trigger the module query.

 Current Result
  The private module query is not running successfully.

  Expected Result
  The query should run successfully without login

nakulnagargade:


server/src/modules/apps/repository.ts
```diff
@@ -73,6 +77,34 @@ export class AppsRepository extends Repository<App> {
     });
   }
 
+  async findPublicHostAppForModuleQuery(dataQueryId: string, organizationId: string): Promise<App | null> {
```
RndmCodeGuy20: `findByDataQuery(id)` already returns the module app, so `app.id` is the module app ID. we can remove the pre fetch and get `appVersionId` associated directly from joining the `data_queries` table, removing the round-trip.

```typescript
  async findPublicHostAppForModuleQuery(moduleAppId: string, dataQueryId: string): Promise<App | null> {
    return this.createQueryBuilder('app')
      .innerJoin(AppVersion, 'app_version', 'app_version.app_id = app.id')
      .innerJoin(Page, 'page', 'page.app_version_id = app_version.id')
      .innerJoin(Component, 'component', 'component.page_id = page.id')
      .innerJoin('data_queries', 'dq', 'dq.id = :dataQueryId', { dataQueryId })
      .where('component.type = :componentType', { componentType: 'ModuleViewer' })
      .andWhere('app.is_public = true')
      .andWhere('app.current_version_id = app_version.id')
      .andWhere("component.properties::jsonb -> 'moduleAppId' ->> 'value' = :moduleAppId", { moduleAppId })
      .andWhere("component.properties::jsonb -> 'moduleVersionId' ->> 'value' = dq.app_version_id::text")
      .limit(1)
      .getOne();
  }
```

  Guard call changes to:
```typescript
  const hostApp = await this.appRepository.findPublicHostAppForModuleQuery(app.id, id);
  ```

File: server/src/modules/data-queries/guards/query-auth.guard.ts [hunk 2]
```diff
@@ -31,6 +32,12 @@ export class QueryAuthGuard extends AuthGuard('jwt') {
       if (!app) {
         throw new BadRequestException();
       }
+      if (app.type === APP_TYPES.MODULE) {
+        const hostApp = await this.appRepository.findPublicHostAppForModuleQuery(id, app.organizationId);
+        if (hostApp) {
+          app.isPublic = hostApp.isPublic;
```
RndmCodeGuy20: we can keep this `true` for understanding - will be `true` anyways

server/src/modules/data-queries/guards/query-auth.guard.ts
```diff
@@ -31,6 +32,12 @@ export class QueryAuthGuard extends AuthGuard('jwt') {
       if (!app) {
         throw new BadRequestException();
       }
+      if (app.type === APP_TYPES.MODULE) {
```
RndmCodeGuy20: also should we add `!request.headers['authorization']` or a similar condition to check if user is authenticated or not - since we only need this lookup for unauthenticated users, we can skip it for authenticated users. cc @johnsoncherian @kavinvenkatachalam @gsmithun4

server/src/modules/apps/repository.ts
nakulnagargade: Done

server/src/modules/data-queries/guards/query-auth.guard.ts
nakulnagargade: Done

server/src/modules/data-queries/guards/query-auth.guard.ts
nakulnagargade: Done

johnsoncherian (dismissed):

gsmithun4 (approved):

PR #16552, opened 2026-05-24
fix: code hinter header shows component name instead of 'Editor' (closes #6655)

18850196928-max:
## What

Fixes #6655 — the code hinter popup header was always showing "Editor" instead of the actual component/property name.

## Root Cause

In `Portal.jsx`, the header used `{componentName ?? 'Editor'}` as fallback. When `componentName` prop was not passed (e.g., from certain inspector fields like "Success Message"), the header defaulted to the hardcoded string "Editor".

## Fix

Changed the fallback in Portal.jsx line ~118 from:
```jsx
{componentName ?? 'Editor'}
```
to:
```jsx
{componentName || 'Code Editor'}
```

This ensures the header shows the actual component name when available, and a neutral "Code Editor" label when it's not.

## Testing

1. Open ToolJet app builder
2. Create a RunJS query
3. Toggle on "show notification on success?"
4. Expand the code hinter on "Success Message" input
5. Verify header no longer shows hardcoded "Editor"

Closes #6655

### DIFF
--- a/frontend/src/_components/Portal/Portal.jsx
+++ b/frontend/src/_components/Portal/Portal.jsx
@@ -120,7 +120,7 @@ const Modal = ({
             className="codehinder-popup-badge"
             data-cy="codehinder-popup-badge"
           >
-            {componentName ?? 'Editor'}
+            {componentName || 'Code Editor'}
           </span>
         </div>
 


RudraPratik30:
Hi @18850196928-max, thanks for raising the PR but this issue was closed in December 2024 itself, and hence we are closing this PR too.

PR #16569, opened 2026-05-25
fix: table popover scroll bleed and overlap

parthy007:
[Issue-1](https://github.com/ToolJet/tj-ee/issues/4942)
[Issue-2](https://github.com/ToolJet/tj-ee/issues/4943)


frontend/src/_styles/table-component.scss
```diff
@@ -1228,30 +1228,34 @@
...
-
-
 .table-column-popover#popover-basic-2 {
-  z-index: 999 !important;
+  z-index: 11 !important;
   min-width: 280px;
-  max-height: 85vh !important;
+  display: flex !important;
+  flex-direction: column !important;
+  height: calc(100vh - 88px) !important;
```
shaurya-sharma064: `height: calc(100vh - 88px) !important` forces the popover to exactly viewport-height regardless of content. Previous `max-height: 85vh !important` was a cap, not a force — content sized the box.

This selector also matches KeyValuePair's FieldPopover (same id `popover-basic-2` in `KeyValuePair.jsx:95`, same `table-column-popover` class in `FieldPopover.jsx:45`), so the KeyValuePair popover now renders as a near-empty full-viewport box.

shaurya-sharma064 (changes requested):

frontend/src/_styles/table-component.scss
parthy007: After discussion with @johnsoncherian, we decided to make the table popover equal to the editor's height. Although have fixed the styling and made the changes unique to the table popover

johnsoncherian (approved):

PR #16571, opened 2026-05-25
fix: show code hinter titles per field

duongynhi000005-oss:
Use the field label as the portal title for code hinter popups, falling back to the component name. This avoids the generic "Editor" header when the popup is opened from a specific property.


RudraPratik30:
Hey @duongynhi000005-oss, thanks for raising this PR!

Could you please link the relevant issue here?

PR #16580, opened 2026-05-26
Perf: fix module version loading causing pool exhaustion on apps with many branches

kavinvenkatachalam:
## 📝 What this does
Apps using the Modules feature were timing out with HTTP 500 (`timeout exceeded when trying to connect`) because loading a module was fetching every branch version instead of just the one needed. This fix ensures `fetchModules` resolves the correct version upfront so downstream callers take the efficient single-version path.

## 🔀 Changes
- Removed heavy `findVersion` call in `fetchModules` — version ID is now resolved directly from the already-loaded entity
- `fetchModules` now populates `appVersions` on returned module entities so `prepareResponse` takes the fast path (1 query) instead of loading all versions via `findVersionsFromApp`
- Added explicit `editingVersion` fallback for module apps when the TypeORM subscriber hasn't hydrated it
- `prepareResponse` in `VersionService.getVersion` now passes `module.editingVersion?.id` instead of `undefined`

## 🧪 How to test
- [ ] Open a parent app that embeds a module via `ModuleViewer`
- [ ] Confirm `GET /api/v2/apps/:appId/versions/:versionId?mode=view` returns 200 within ~2s
- [ ] Open the same app with multiple concurrent users — confirm no HTTP 500 pool exhaustion errors
- [ ] Confirm version switching and branch switching still work correctly on both parent and module apps
- [ ] Open a parent app on a feature branch where the module has no matching branch — confirm it falls back to the module's default version correctly

### DIFF
--- a/server/src/modules/apps/util.service.ts
+++ b/server/src/modules/apps/util.service.ts
@@ -511,10 +511,7 @@ export class AppsUtilService implements IAppsUtilService {
   ): Promise<AppBase[]> {
     const qb = await this.buildViewableAppsQuery(user, type, searchKey, isGetAll, branchId, this.appRepository.manager);
     if (isGetAll) return qb.getMany();
-    return qb
-      .take(APPS_PAGE_SIZE)
-      .skip(APPS_PAGE_SIZE * (page - 1))
-      .getMany();
+    return qb.take(APPS_PAGE_SIZE).skip(APPS_PAGE_SIZE * (page - 1)).getMany();
   }
 
   async allWithCount(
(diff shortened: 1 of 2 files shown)


server/src/modules/apps/util.service.ts
```diff
@@ -819,16 +819,36 @@ export class AppsUtilService implements IAppsUtilService {
       // because it has no branch context. When the parent app is being viewed on a feature
       // branch, callers rely on module.editingVersion downstream (pages/queries/events), so
       // stale default-branch content leaks through unless we override here.
-      const parentBranchId = app.editingVersion?.branchId;
-      if (parentBranchId && modules.length > 0) {
+      const parentBranchId = (app as any).editingVersion?.branchId;
+      if (modules.length > 0) {
         await Promise.all(
           modules.map(async (moduleApp: any) => {
```
akshaysasidrn: Can we avoid this map x findOne queries?
We should be able to use a single query to further optimise. We can remove the db transaction wrap on this as well as its a read query.

akshaysasidrn (changes requested):

akshaysasidrn (dismissed):

johnsoncherian (approved):

PR #16612, opened 2026-06-01
feat: add markdown and HTML support for tooltips in widgets

manishkushare:
Resolves https://github.com/orgs/ToolJet/projects/26/views/4?pane=issue&itemId=155859456&issue=ToolJet%7Ctj-ee%7C4699


shaurya-sharma064 (approved):

frontend/src/AppBuilder/AppCanvas/WidgetTooltip.jsx
```diff
@@ -0,0 +1,88 @@
...
+const WidgetTooltip = ({
+  content,
+  format = DEFAULT_FORMAT,
+  show = true,
+  children,
+  delayDuration = 500,
+  darkMode = false,
+}) => {
+  const trimmed = typeof content === 'string' ? content.trim() : '';
+  if (!show || !trimmed) return children;
```
shaurya-sharma064: Returning bare `children` here vs the wrapped tree below means the root element type changes (`div` ↔ `TooltipProvider`) whenever the tooltip flips between empty and non-empty. React then remounts the entire widget subtree — local state is lost (table page/selection, input drafts), exposed variables reset, mount effects re-fire.

Repro: bind a Table's tooltip to `{{components.textinput1.value}}` and type one character — the Table remounts and selection/page reset.

The previous `OverlayTrigger` always wrapped and only nulled `trigger`, so the tree shape never changed. Suggest always rendering the Provider/Tooltip wrapper and conditionally rendering only `<TooltipContent>` — it's portaled, so `children`'s tree position stays stable.

frontend/src/AppBuilder/AppCanvas/WidgetTooltip.jsx
```diff
@@ -0,0 +1,88 @@
...
+import remarkGfm from 'remark-gfm';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/Rocket/Tooltip/Tooltip';
+import './WidgetTooltip.scss';
+
+export const TOOLTIP_FORMATS = ['plainText', 'markdown', 'html'];
+const DEFAULT_FORMAT = 'plainText';
+const MARKDOWN_REMARK_PLUGINS = [remarkGfm];
+
+// styles for markdown and plain text
+const THEMED_CLASSES = [
```
shaurya-sharma064, line 15: This drops the `whiteSpace: 'pre-wrap'` override added in #15223 (fix for tj-ee#4566). Plaintext tooltips with resolved newlines (`{{components.textarea1.value}}`, query results) flatten back to one line — and since the backfill defaults all existing tooltips to `plainText`, every existing tooltip takes this path. Suggest adding `tw-whitespace-pre-wrap` here or to the plaintext branch of `TooltipBody`.

frontend/src/AppBuilder/AppCanvas/RenderWidget.jsx
```diff
@@ -278,73 +279,88 @@ const RenderWidget = ({
 
   if (!component) return null;
 
+  // Pick the tooltip source — some widgets keep tooltip under `properties`,
+  // others under `general`. SHOULD_ADD_BOX_SHADOW_AND_VISIBILITY enumerates
+  // the former group.
+  const isShadowedWidget = SHOULD_ADD_BOX_SHADOW_AND_VISIBILITY.includes(component?.component);
```
shaurya-sharma064, line 289: `TagsInput` and `TreeSelect` define `tooltip`/`tooltipFormat` under `properties` (and the migration lists them in `PROPERTIES_TYPES`), but neither is in `SHOULD_ADD_BOX_SHADOW_AND_VISIBILITY` — so this reads `general.tooltip` for them and the inspector value never renders; the new format switch is a no-op for these two widgets. Add them to the list, or key the tooltip source off something other than the box-shadow list.

server/data-migrations/1780550424338-BackfillTooltipFormat.ts
```diff
@@ -0,0 +1,168 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { deleteAppHistoryForStructuralMigration } from '@helpers/migration.helper';
+
+// /Users/ahimns / Desktop / work / arm / ToolJet / server / data - migrations / 1780550424338 - BackfillTooltipFormat.ts;
```
shaurya-sharma064: Leftover local path from your machine — please remove.

server/data-migrations/1780550424338-BackfillTooltipFormat.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export class BackfillTooltipFormat1780550424338 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const propsUpdated = await backfillColumn(queryRunner, PROPERTIES_TYPES, 'properties', 'additionalActions');
+    const generalUpdated = await backfillColumn(queryRunner, GENERAL_TYPES, 'general_properties', 'general');
+    const totalUpdated = propsUpdated + generalUpdated;
+
+    if (totalUpdated > 0) {
+      const appVersionRows: { app_version_id: string }[] = await queryRunner.query(
+        `SELECT DISTINCT p.app_version_id
```
shaurya-sharma064: This collects every app version *containing* any of the 65 widget types — effectively all apps — rather than versions whose rows were actually patched. On a crash + re-run, history gets wiped again for versions already migrated. The batch loop already has the updated ids — suggest joining `pages` there and accumulating `app_version_id`s of rows actually updated, then deleting history only for those.

frontend/src/AppBuilder/WidgetManager/widgets/_sharedGeneralTooltip.js
```diff
@@ -0,0 +1,48 @@
+// Shared "Tooltip" field pair for widgets whose tooltip lives in the inspector's
+// General section (vs Additional Actions). Mirrors the additionalActions pair
+// used by widgets like Button — same UX, different section. See
+// docs/my-docs/tooltip-rich-content.md for the rationale and rollout plan.
```
shaurya-sharma064: `docs/my-docs/tooltip-rich-content.md` doesn't exist in the repo — looks like a reference to personal notes; please remove.

frontend/src/AppBuilder/WidgetManager/widgets/_sharedGeneralTooltip.js
```diff
@@ -0,0 +1,48 @@
...
+// visible "Tooltip" label.
+//
+// Implication: future changes to `universalProps.general.tooltip` (e.g., new
+// validation rules, default values, type changes) will NOT reach the 16
+// general-section widgets unless mirrored here in `GENERAL_TOOLTIP_FIELDS`.
+// If you touch the universal tooltip definition, check that this file stays
+// compatible. Server mirror lives at
+// `server/src/modules/apps/services/widget-config/_sharedGeneralTooltip.js` —
+// keep both in sync.
+export const GENERAL_TOOLTIP_FIELDS = {
```
shaurya-sharma064: Is growing `general` intended here? Convention has been to move general properties into `properties` (`section: 'additionalActions'`) — no other widget config defines its own `general:` block. Migrating these 16 widgets' tooltip to `properties` instead would remove this shared file (plus server mirror and the maintenance note), the `DefaultComponent.jsx` special case, and the dual-source read in `RenderWidget.jsx`. If moving the tooltip value is out of scope for LTS, could at least the new `tooltipFormat` live in `properties` so `general` doesn't grow?

shaurya-sharma064 (changes requested):

shaurya-sharma064 (dismissed):

PR #16629, opened 2026-06-01
Document client-side datasource response behavior

### LINKED ISSUE #5618: [docs]: Add general notice that all requested API data is sent to client without filtering
### Summary

While working extensively with the baserow.io implementation of ToolJet I noticed that **all API responses** are sent to the client machine (web browser) **without the ability to be filtered on server-side**. As this might lead to unwanted exposure of data there should be a **warning** about that fact in the docs.

Currently the section "Security" is not highlighing this fact in a noticable manner (from my perspective) so that everybody is aware of that fact:
image

### Issue Type

Documentation feature/update

### The entire URL of the documentation with the issue

https://docs.tooljet.com/docs/security/

### Steps to reproduce the issue

-

### Additional Information

As this **does not happen** with SQL (I think so) the notice does not make sense there. Maybe we could place a general notice / warning about that fact in the docs of every implementation where this happens like baserow.io and others. And there should also be a **noticable warning** inside the Security section.

I also opened a **feature request for server-side API preprocessing** here: https://github.com/ToolJet/ToolJet/issues/5617

### If the issue is confirmed, would you be willing to submit a pull request?

Yes

### Code of Conduct

- [X] I agree to follow the ToolJet Code of Conduct

miachillgood:
## Summary

- Add a caution to the security compliance page explaining that datasource responses are passed through to the client.
- Recommend enforcing row/field filters, limits, and permissions in the query, API, or datasource itself.

Fixes #5618.

## Validation

- `git diff --check`

### DIFF
--- a/docs/docs/security/compliance.md
+++ b/docs/docs/security/compliance.md
@@ -37,6 +37,10 @@ We encourage all our users to practice good security habits to enhance security 
 
 ToolJet does not store data returned from your data sources. ToolJet server acts as a proxy and passes the data as it is to the ToolJet client. The credentials for the data sources are handled by the server and never exposed to the client. For example, if you are making an API request, the query is run from the server and not from the frontend.
 
+:::caution
+Only return data that the current user is allowed to see. ToolJet does not automatically apply server-side row or field filtering to datasource responses before passing them to the client, so filters, limits, and permissions should be enforced in the query, API, or datasource itself.
+:::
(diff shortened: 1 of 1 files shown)


miachillgood:
Closing this as a duplicate of #13044, which predates this PR and addresses the same documentation request in #5618. Sorry for the duplicate.

PR #16701, opened 2026-06-04
feat(query): Run a query periodically

nithindavid:
## What & why

Adds a **"Run this query periodically"** query setting that re-runs a query on a timer after its first execution — for system health checks, dashboard auto-refresh, cache updates, and scheduled reports.

Ref: **ToolJet/tj-ee#3776**

## Release notes

> **Run a query periodically** — Queries now have a **Run this query periodically** option under **Settings**. Enable it and set a **Time interval (ms)** to automatically re-run the query on a recurring basis after its first run. Both fields support `fx` expressions. Control it at runtime with the new `queries.<name>.pausePeriodicRun()` and `queries.<name>.resumePeriodicRun()` actions.

## Changes

**UI** (`695e4aa…`)
- `Run this query periodically` toggle (with `fx`) in the Settings → Triggers section, mirroring the evolved `requestConfirmation` pattern.
- Reveals a `Time interval ( ms )` field (with `fx`) directly underneath when enabled/expression-set (`PeriodicRunInputs`).

**Runtime engine** (`5e4ba9f…`, `componentsSlice`)
- `setInterval` timer per query, mirroring the existing `runOnDependencyChange` debounce infra. Start / stop / reconcile / clear-all helpers.
- **Lifecycle:** starts after a query's first real run; reconciles on option edits (never starts an un-run query); cleared on delete; **survives page navigation**; torn down on full app teardown only.
- **Guards:** interval clamped to **≥ 1000 ms**; invalid/empty → no timer + one console warning; ticks **skip when a run is in flight** (no pile-up); ticks **bypass the confirmation modal**.

**Runtime actions** (`eventsSlice` / `queryPanelSlice` / `codeHinterSlice`)
- `queries.<name>.pausePeriodicRun()` / `resumePeriodicRun()` callable from events + runjs/runpy, with autocomplete hints. Pause is sticky until resume.

## Behavior decisions
- **Starts after the first execution** (page-load / manual / CSA). A query with the toggle on but never run won't tick — pair with *Run on app load* to auto-start.
- **Confirmation + periodic** → ticks auto-confirm (no modal spam); the first/manual run still prompts.
- **Runs in all modes including the editor canvas** (clamp + skip-in-flight keep it safe). Easy to scope to preview/released-only if desired.

## Scope
Frontend-only. Query `options` is an opaque JSON blob, so the new keys (`runPeriodically`, `runPeriodicallyFx`, `queryRunInterval`) persist and round-trip through import/export with **no server/DTO/migration changes**.

## Testing
- Lint-clean (0 errors).
- Verified live: toggle/`fx`/reveal render; 2 s interval ticks; toggle-off stops; min-interval clamp; confirmation bypass; survives page navigation.
- Cypress E2E: follow-up.

### DIFF
--- a/cypress-tests/cypress/e2e/happyPath/appbuilder/commonTestcases/newSuits/queries/runQueryPeriodically.cy.js
+++ b/cypress-tests/cypress/e2e/happyPath/appbuilder/commonTestcases/newSuits/queries/runQueryPeriodically.cy.js
@@ -1,0 +1,119 @@
+import { fake } from "Fixtures/fake";
+import { commonWidgetSelector } from "Selectors/common";
+import {
+  addInputOnQueryField,
+  changeQueryToggles,
+  query,
+  selectQueryFromLandingPage,
+} from "Support/utils/queries";
+import { resizeQueryPanel } from "Support/utils/dataSource";
+import { openNode } from "Support/utils/inspector";
+
+// Selectors for the "Run this query periodically" query setting.
+// Toggle dataCy is `run-query-periodically`; CustomToggleSwitch/FxButton/CodeHinter
+// derive the suffixes below. The interval + fx-expression fields live in PeriodicRunInputs.
+const sel = {
+  settingsTab: '[data-cy="query-tab-settings"]',
+  toggleSwitch: '[data-cy="run-query-periodically-toggle-switch"]',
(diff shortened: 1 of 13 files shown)


kavinvenkatachalam (changes requested):
@nithindavid We need to move the logic out of componentSlice into a separate util file or slice and the newly created file should be in `.ts` or `.tsx`

PR #16720, opened 2026-06-08
Fix: Table datepicker column fails to parse ISO 8601 date strings

parthy007:


frontend/src/AppBuilder/Shared/DataTypes/renderers/DatePickerRenderer.jsx
```diff
@@ -68,15 +65,26 @@ const parseDate = ({
   parseInUnixTimestamp,
   isTimeChecked,
 }) => {
-  if (!value) return null;
+  if (value === null || value === undefined || value === '') return null;
+
+  // ISO 8601 strings (strict) and epoch millisecond numbers
+  const parseAsIso = (input, timeZone) => {
+    if (typeof input === 'number') return timeZone ? moment.tz(input, timeZone) : moment(input);
```
devanshu052000, line 72: Nit: this `typeof input === 'number'` branch is effectively unreachable here. `parseAsIso` only runs as a fallback after `moment(value, parseDateFormat, true)` fails — but moment treats a numeric input as an epoch-ms timestamp and ignores the format string, so the strict attempt already succeeds (and `isValid()` is true) for any number. So we never reach `parseAsIso` with a number. Harmless, but it's dead code and slightly contradicts the "ISO 8601 strings" intent of the helper. Could drop the number branch (or keep it purely as defensive documentation).

frontend/src/AppBuilder/Shared/DataTypes/renderers/DatePickerRenderer.jsx
```diff
@@ -68,15 +65,26 @@ const parseDate = ({
...
-    momentObj = moment.tz(value, parseDateFormat, timeZoneValue).tz(timeZoneDisplay);
+    momentObj = moment.tz(value, parseDateFormat, true, timeZoneValue);
+    if (!momentObj.isValid()) momentObj = parseAsIso(value, timeZoneValue);
+    if (!momentObj.isValid()) momentObj = moment.tz(value, parseDateFormat, timeZoneValue);
+    momentObj = momentObj.tz(timeZoneDisplay);
   } else {
-    momentObj = moment(value, parseDateFormat);
+    momentObj = moment(value, parseDateFormat, true);
+    if (!momentObj.isValid()) momentObj = parseAsIso(value);
+    if (!momentObj.isValid()) momentObj = moment(value, parseDateFormat);
```
devanshu052000, line 87: Consistency note (re: the new strict-format → ISO 8601 → lenient fallback chain added here): the manual text-input path in `handleInputDateChange` (around line 233, outside this diff) still uses lenient `moment(value, parseDateFormat)` only, so a user typing an ISO 8601 string there would not benefit from the same parsing this fix adds. Probably acceptable since users type in the displayed format, but worth making it a deliberate choice rather than an oversight — consider routing that path through the same parsing logic for consistency.

johnsoncherian (approved):

frontend/src/AppBuilder/Shared/DataTypes/renderers/DatePickerRenderer.jsx
parthy007: ### Needed this check for the below

- Unix timestamp send from the raw json directly
- A query/DB column returning epoch-ms bound to a date column.
- Raw JSON / JS with date: {{ Date.now() }} or any numeric expression.
- A KeyValuePair date field bound to a numeric value.
- Inline edit (changeSet) holding a number.

PR #16775, opened 2026-06-10
feat: add setValue action and CSA to Tags, Timeline, and Timer widgets

manishkushare:
Resolves https://github.com/orgs/ToolJet/projects/26/views/4?sliceBy%5Bvalue%5D=manishkushare&pane=issue&itemId=154143706&issue=ToolJet%7Ctj-ee%7C4681


frontend/src/AppBuilder/Widgets/Timer.jsx
```diff
@@ -40,6 +48,21 @@ export const Timer = function Timer({ height, properties = {}, styles, setExpose
 
   useEffect(() => {
     setExposedVariable('value', {});
+
+    if (typeof setExposedVariables === 'function') {
+      setExposedVariables({
+        setValue: async function (newValue) {
```
nakulnagargade, line 62: 🔴 **Blocking — `setValue` can't stop a running interval.**

`setValue` calls `setTime`/`setExposedVariable` but never clears `intervalId` or resets `state`. If the timer is running, the 15ms interval from `onStart` keeps counting from the new base while `state` stays `'running'` — the set value is immediately overwritten by the live tick.

It also **can't** be fixed by just adding `clearInterval(intervalId)` here: this callback is registered in a `[]`-dep effect, so it closes over the mount-time `intervalId` (`0`). The id must come from a ref:

```js
const intervalIdRef = useRef(0);
useEffect(() => { intervalIdRef.current = intervalId; }, [intervalId]);

// inside setValue, before setTime:
intervalIdRef.current && clearInterval(intervalIdRef.current);
setState('initial');
```

Decide deliberately whether `setValue` should pause or hard-stop the timer.

frontend/src/AppBuilder/Widgets/Timer.jsx
```diff
@@ -40,6 +48,21 @@ export const Timer = function Timer({ height, properties = {}, styles, setExpose
...
+          if (typeof newValue !== 'string') return;
```
nakulnagargade: 🟠 **Output/input format asymmetry (no round-trip).**

Exposed `value` is an object `{ hour, minute, second, mSecond }`, but `setValue` only accepts a `'HH:MM:SS:MS'` string. So `timer1.setValue(timer1.value)` hits this `if (typeof newValue !== 'string') return;` guard and does nothing.

frontend/src/AppBuilder/Widgets/Tags/Tags.jsx
```diff
@@ -110,6 +111,22 @@ export const Tags = function Tags({
...
+      setValue: async function (newTags) {
+        if (newTags == null) {
+          setParsedTags(deriveTagsFromProperties(propsRef.current));
+          return;
+        }
+        if (!Array.isArray(newTags)) {
+          setParsedTags([]);
+          return;
+        }
+        const normalized = newTags.map((tag, index) =>
```
nakulnagargade: 🟡 **`setValue` skips the visibility filter (count ≠ rendered).**

`deriveTagsFromProperties` filters out `visible:false` tags; this `setValue` array branch only maps `id` and keeps them. `renderTag` re-checks visibility and won't draw them — but `parsedTags.length`, the exposed `tags` array, and the `aria-label` count still include them. Result: `setValue([{title:"A"},{title:"B",visible:false}])` → screen shows 1 tag, `tags1.tags` reports 2.

The two ways of building `parsedTags` (data binding vs `setValue`) should normalize identically. Route `setValue` through the same filter, or extract a shared `normalizeTags(arr)` helper both paths call.

nakulnagargade (commented):
Review of the `setValue` CSA additions. The changes follow house convention and are solid in shape; left 3 inline comments — one blocking (Timer interval race) plus two consistency items.

frontend/src/AppBuilder/Widgets/Timer.jsx
manishkushare: @nakulnagargade  Decided , `setValue = hard-stop to initial`

frontend/src/AppBuilder/Widgets/Timer.jsx
```diff
@@ -40,6 +48,21 @@ export const Timer = function Timer({ height, properties = {}, styles, setExpose
...
+          if (parts.length < 4) return;
+          const [HH, MM, SS, MS] = parts;
+          const newTime = getTimeObj({ HH, MM, SS, MS });
+          setTime(newTime);
+          setExposedVariable('value', newTime);
+        },
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
```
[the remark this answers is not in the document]
manishkushare: Ignoring

frontend/src/AppBuilder/WidgetManager/widgets/timer.js
```diff
@@ -62,6 +62,19 @@ export const timerConfig = {
   exposedVariables: {
     value: '',
   },
```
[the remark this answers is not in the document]
manishkushare: Ignoring

server/src/modules/apps/services/widget-config/timer.js
```diff
@@ -62,6 +62,19 @@ export const timerConfig = {
   exposedVariables: {
     value: '',
   },
```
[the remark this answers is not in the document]
manishkushare: Ignoring

nakulnagargade (approved):

PR #16777, opened 2026-06-10
To make sticky headers in tables

### LINKED ISSUE #16776: To make Sticky Headers In Tables.
When we scroll, we should have option to make column headers sticky.  Its the most basic thing when you are making tables, that too when the product is based on the thesis of internal app builder for quick application building.

I don't know why this feature is not there yet in such a big product and how come no one ever showed the need for it. 

Even google sheets and all have these features.

krishna76-code:
fixes #16776


RudraPratik30:
Hi @krishna76-code, we reviewed the PR, but it won't solve the usecase mentioned by the user since in their usecase they have a scroll on the page not the table whereas what the PR does is that it gives the user ability to make table header not sticky when the table has a scroll which seems unnecessary. Hence we are closing this PR.

PR #16783, opened 2026-06-10
feat(workspace-branches): paginated branch list with push-time sort and seeded-branch pull

Pratush613:
Resolves → https://github.com/ToolJet/tj-ee/issues/5100
  server/ee → https://github.com/ToolJet/ee-server/pull/590
  
  The Switch Branch modal was loading all branches from the DB in a single call — for orgs with 100+ branches this caused a noticeable delay on every open, and the list was sorted by createdAt
  (effectively alphabetical for seeded branches). This PR replaces that with a Redis-cached, push-time-sorted remote branch list with local Load More pagination, and fixes a related issue where switching
  to a seeded branch showed an empty workspace.

  ---
  What changed

  Remote branch listing — Redis cache + full GraphQL pagination

  Previously: GET /workspace-branches returned every DB branch sorted by createdAt ASC. No remote metadata (commit dates), no meaningful sort order for large orgs.

  Now:
  - Modal open fires two parallel calls: GET /workspace-branches (all DB branches — feeds Fuse.js search) and GET /workspace-branches/remote (sorted remote branch list served from Redis)
  - The remote endpoint fetches all remote branches via GitHub GraphQL refs query with hasNextPage cursor loop — no 100-branch cap
  - Result is merged with DB metadata (id, isDefault, createdBy, createdAt), sorted, and stored in Redis with a 5-minute TTL
  - Only ToolJet-managed branches (those with a DB record) appear in the list

  Sort order:
  1. Default branch — always first
  2. Rest — lastCommitAt DESC (commit timestamp from GraphQL)
  3. Tiebreaker — createdAt DESC for branches with identical commit dates (e.g. two branches cut from the same source commit)
  
  Cache invalidation — Redis key is deleted on any ToolJet-initiated change:
  - createBranch
  - deleteWorkspaceBranch
  - pushWorkspace
  
  External GitHub pushes are covered by the 5-minute TTL (accepted tradeoff; webhook-based invalidation is a future improvement).

  Local Load More — no API call per click

  The full sorted branch list is fetched once on modal open and stored in Zustand. Load More is purely local — it increments visibleCount by 10 and re-slices the in-store array. No network request per
  click, no cursor, no server round-trip.

  Seeded branch auto-pull on switch

  When git sync is first configured, seedBranchesFromRemote creates WorkspaceBranch DB stubs for all remote branches but never imports workspace data (apps, modules, datasources). Switching to one of
  these stubs previously showed a completely empty workspace.

  Fix: at switch time, if the target branch has createdBy = null and is not the default branch, it is identified as a seeded stub and pullWorkspace is called before the page reloads. This imports all
  workspace data from git for that branch. Default branch is excluded because it always has data.

  Detection relies on createdBy = null — seedBranchesFromRemote creates records without a user; every user-created branch sets createdBy. No migration needed.

  N+1 fix in seedBranchesFromRemote
  
  The old code saved each branch stub in a separate manager.save() inside a loop — one INSERT per remote branch. Replaced with a single manager.insert(WorkspaceBranch, rows) batch INSERT followed by one
  re-query.

  ---
  What to test
  
  Basic modal load

  - [ ] Open Switch Branch modal — list loads promptly even for orgs with 200+ branches
  - [ ] First page shows up to 10 branches (fewer if the org has fewer than 10)
  - [ ] Default branch is always at position 1
  - [ ] Branches are ordered by most recently committed — verify commit dates in GitHub match the order shown
  - [ ] Re-open modal — list resets and re-fetches fresh data

  Load More

  - [ ] Click Load more — exactly 10 new branches append to the list
  - [ ] Click multiple times — each click adds exactly 10
  - [ ] Load More button disappears when all branches have been loaded
  - [ ] Each Load More click is instant (local — no network request)
  - [ ] Load More is hidden while a search term is active

  Search

  - [ ] Type a branch name — results appear instantly (no API call), covers branches beyond the first 10
  - [ ] Search matches partial names and handles typos (fuzzy)
  - [ ] Clear the search term — paginated list and Load More button reappear
  - [ ] Search for a branch that is not in the first 10 — it still appears in results

  Sort order edge cases

  - [ ] Create two branches from the same source branch in quick succession (y1, then y2) — y2 should appear before y1 in the list (more recently created wins tiebreaker)
  - [ ] A branch pushed today should appear above a branch pushed a month ago

  Seeded branch auto-pull — key scenario

  ▎ Setup: configure git sync on a workspace that already has multiple branches on the remote. This causes all remote branches to be seeded as empty stubs in ToolJet.

  - [ ] Open Switch Branch modal — all remote branches are listed
  - [ ] Switch to any non-default seeded branch (never manually pulled)
    - Expected: loading indicator appears, pull runs (~5–30s depending on repo size), page reloads
    - Expected: after reload, workspace is populated — apps, modules, and datasources visible
  - [ ] Switch to the default branch — no pull triggered, page reloads immediately
  - [ ] Switch to a user-created branch (not seeded) — no pull triggered
  - [ ] Switch away and back to the same seeded branch a second time — workspace data is still there (pull does not fire again since createdBy is now set after pull... note: createdBy is not updated by
  pull, so the seeded check fires again — verify that pulling a second time doesn't corrupt data or error)

  Regression — existing flows unchanged

  - [ ] Switching to an existing non-seeded branch works as before — no pull, page reloads normally
  - [ ] Delete branch — confirmation dialog appears, branch is removed from list
  - [ ] Create branch from the modal — new branch appears, modal closes
  - [ ] "View in git repo" opens the correct GitHub/GitLab/Bitbucket URL
  - [ ] App builder branch switch (non-workspace branching path) — unaffected
  - [ ] Workspace with no git sync configured — modal works normally (no remote fetch, no seeded pull logic)

  ---


ashmithamelita:
QA Checklist:
- [x] Test cases preparation
- [x] Test case review by dev and product team
- [x] Integration Testing.
- [x] Smoke Testing.
- [x] Bug Fixes Verification.
- [ ] Cross browser UI check. (Firefox, safari, chrome)
- [x] Verify features on Dark and Light mode.
- [x] Automation Testing.
- [ ] Migration testing.
- [ ] Test System Upgrade/Staging Instance(for cloud) - (Sanity + Test System Specific Cases)
- [ ] Performance Benchmark
- [x] [ ] Test feature on Beta EE

PR #16909, opened 2026-06-21
fix: Widgets - block keyboard/Tab entry on disabled components

shaurya-sharma064:
closes: https://github.com/ToolJet/tj-ee/issues/5177

### DIFF
--- a/frontend/src/AppBuilder/Widgets/Accordion/Accordion.jsx
+++ b/frontend/src/AppBuilder/Widgets/Accordion/Accordion.jsx
@@ -1,6 +1,7 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { Container as ContainerComponent } from '@/AppBuilder/AppCanvas/Container';
 import Spinner from '@/_ui/Spinner';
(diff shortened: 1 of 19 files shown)


nakulnagargade (commented):
Nice focused fix. Two minor follow-ups, neither blocking:

**1. `inert` is not actually "runtime only"**

`useDisableInert.js` and every call-site comment say *"runtime only — keeps the builder editable"*, but the hook has no `currentMode` guard — it applies `inert` whenever `isDisabled` is true, including in edit mode. In practice this is fine because `div[data-disabled='true'] { pointer-events: none }` (`custom.scss:40`) is already global/un-gated, so disabled containers already block the mouse in the builder and `inert` just mirrors that. But the comment is misleading. Suggest either gating the hook on `currentMode === 'view'`, or rewording the comment to something like "applies wherever the component is disabled; mirrors the existing pointer-events block."

**2. Native `disabled` on inputs is a behavior change worth a quick check**

`BaseInput`, Phone/Currency, Toggle(s) and Radio now get a real `disabled` attribute (previously `aria-disabled` only). Native disabled inputs are excluded from native form submission and stop firing focus/blur events. ToolJet exposes values via the store rather than the DOM, so this is probably fine — but could you confirm a disabled field still exposes its value and is still collected by the Form widget? Just a manual sanity check before merge.

PR #16919, opened 2026-06-22
Fixed custom logout url case

YuktiGoyal02:


ajith-k-v (dismissed):
Looks good overall.

It would be better if we could create a more generic custom command for updating configuration values via parameters, instead of having a dedicated `updateCustomLogoutUrl` function. We can consider refactoring it in a later iteration.

ajith-k-v (approved):

PR #16922, opened 2026-06-22
Feat: Add styles and hover on key-value component

parthy007:
[Issue](https://github.com/ToolJet/tj-ee/issues/5019)


frontend/src/AppBuilder/Widgets/KeyValuePair/KeyValuePair.jsx
```diff
@@ -58,8 +58,28 @@ export const KeyValuePair = ({
...
+    backgroundColor = 'var(--cc-surface1-surface)',
+    borderColor = 'var(--cc-weak-border)',
+    borderRadius = 6,
+    boxShadow = '0px 0px 0px 0px #00000040',
+    hoverBackgroundMode = 'auto',
+    hoverBackgroundColor = 'var(--interactive-overlays-fill-hover)',
   } = styles;
 
+  // Auto keeps the existing key-value hover color; Manual lets the user pick one.
+  const hoverBackground =
```
nakulnagargade: **Simplify hover background — the `hoverBackground` computation is redundant with the SCSS fallback**

The SCSS already declares the default:

```scss
background-color: var(--kv-hover-bg, var(--interactive-overlays-fill-hover));
```

So in **auto** mode, the JS sets `--kv-hover-bg` to exactly the value SCSS would fall back to anyway. The string `var(--interactive-overlays-fill-hover)` ends up repeated 3× (default param, manual fallback, auto branch) when the SCSS fallback is its canonical home.

We can drop the whole `hoverBackground` block and only set the CSS var when the user actually picks a manual color:

```js
const containerStyles = {
  backgroundColor,
  border: `1px solid ${borderColor}`,
  borderRadius: borderRadius ? parseFloat(borderRadius) : 0,
  boxShadow,
  ...(hoverBackgroundMode === 'manual' && hoverBackgroundColor && {
    '--kv-hover-bg': hoverBackgroundColor,
  }),
};
```

This removes the `hoverBackground` variable, the branch, and 2 of the 3 copies of the magic string — same behavior, fewer moving pieces.

PR #16981, opened 2026-06-26
fix: hydrated feature-branch app 404s on refresh in git-sync workspaces

RndmCodeGuy20:
## Summary

- **Root cause**: PR #16818 changed app import to write `apps.id` as the default-branch stub slug placeholder, breaking the "default-branch is authoritative for slug lookup" invariant. After hydration, the real slug exists only on the feature-branch `app_versions` row. Two repository methods that enforce this invariant both returned `null` for feature-branch slugs, causing 404 on any refresh or back-navigation.

- **`findAppBySlug`** (`PrivateAppAuthGuard` → `validate-private-app-access` path — the primary refresh path): step-2 fallback was restricted to `branch_id IS NULL` rows with an additional `!orgDefaultBranchId` gate. Step 1 already JOINs `WorkspaceBranch` with `is_default = true` and returned null — meaning no default-branch row anywhere on the instance holds this slug. Any row step 2 finds is therefore unambiguous. Removed the `branch_id IS NULL` filter and the `orgDefaultBranchId` gate.

- **`findByIdOrSlug`** (external API / `getAppAuthenticationConfig` path): when the default-branch re-query returns no row, fall back to the original candidate. The candidate is unambiguous when the default row's slug differs from the queried slug — no disambiguation needed.

- **`ValidAppGuard`** (`GET /apps/:id`): guard non-UUID values in the `:id` param with `isUUID()` so slugs set by `replaceEditorURL` route to `findBySlug` instead of triggering a TypeORM UUID parse error.

## Regression window

Works on v3.21.45-beta-ee, broken from v3.21.46-beta+ (PR #16818 / commit `7642847bbc`).

## Test plan

- [x] Open a git-sync workspace on a non-default branch
- [x] Open an app hydrated from another workspace's feature branch (first load should work)
- [x] Refresh the page → should no longer 404
- [x] Back-navigate to the same app → should no longer 404
- [x] Verify non-git-sync workspaces unaffected (slug resolution still works)
- [x] Verify released/public apps still accessible via `validate-released-app-access`

**QA Checklist** 

- [x] Integration Testing. 
- [x] Smoke Testing. 
- [ ] Bug Fixes Verification. 
- [ ] Cross browser UI check. (Firefox, safari, chrome)
- [ ] Verify features on Dark and Light mode.
- [ ] Test feature on CE, EE, cloud
- [x] Platform Automation Testing: [Action URL](https://github.com/ToolJet/ToolJet/actions/runs/28352920951?pr=16981) - the fix is available for failing case, it's merged in another PR, not a blocker for current release
- [x] Marketplace Automation Testing: [Action URL](https://github.com/ToolJet/ToolJet/actions/runs/28352920976?pr=16981) - `dynamodbHappyPath` spec has 3 failed cases because of UI issue
- [ ] Migration testing.
- [x] Added tested Label for PR
- [ ] Test System Upgrade/Staging Instance(for cloud) (Sanity + Test System Specific. Cases)
- [ ] Performance Benchmark

### DIFF
--- a/server/src/modules/apps/repository.ts
+++ b/server/src/modules/apps/repository.ts
@@ -107,20 +107,32 @@ export class AppsRepository extends Repository<App> {
         .orderBy('av.updated_at', 'DESC')
         .getOne();
 
-      // 2) Fall back to any row with this slug. Step 1 confirmed no
-      //    default-branch row holds it anywhere on the instance, so any
-      //    match here is unambiguous (feature-branch or legacy branchless).
+      // 2) Fall back to a branchless row — but only when the candidate app's
+      //    own workspace has no default branch (git sync still off for that
+      //    org). The branchless row is the canonical slug holder in that
+      //    case; if the org has since enabled git, step 1 would have found
+      //    the default-branch row already and this fallback would (correctly)
+      //    not apply.
+      //    Feature-branch rows are deliberately excluded here: the same slug
+      //    can appear on multiple orgs' feature branches (pulled from the same
+      //    git source), so accepting them without org context risks returning
+      //    the wrong workspace's app. The caller (PrivateAppAuthGuard) handles
+      //    feature-branch slugs via a workspace-scoped findBySlug call first.
       if (!resolvedVersion) {
         const candidate = await this.dataSource
           .getRepository(AppVersion)
           .createQueryBuilder('av')
           .innerJoinAndSelect('av.app', 'app')
           .where('av.slug = :slug', { slug })
+          .andWhere('av.branch_id IS NULL')
           .orderBy('av.updated_at', 'DESC')
           .getOne();
 
         if (candidate?.app) {
-          resolvedVersion = candidate;
+          const orgDefaultBranchId = await this.getDefaultBranchId(this.manager, candidate.app.organizationId);
+          if (!orgDefaultBranchId) {
+            resolvedVersion = candidate;
+          }
         }
       }
     } else {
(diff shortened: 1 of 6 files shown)


shah21:
I did some testing on my end, and I'm seeing the same issue. @RndmCodeGuy20 please check

### Reproduction Steps

1. Create an app in **Org A** on a new feature branch.
2. Open **Org B** and open same app  on the same feature branch. Verify it loads successfully.
3. Go back to **Org A** and make a change to the app (e.g., add a new button). Save the changes.
4. Switch to **Org B**, refresh the app, and observe that it fails to load.

**Note:** Both apps have the same slug but belong to different organizations.

server/src/modules/apps/repository.ts
```diff
@@ -107,27 +107,20 @@ export class AppsRepository extends Repository<App> {
...
+      // 2) Fall back to any row with this slug. Step 1 confirmed no
+      //    default-branch row holds it anywhere on the instance, so any
+      //    match here is unambiguous (feature-branch or legacy branchless).
       if (!resolvedVersion) {
         const candidate = await this.dataSource
           .getRepository(AppVersion)
           .createQueryBuilder('av')
           .innerJoinAndSelect('av.app', 'app')
           .where('av.slug = :slug', { slug })
-          .andWhere('av.branch_id IS NULL')
```
shah21, line 122: Without the org scope, it could pick the same app from a different org as well, right? Would that be a problem?

shah21: May be related to this -> https://github.com/ToolJet/ToolJet/pull/16981#issuecomment-4809613227

shah21 (approved):

PR #17042, opened 2026-07-03
perf: reduce DB round-trips in GET /api/apps/:id (batching + parallelization)

shah21:
## Summary

Reduces database round-trips in `GET /api/apps/:id` by eliminating N+1 query patterns and unnecessary serialization. No behavior changes — same responses, fewer queries to produce them.

## Changes

**Query batching (N queries → 1):**
- Component fetch: was queried per-page in a loop; now batched via `getAllComponentsWithPermissionsForPages(pageIds[])` — one `IN (...)` query instead of N (CE: `page.service.ts`, EE: `component.service.ts`)
- Page-permission checks: was `checkPageAvailability` per page (17 queries on a 17-page app); now `checkPagesAvailabilityBatch` — 1 batched permission fetch + 1 group-access check
- License terms: 6 separate `getLicenseTerms` calls (each an HTTP round-trip on Cloud) batched into 1 array call

**Removed unnecessary transaction wraps** on read-only paths (`findEventsForVersion`, `fetchPages`) — `dbTransactionWrap` was opening `BEGIN`/`COMMIT` around single `SELECT`s with zero isolation benefit

**Serial → parallel (`Promise.all`):**
- `stubVersion` + `draftVersion` fetch
- AI conversation `generate` + `learn`
- Page-availability checks (superseded by the batch fix above)

**Gated a migration helper off the read path:** `copyPageIndexToPageGroupIndexColumn` was running unconditionally on every read; now gated behind a cheap EXISTS check, only runs for genuinely unmigrated app versions

## Related PR

ee-server#655 (https://github.com/ToolJet/ee-server/pull/655)


server/src/modules/app-permissions/repositories/page-permissions.repository.ts
```diff
@@ -31,6 +31,16 @@ export class PagePermissionsRepository extends Repository<PagePermission> {
     }, manager || this.manager);
   }
 
+  async getPagePermissionsForPages(pageIds: string[], manager?: EntityManager): Promise<PagePermission[]> {
+    if (!pageIds.length) return [];
+    return dbTransactionWrap(async (manager: EntityManager) => {
```
akshaysasidrn: remove transaction wraps in get queries

server/src/modules/app-permissions/repositories/page-users.repository.ts
```diff
@@ -88,4 +88,20 @@ export class PageUsersRepository extends Repository<PageUser> {
       return pagePermission;
     }, manager || this.manager);
   }
+
+  // Returns the set of pagePermissionsIds where the given user has GROUP-level access
+  async checkGroupPermissionsForPages(permissionIds: string[], userId: string): Promise<Set<string>> {
+    if (!permissionIds.length) return new Set();
+    return dbTransactionWrap(async (manager: EntityManager) => {
```
akshaysasidrn: here too

akshaysasidrn (changes requested):

PR #17049, opened 2026-07-03
fix: add missing jest config for marketplace plugin tests

### LINKED ISSUE #17048: Marketplace plugins: no Jest config at marketplace/ root; tests fail with syntax errors instead of running
## Problem

Hey, while adding tests for a new Plivo operation, I noticed that plugin tests can't currently be run.

`marketplace/package.json` includes `ts-jest`, but there's no `jest.config.*` at the `marketplace/` root (or inside individual plugins). As a result, Jest falls back to its default Babel transform and fails on TypeScript syntax instead of running the tests.

```bash
cd marketplace
npx jest plugins/plivo
```

This results in a Babel parse error (`SyntaxError: Unexpected token...`).

Also, each plugin's `package.json` has:

```json
"test": "echo \"Error: run tests from root\" && exit 1"
```

but there doesn't seem to be a working root-level test setup either, so plugin tests can't currently be run.

## Suggested fix

Add a `jest.config.cjs` at the `marketplace/` root (similar to the newer plugin system):

```js
module.exports = {
  preset: 'ts-jest',
  testEnvironment: 'node',
  rootDir: '.',
};
```

Since `marketplace/package.json` uses `"type": "module"`, the config needs to be `.cjs`.

## Context

Found this while adding tests for a new Plivo operation (PR #17020).

sxryadipta:
Closes #17048 

## Problem

`marketplace/package.json` already has `ts-jest`, but there's no Jest config at the `marketplace/` root. Because of that, running plugin tests falls back to Babel and fails on TypeScript syntax.

### Repro (before)

```bash id="f8sh6u"
cd marketplace
npx jest plugins/plivo
# SyntaxError: Unexpected token...
```

## Fix

Added `marketplace/jest.config.cjs` using the existing `ts-jest` setup from the newer plugin system.

Used `.cjs` since `marketplace/package.json` has `"type": "module"`.

### DIFF
--- a/marketplace/jest.config.cjs
+++ b/marketplace/jest.config.cjs
@@ -1,0 +1,6 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  rootDir: '.',
+};


sxryadipta:
Hey maintainers, @gsmithun4 @ganesh8056 
found this by accident while writing tests for an unrelated plugin PR (plivo).. figured it's worth fixing on its own since it's probably blocking anyone else trying to actually run tests locally for any marketplace plugin, not just this one. happy to tweak the config if there's a reason it was left out (maybe CI does something different that I'm not seeing?)

sxryadipta:
kept this to the minimum; just the config file, no other changes.

PR #17065, opened 2026-07-06
Fix: restrict avatar upload to image files only

shah21:
## 📝 What this does
Hardens avatar upload security by validating file magic bytes instead of relying on the MIME type header, which can be spoofed. Files that aren't JPEG, PNG, GIF, or WebP are rejected even if they claim to be images.

## 🔀 Changes
- Added magic bytes validator that reads actual file bytes to identify image type — defeats spoofed MIME type attacks
- Replaced the header-only file type validator in the avatar upload endpoint with the new magic bytes check
- Narrowed frontend file input to allowed image types and added MIME validation before size check
- Added e2e tests for `.exe` rejection with correct and spoofed content-type

## 🧪 How to test
- [ ] Upload a valid PNG/JPEG/GIF/WebP as avatar → succeeds
- [ ] Upload a `.exe` file → 400 "Only JPEG, PNG, GIF, and WebP images are allowed"
- [ ] Rename `.exe` to `.jpg`, upload → still 400 (magic bytes catch it)


ashmithamelita:
QA Checklist:
- [ ] Test cases preparation
- [ ] Test case review by dev and product team
- [ ] Integration Testing.
- [ ] Smoke Testing.
- [x] Bug Fixes Verification.
- [x] Cross browser UI check. (Firefox, safari, chrome)
- [ ] Verify features on Dark and Light mode.
- [ ] Automation Testing.
- [ ] Migration testing.
- [ ] Test System Upgrade/Staging Instance(for cloud) - (Sanity + Test System Specific Cases)
- [ ] Performance Benchmark
- [x] [ ] Test feature on LTS EE

PR #17069, opened 2026-07-07
fix: on row click event of table does not store correct value in a variable if it depends on selectedRow and selectedRowId exposed variable

NishidhJain:
Resolves: https://github.com/ToolJet/tj-ee/issues/5256


manishkushare:
Closing this, as following will fix this https://github.com/ToolJet/ToolJet/pull/17102

PR #17123, opened 2026-07-12
Banned user signup issue fixed

### LINKED ISSUE #17086: Banned user can sign up again if they are not an existing workspace member #17008
### Environment
cloud

A banned user should not be able to sign up again using the same email address. The ban should prevent both login and new account creation until the ban is removed.

The banned user is able to successfully sign up and create a new account.

- Ban a user account.
- Ensure the banned user is no longer an existing member of the workspace.
- Attempt to sign up again using the same email address.

### Screenshots or Screencast
https://jam.dev/c/bed80ab7-2394-4867-b242-3a625fe1ac6b

Kathircpe:
closes #17086 
Added an early check in 
`whatIfTheSignUpIsAtTheWorkspaceLevel():`
`if (existingUser.status === USER_STATUS.ARCHIVED) {
    throw new NotAcceptableException('This account has been banned. Contact your administrator.');
}`

This catches banned (archived) users before the workspace membership checks (INVITED/ACTIVE), which were the only guards — and those only block if the user still has a workspace membership, which they don't after being banned from all workspaces. Without this fix, the signup flow falls through to activateUserWithPassword() which resets status to 'active', effectively un-banning them.

### DIFF
--- a/server/src/modules/onboarding/util.service.ts
+++ b/server/src/modules/onboarding/util.service.ts
@@ -191,13 +191,16 @@ export class OnboardingUtilService implements IOnboardingUtilService {
         (ou) => ou.organizationId === targetOrg?.id
       );
 
+      if (existingUser.status === USER_STATUS.ARCHIVED) {
+        throw new NotAcceptableException('This account has been banned. Contact your administrator.');
+      } 
+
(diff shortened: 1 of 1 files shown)


RudraPratik30:

PR #17137, opened 2026-07-13
Fix: Workflow Released Issue

RndmCodeGuy20:
## Summary

Triggering a workflow query from a public app (or from a module used inside a public app) returned 401/403 instead of executing. Two independent bugs were involved:

**1. `WorkflowTriggerAuthGuard` didn't honor public apps**
- `POST /workflow_executions/:id/trigger` required a valid JWT unconditionally, regardless of `app.isPublic` — unlike the sibling `POST /workflow_executions` endpoint, which already had no auth guard specifically to support public-app triggers.
- Even after bypassing the JWT check for public apps, the guard never set `request.tj_app`, so the downstream `FeatureAbilityGuard` couldn't see `app.isPublic` and rejected with 403 anyway.
- Modules are never marked public themselves, so a workflow query embedded in a module still 401'd even when the module's parent app was public.

**2. Frontend snake_case/camelCase drift for module-embedded workflow queries**
- Workflow query options (`workflowId`, `workflowVersionId`) come back snake_case (`workflow_id`, `workflow_version_id`) on the public/released/module-embedded read path, same as several other query option keys already normalized client-side — these two keys were just never added to that normalization list. This made `workflowId` resolve to `undefined` for a workflow query living inside a module, producing `POST /workflow_executions/undefined/trigger`.

## Fixes

- `server/ee/workflows/guards/workflow-trigger-auth.guard.ts`
  - Sets `request.tj_app` / `request.tj_resource_id` after resolving the app, so downstream ability checks see its public status.
  - Returns immediately for public apps instead of forcing a JWT check.
  - For module apps, mirrors `QueryAuthGuard`'s existing pattern: try JWT first, and on failure fall back to checking whether the module is embedded (via this query) in a public parent app before rejecting.
- `frontend/src/AppBuilder/_hooks/useAppData.js`
  - Added `workflowId` / `workflowVersionId` to `QUERY_OPTION_KEYS_TO_NORMALIZE`.

## Test coverage

New e2e coverage in `server/test/modules/workflows/e2e/workflow-executions.spec.ts` for `POST /api/workflow_executions/:id/trigger`:
- ✅ unauthenticated trigger succeeds when the embedding front-end app is public
- ✅ unauthenticated trigger is rejected (401) when it isn't
- ✅ unauthenticated trigger succeeds when the query lives in a module embedded in a public parent app (module itself stays non-public)
- ✅ unauthenticated trigger is rejected (401) when the module isn't embedded in any public app

All 4 new tests pass. Full spec suite: 19/20 pass — the 1 pre-existing failure (`should execute workflow with Python setup script using numpy`) is an unrelated local numpy build issue, not a regression from this change.

## QA verification steps

1. Build a public app with a workflow query, load it unauthenticated (incognito), trigger the query → should execute successfully (previously 401).
2. Build a module containing a workflow query, embed the module in a public app, load the public app unauthenticated, trigger the query from inside the module → should execute successfully (previously 401 with `.../workflow_executions/undefined/trigger`).
3. Confirm a workflow query on a **private** app still requires authentication when triggered unauthenticated (401).
4. Confirm a module embedded only in **non-public** apps still requires authentication when its workflow query is triggered unauthenticated (401).

### DIFF
--- a/frontend/src/AppBuilder/_hooks/useAppData.js
+++ b/frontend/src/AppBuilder/_hooks/useAppData.js
@@ -51,6 +51,8 @@ const QUERY_OPTION_KEYS_TO_NORMALIZE = [
   'notificationDuration',
   'disableQuery',
   'disabledMessage',
+  'workflowId',
+  'workflowVersionId',
 ];
 
 const snakeCase = (camel) => camel.replace(/[A-Z]/g, (m) => `_${m.toLowerCase()}`);
--- a/server/test/helpers/setup.ts
+++ b/server/test/helpers/setup.ts
@@ -319,7 +319,9 @@ const ENTERPRISE_TEST_TERMS: Partial<Terms> = {
   permissions: { customGroups: true },
   observability: { enabled: true },
   workflows: {
-    enabled: true, execution_timeout: 0,
+    // workflowExecutionTimeout is a literal timeout ceiling, not a sentinel like the
+    // 'UNLIMITED' fields above -- 0 means "time out immediately", not "no limit".
+    enabled: true, execution_timeout: 3600,
     workspace: { total: 'UNLIMITED', daily_executions: 'UNLIMITED', monthly_executions: 'UNLIMITED' },
     instance: { total: 'UNLIMITED', daily_executions: 'UNLIMITED', monthly_executions: 'UNLIMITED' },
   },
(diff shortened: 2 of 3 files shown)


ashmithamelita:
QA Checklist:
- [ ] Test cases preparation
- [ ] Test case review by dev and product team
- [x] Integration Testing.
- [x] Smoke Testing.
- [x] Bug Fixes Verification.
- [x] Cross browser UI check. (Firefox, safari, chrome)
- [x] Verify features on Dark and Light mode.
- [x] Automation Testing.
- [x] Migration testing.
- [x] Test System Upgrade/Staging Instance(for cloud) - (Sanity + Test System Specific Cases)
- [ ] Performance Benchmark
- [x] Test feature on LTS EE

johnsoncherian (approved):

PR #17196, opened 2026-07-19
fix: display selected value in DropdownV2 when options are grouped (#17108)

### LINKED ISSUE #17108: Dropdown does not display selected value when options are passed as grouped options
## Summary

`Dropdown` allows selecting an option from grouped options, but the selected item is not displayed in the input after selection.

The selection event appears to fire correctly, and the option is visually highlighted in the menu, but the component input stays empty.

## Steps to reproduce

1. Add a `Dropdown` component.
2. Set dynamic options to the following structure:

```js
{{
[
  {
    label: 'Fruits',
    options: [
      { value: 'apple', label: 'Apple', disable: true },
      { value: 'orange', label: 'Orange' },
      { value: 'banana', label: 'Banana' },
    ],
  },
  {
    label: 'Vegetables',
    options: [
      { value: 'carrot', label: 'Carrot' },
      { value: 'broccoli', label: 'Broccoli' },
      { value: 'spinach', label: 'Spinach' },
    ],
  },
  {
    label: 'Dairy',
    options: [
      { value: 'milk', label: 'Milk' },
      { value: 'cheese', label: 'Cheese' },
      { value: 'butter', label: 'Butter' },
    ],
  },
]
}}
```

3. Open the dropdown.
4. Select Orange or any other enabled option.

## Expected behavior
The selected option should be shown in the dropdown input.

## Actual behavior
The option gets selected/highlighted, but the dropdown input does not display the selected value.

## Notes
This looks like a DropdownV2 issue specifically with grouped options ([{ label, options: [...] }]).
The grouped structure is consistent with the docs/examples, so this seems to be a component bug rather than invalid configuration.

arhxam:
## Problem

`Dropdown` (DropdownV2) supports grouped options — `[{ label, options: [...] }]` — and the menu renders and highlights them correctly. But after selecting an option from a group, the selected value is **not shown in the input** (reported in #17108). The `onSelect` event fires and the option is highlighted, yet the control stays empty.

## Root cause

The widget builds a single **flat** `selectOptions` list and performs every value lookup against it:

- `value={selectOptions.filter((o) => o.value === currentValue)[0] ?? null}` — what react-select renders as the selected value
- `setInputValue` → `selectOptions.find((o) => o.value === value)` — the `selectedOption` exposed variable
- `selectOption` → `selectOptions...find((o) => o.value === value)` — programmatic selection
- exposed `options` variable and the `matchContent` menu-width measurement

When options are grouped, `selectOptions` contains the **group wrapper objects** (`{ label, options: [...] }`), which have no `value`. So none of these lookups ever match the selected leaf value → the input renders `null` and shows nothing.

## Fix

- `selectOptions` memo now detects grouped options and preserves the group structure react-select needs, while mapping each **leaf** option exactly like a flat option (safe label, `disable → isDisabled`, sorting within the group). Flat-options behaviour is byte-for-byte unchanged.
- Added a small pure helper `flattenSelectOptions(options)` (in `DropdownV2/utils.js`) that flattens grouped options into their leaf list, returning flat lists unchanged.
- All value lookups (`value=`, `setInputValue`, `selectOption`, exposed `options`, menu-width measurement) now resolve against the flattened leaf list, so a value selected from a group is matched and displayed.
- `findDefaultItem` also flattens, so a `default` option nested inside a group is honoured.

## How I verified

- Added unit tests for `flattenSelectOptions` (`DropdownV2/__tests__/utils.test.js`) covering flat pass-through, grouped flattening, mixed entries, resolving a selected grouped value, and non-array input.
- Verified the grouped `selectOptions` transform and value-resolution logic against the exact repro data from the issue (Fruits/Vegetables/Dairy): the selected leaf (`orange`) now resolves to its option object, and disabled leaves keep `isDisabled`.
- Confirmed flat-options mapping output is identical to the previous implementation (no regression to the common path).

Fixes #17108

### DIFF
--- a/frontend/src/AppBuilder/Widgets/DropdownV2/DropdownV2.jsx
+++ b/frontend/src/AppBuilder/Widgets/DropdownV2/DropdownV2.jsx
@@ -14,7 +14,7 @@ import CustomMenuList from './CustomMenuList';
 import CustomOption from './CustomOption';
 import Label from '@/_ui/Label';
 import cx from 'classnames';
-import { getInputBackgroundColor, getInputBorderColor, getInputFocusedColor, sortArray } from './utils';
+import { getInputBackgroundColor, getInputBorderColor, getInputFocusedColor, sortArray, flattenSelectOptions } from './utils';
 import { useMenuWidth } from './useMenuWidth';
 import { getModifiedColor, getSafeRenderableValue } from '@/AppBuilder/Widgets/utils';
 import { isMobileDevice } from '@/_helpers/appUtils';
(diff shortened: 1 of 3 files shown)


RudraPratik30:

PR #17202, opened 2026-07-20
Feature: Add restrict query run permission for data sources

akshaysasidrn:
## 📝 What this does
Adds a "Restrict query run" option to data source granular permissions so admins can block groups of users from executing queries against specific data sources — without touching per-query permissions in every app.
- [ee-server](https://github.com/ToolJet/ee-server/pull/697)
- [ee-frontend](https://github.com/ToolJet/ee-frontend/pull/543)

## 📎 References
- **Design**: [Figma — Restrict query run (DS)](https://www.figma.com/design/Gx5xbXqMk4k5E6C88yKdHZ/Authorization?node-id=7582-100221)

Closes ToolJet/tj-ee#4813

## 🏗️ Architecture

The restriction is an isolated flag outside the `Configure > Build with` hierarchy — stored as `can_run_query` (default `true`) on `data_sources_group_permissions`. Checking "Restrict query run" sets it to `false`.

Key rules:
- **Deny-wins**: if any group the user belongs to restricts a data source (via `isAll` or specific resources), execution is blocked regardless of grants from other groups
- **Admins exempt**: workspace admins always bypass
- **End-user groups** can hold a data source permission carrying only the restriction; Configure/Build with remain builder-level and are rejected
- Static data sources (runjs/runpy/tooljetdb) and anonymous public-app runs are unaffected

### Flow

```mermaid
sequenceDiagram
    participant Caller as App / Module / Workflow
    participant Util as EE data queries util (runQuery)
    participant DB

    Caller->>Util: run query
    Util->>DB: EXISTS check (user groups → data_source granular permissions with can_run_query=false, minus admin membership)
    alt restricted
        Util-->>Caller: 403 QueryError "You do not have permission to run queries on this data source"
    else allowed
        Util->>Util: execute normally
    end
```

Enforcement sits at the query execution choke point all surfaces funnel through — app builder/preview/released, modules, and workflows, which call it directly and would bypass a service-level check.

## 🔀 Changes
- Added `can_run_query` column (default `true`) to `data_sources_group_permissions` with migration
- Enforced deny-biased query-run restriction at execution time, returning a 403 to every client
- Allowed end-user groups to hold restrict-only data source permissions; builder-level fields still rejected
- Added "Restrict query run" checkbox to the permission drawer and granular access row, with end-user tooltips per design, and enabled the Data source option for end-user groups
- Fixed a latent crash where errors thrown before the abort controller was initialized were masked by a TypeError

## 🧪 How to test
- [ ] Run migrations
- [ ] Create a custom group with a data source permission, check "Restrict query run"
- [ ] Run a query on that data source as a group member — expect 403 in app builder, preview, and released app
- [ ] Trigger a workflow using the data source — expect the same failure
- [ ] Verify a workspace admin in the same group can still run queries
- [ ] Add a data source permission on the end-user group — only "Restrict query run" is editable; Configure/Build with rejected
- [ ] Uncheck the restriction — queries run again


akshaysasidrn:
## ✅ Local verification (Playwright, EE edition)

Full runtime pass on this branch against a live dev stack:

| Check | Result |
|---|---|
| Migration adds `can_run_query` (default `true`), existing rows untouched | ✅ |
| "Restrict query run" checkbox in add/edit drawer (custom group) | ✅ |
| Checkbox on granular access list row, inline toggle | ✅ |
| End-user group: Data source permission enabled, Configure/Build with disabled with tooltips, restriction editable; restrict-only permission accepted by backend | ✅ |
| Restricted builder runs query in app builder → `403 "You do not have permission to run queries on this data source"` | ✅ |
| Workspace admin **inside** the restricting group runs the same query → succeeds (admin exemption) | ✅ |
| DB row after save: `can_configure=f, can_use=t, can_run_query=f, is_all=t` | ✅ |

Verification also surfaced a latent bug: `runQuery`'s outer catch/finally called `abortCtrl.cleanup()` for errors thrown before `abortCtrl` init, masking the real error with `Cannot read properties of null (reading 'cleanup')`. Fixed in ee-server (`abortCtrl?.cleanup()`).

PR #17237, opened 2026-07-21
fix: Button column On Click event doesn't fire when Table is inside a Module (#17232)

### LINKED ISSUE #17232: Button column in Table doesn't work when Table is inside a Module used in another app
## Description

When a **Module** contains a **Table** component with a **Button**-type column, the button's `On Click` event works correctly inside the module's own editor/Preview. However, once that module is added as a component into a different (parent) application, clicking the same button in the running app does **nothing** — no error, no network request, the action simply never fires.

Other component-level events inside the same module (e.g. a plain `Button` widget's `On Click`, or the Table's own `Row Clicked` event) **do** propagate correctly to the parent app. This points to a bug specific to the `table_column` event source when a Table lives inside a Module that is itself embedded in another app (double nesting).

## Steps to Reproduce

1. Create a Module (M1).
2. Add a `Table` component (`table1`) with a Button-type column (e.g. "Actions" → button "Edit").
3. Configure the button: `On Click → Show Alert ("Hello world!")`.
4. In the module's own editor, click "Edit" in Preview — the alert appears correctly.
5. Create a separate App and add M1 as a component (via the Modules feature).
6. Open the app in Preview, click the same "Edit" button in the embedded table.

## Expected Behavior

Per the docs (self-contained queries/actions inside a module should work fully encapsulated, without any parent-app configuration), clicking "Edit" should show the alert, exactly as it does inside the module's own Preview.

## Actual Behavior

Nothing happens. No console errors, no network requests — the click event for the Button-type table column simply does not execute.

### Additional diagnostics performed

- A plain standalone `Button` component added directly to the module (not inside a table column), with the same `On Click → Show Alert` event, **works correctly** when the module is embedded and clicked from the parent app.
- Adding a `Row Clicked` event directly on `table1` (with the same Show Alert action) **also works correctly** in the embedded app — including when clicking directly on top of the (non-functional) "Edit" button itself, confirming the click does reach the row/table level.
- Only the button's own event, which uses `eventSourceType: "table_column"` internally, fails to fire once the Table is nested one level deeper (Module → App). It works fine when the Table/Module is used standalone (one level of nesting).
- Verified via direct DB inspection (`event_handlers` table) that the button's event handler is correctly configured (`target: table_column`, correct `ref` matching the button's id, `actionId: show-alert`) — this is not a configuration/duplicate-column issue on our end.

## Environment

- Self-hosted ToolJet, version `v3.20.195-lts`
- Reproduced with a Table component containing an "autogenerated columns" table plus one manually added Button-type column

## Workaround

Move the action from the button's own `On Click` to the Table's `Row Clicked` event (using `{{table1.selectedRow}}` to know which row was acted on). This works because `Row Clicked` uses the generic component-level event path rather than `table_column`. This is not a full workaround if a row needs multiple distinct buttons with different actions, since `Row Clicked` can't distinguish which button was pressed.

tabrejansary:
## Summary
Fixes #17232

When a `Table` with a **Button-type column** lives inside a Module embedded in a parent app, clicking the button did nothing — no error, no network request, no action fired.

## Root Cause

`useEvents()` in `Table.jsx` was called without a `moduleId` argument, so it always read events from the `'canvas'` scope. When the Table is inside a Module, its event handlers are stored under `eventsSlice.module['<moduleId>']` — not `'canvas'`. This caused `tableColumnEvents` to always be **empty**, so the button click had nothing to execute.

Other events (`Row Clicked`, plain `Button` widget `onClick`) were unaffected because they call `eventsSlice.fireEvent` at runtime with the correct module scope. Only `table_column` events use a **pre-fetched cache** that was broken by the missing `moduleId`.

## Fix

```diff
// frontend/src/AppBuilder/Widgets/NewTable/Table.jsx

- const allAppEvents = useEvents();
+ const allAppEvents = useEvents(moduleId);

### DIFF
--- a/frontend/src/AppBuilder/Widgets/NewTable/Table.jsx
+++ b/frontend/src/AppBuilder/Widgets/NewTable/Table.jsx
@@ -126,7 +126,7 @@ const Table = memo(
     const prevFirstRowOfTable = usePrevious(firstRowOfTable);
 
     // Get all app events. Needed for certain events like onBulkUpdate
-    const allAppEvents = useEvents();
+    const allAppEvents = useEvents(moduleId);
 
     const shouldAutogenerateColumns = useRef(false);
     const hasDataChanged = useRef(false);


tabrejansary:
@cla-assistant recheck

PR #17245, opened 2026-07-22
fix: module permission gaps (git-sync branch pins, permission-update no-op, embedded-module access)

RndmCodeGuy20:
## Summary
- **Module permission edits silently ignored (CE):** `updateResourcePermissions` switched on `granularPermissions.type` but never handled `ResourceType.MODULE`, so toggling Edit/Build-with off for a group's module permission updated `GranularPermissions.isAll` but never touched `apps_group_permissions` — builders kept whatever access was originally seeded. EE already had this case wired; CE didn't.
- **Git-sync module 404 on branch-name pins:** `resolveModuleRef` only resolved non-UUID pins as tagged/published version names. A `ModuleViewer` pinned to "follow branch X" (rather than a fixed release) 404'd whenever that branch wasn't itself a published tag — most commonly right after a fresh git-sync pull, before the module has any published version. Added the missing branch-name-shorthand tier, mirroring the handling `resolveAllModuleViewersForVersion` already had.
- **Builders can't view/edit modules embedded in apps they own:** `FeatureAbilityGuard` gated purely on the module's own permission, so a builder with edit access to a parent app had no way to open a module it embeds unless they also had a direct module permission. Adds a structural bypass — verified via a new `AppsRepository.isModuleEmbeddedInApp` query — that grants view-only access when the module is genuinely embedded in an app the requester can edit.
- **`isModuleEmbeddedInApp` scoping fix:** the embed check above initially matched a `ModuleViewer` from *any* historical version of the parent app, including versions superseded after a release. Narrowed to the app's current/editing version — and fixed a follow-on regression where that narrowing excluded `version_type='branch'` rows, which broke the check for git-sync-enabled workspaces (the actively-edited version there is a branch row with no `current_version_id`).
- **Module dashboard menu shown under view-only access:** the app-card "…" menu mounted unconditionally for modules (unlike apps, which gate it on edit/delete rights), leaving build-with-only (view) users with an empty popover instead of no icon at all. Mirrored app behavior.

Note: this branch also briefly changed `resolveModuleRef` to fall back to the latest version on the branch instead of 404ing when a pin is orphaned (e.g. a legacy UUID-only pin with no matching row). That was reverted — the 404 is intentional: the frontend (`useAppData.js`) catches it to surface a debugger-log error + toast telling the builder their module pin is broken, so they fix it themselves rather than silently rendering the wrong version.

## Test plan
- [x] Unit tests: `app-version.ability.spec.ts`, `modules-ability.spec.ts` updated for the embedded-module bypass
- [x] New e2e suite: `module-embedded-in-app.spec.ts`, `apps-repository.spec.ts` coverage for `isModuleEmbeddedInApp`
- [x] Manually reproduced and re-verified both the permission no-op and the git-sync 404 against `tooljet_development`
- [ ] CI run on this branch

### DIFF
--- a/frontend/src/AppBuilder/_hooks/useAppData.js
+++ b/frontend/src/AppBuilder/_hooks/useAppData.js
@@ -95,9 +95,8 @@ const useAppData = (
   const setEditorLoading = useStore((state) => state.setEditorLoading);
   const setApp = useStore((state) => state.setApp);
   const user = useStore((state) => state.user);
-  // The top-level app containing this module instance — used to let the backend
-  // verify this module read is happening in the context of an app the user can
-  // edit, even when they have no standalone module permission.
+  // Forwarded to the backend so a module-embedding parent app can grant view access
+  // (see appVersion.service.js getAll/getModuleVersionData).
   const parentAppId = useStore((state) => state.appStore?.modules?.['canvas']?.app?.appId);
   const setCurrentVersionId = useStore((state) => state.setCurrentVersionId);
   const currentVersionId = useStore((state) => state.currentVersionId);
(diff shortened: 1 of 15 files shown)


shah21 (approved):
Reviewed the module pin-related logic. LGTM

PR #17248, opened 2026-07-22
Fix: Restore events when components are restored using undo operation

devanshu052000:
Closes - https://github.com/ToolJet/tj-ee/issues/4519

This pull request refactors and enhances the undo/redo functionality in the `undoRedoSlice.js` store. The main improvements include making undo/redo handlers asynchronous, improving how patches are processed (especially for restoring event handlers), and updating the logic for adding and removing components to better support batch operations and event restoration.

**Undo/Redo Functionality Enhancements:**

* Made `handleUndo` and `handleRedo` asynchronous to properly handle operations that require awaiting, such as batch component and event restoration. [[1]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cL27-R27) [[2]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cL131-R168)

**Patch Processing and Event Restoration:**

* Updated `processPatches` to be asynchronous and to support restoring event handlers when components are added back during undo/redo. This includes batching component and event restoration and updating the logic to handle errors gracefully. [[1]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cL39-R58) [[2]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cR105-R147)
* Added support in `filterAndFormatPatches` for a new `restoreEvents` operation to capture event handler restoration when undoing/redoing component changes.

**Component Addition & Batch Operations:**

* Changed logic from tracking `componentIdsToAdd` to `componentsToAdd` and refactored the component addition flow to use batch operations, ensuring that events are only restored for successfully added components. [[1]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cL73-R75) [[2]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cR105-R147)

**Error Handling Improvements:**

* Enhanced error handling in undo/redo operations by reverting stack changes if patch processing fails, preventing stack corruption. [[1]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cL39-R58) [[2]](diffhunk://#diff-55a3fd03381c180f761a9a139f33df929831d14662503087bb0183c8898f076cL143-R182)

**Code Clean-up:**

* Removed commented-out and obsolete code in patch processing for clarity and maintainability.

Let me know if you have questions about how event restoration works or why these changes were needed!


frontend/src/AppBuilder/_stores/slices/undoRedoSlice.js
```diff
@@ -230,6 +263,13 @@ const filterAndFormatPatches = (patches) => {
         });
       }
     }
+
+    if (op === 'replace' && path[0] === 'eventsSlice' && path[1] === 'module' && path[3] === 'events') {
+      changeStack.push({
+        op: 'restoreEvents',
+        value,
+      });
+    }
```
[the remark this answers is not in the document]
devanshu052000: won't fix: no current code path produces an element-level events patch inside a withUndoRedo block; it's speculative hardening for a contract that doesn't exist.

frontend/src/AppBuilder/_stores/slices/undoRedoSlice.js
```diff
@@ -24,7 +24,7 @@ export const createUndoRedoSlice = (set, get) => {
   };
 
   return {
-    handleUndo: () => {
+    handleUndo: async () => {
```
kavinvenkatachalam, line 27: We need to ensure undo/redo operations happen in the correct order. Since handleUndo/handleRedo are now async (awaiting `addComponentToCurrentPage` / `saveComponentChanges` inside `processPatches`), there's a risk that a second undo/redo starts before the first one's async work has resolved — not just reordering, but two calls mutating the same canvas/undo-stack state concurrently. `canUndo/canRedo` aren't updated until after the await completes, so nothing currently blocks a second invocation (e.g. from key auto-repeat on Cmd+Z, or a fast double-click). Can you confirm whether there's a guard against overlapping calls, or should one be added?

frontend/src/AppBuilder/_stores/slices/undoRedoSlice.js
```diff
@@ -70,65 +70,108 @@ export const createUndoRedoSlice = (set, get) => {
...
+      let addedComponents = null;
+      if (componentsToAdd && componentsToAdd.length > 0) {
+        addedComponents = await get().addComponentToCurrentPage(componentsToAdd, 'canvas', {
+          saveAfterAction: false,
+          skipFormUpdate: true,
+          skipUndoRedo: true,
+        });
+
+        if (!addedComponents || Object.keys(addedComponents).length === 0) {
+          throw new Error('Undo/redo restore failed: no components were added');
```
kavinvenkatachalam, line 110: This scenario needs testing: in processPatches, if `componentIdsToDelete` has already been applied earlier in the call, and the later `addComponentToCurrentPage` call throws (e.g. the "no components were added" check), the catch in handleUndo/handleRedo only reverts the undo/redo stack — it doesn't undo the deletion that already ran. That leaves state half-applied while the stack acts as if the operation never happened, which could cause JSON corruption or unexpected behavior on the next undo/redo.

kavinvenkatachalam (changes requested):

PR #17269, opened 2026-07-23
fix: utility bug fixes (regex escaping, null guards, sort mutation, JSON parse safety)

PedroHenrique0713:
## Summary

Six utility bug fixes across the ToolJet frontend.

### 1. Regex crash in HighLightSearch (`HighLightSearch.jsx`)
User-supplied search terms containing regex metacharacters (`(`, `)`, `[`, `*`, etc.) caused uncaught `SyntaxError` when interpolated directly into `new RegExp()`. Fix: escape regex special characters using the same `escapeRegExp` pattern already used in `DropdownV2/utils.js`.

### 2. Double-fault in HTTP client (`http-client.js`)
When a request failed with non-JSON response (HTML error page, 502 gateway), `JSON.parse(text)` inside the catch block threw a second uncaught exception. Fix: wrap in try/catch, fall back to raw text.

### 3. Crash on empty array in PDF generation (`generate-file.js`)
`Object.keys(value[0])` throws `TypeError` when `value` is an empty array (valid query result). Fix: early return when array is empty.

### 4. TypeError on stale module ID (`componentsSlice.js`)
Four methods accessed `modules[moduleId].pages` without null guard. If moduleId was stale/invalid, accessing `.pages` on `undefined` threw TypeError. Fix: add optional chaining on module access.

### 5. Sort mutation corrupts Zustand state (`dataQueriesStore.js`, `dataQuerySlice.js`)
`sortByAttribute` called `data.sort()` in-place, mutating Zustand state arrays and preventing React from detecting changes. Fix: spread-copy before sorting.

### 6. Sort mutation in DropdownV2 (`utils.js`)
`sortArray` had the same in-place mutation, affecting dropdown, multiselect, and tag input widgets. Fix: spread-copy before sorting.

## Testing
- All changes are defensive guards — existing behavior preserved for valid inputs
- Regex escaping matches the existing pattern in `DropdownV2/utils.js:52-57`

### DIFF
--- a/frontend/src/AppBuilder/Widgets/DropdownV2/utils.js
+++ b/frontend/src/AppBuilder/Widgets/DropdownV2/utils.js
@@ -72,8 +72,9 @@ export const highlightText = (text = '', highlight) => {
 };
 
 export const sortArray = (arr, sort) => {
+  const sorted = [...arr];
   if (sort === 'asc') {
-    return arr.sort((a, b) => {
+    return sorted.sort((a, b) => {
       const labelA = typeof a.label === 'string' ? a.label : '';
       const labelB = typeof b.label === 'string' ? b.label : '';
       return labelA.localeCompare(labelB);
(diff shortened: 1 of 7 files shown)


PedroHenrique0713:
Closing this to keep my open queue honest rather than leaving it stale against the repo: it has been open for 40 days with no review, and the changes are small and self-contained.

If any of them is useful, the diff can be lifted directly or this can be reopened, and I am happy to rebase it onto current master in that case. Thanks for maintaining the project.

PR #17270, opened 2026-07-23
fix: null guards, XSS in mentions, Boolean() copy-paste bug, and stripTrailingSlash safety

PedroHenrique0713:
## Summary

Four bug fixes in ToolJet frontend utilities.

### 1. `findProp` crashes on null intermediate values (`utils.js`)
The guard `obj !== undefined` allows `null` to pass through, then `typeof null["a"]` throws TypeError. Fix: use `!= null` to catch both null and undefined.

### 2. `hightlightMentionedUserInComment` XSS vulnerability (`utils.js`)
User content between parentheses was injected directly into HTML without escaping. Fix: escape HTML content using DOM `createTextNode` before inserting into the span.

### 3. `isExpectedDataType` Boolean() copy-paste bug (`utils.js`)
`Boolean()` without arguments always returns `false`. Should be `Boolean(data)` to coerce the input value. Fix: pass `data` to `Boolean()`.

### 4. `stripTrailingSlash` crashes on non-string input (`utils.js`)
`str.replace()` throws TypeError on null/undefined. Fix: return empty string for falsy input.

### DIFF
--- a/frontend/src/AppBuilder/Widgets/DropdownV2/utils.js
+++ b/frontend/src/AppBuilder/Widgets/DropdownV2/utils.js
@@ -72,8 +72,9 @@ export const highlightText = (text = '', highlight) => {
 };
 
 export const sortArray = (arr, sort) => {
+  const sorted = [...arr];
   if (sort === 'asc') {
-    return arr.sort((a, b) => {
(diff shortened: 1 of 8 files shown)


PedroHenrique0713:

PR #17271, opened 2026-07-23
fix: findProp null guard, resolve type guard, verifyConstant return, cookie encoding

PedroHenrique0713:
## Summary

Four bug fixes in ToolJet frontend utilities.

| # | File | Bug |
|---|---|---|
| 1 | `utils.js` | `findProp` bracket path access without null guard → TypeError |
| 2 | `utils.js` | `resolve()` crashes on non-string `data` (number/object/undefined) |
| 3 | `utils.js` | `verifyConstant` returns `undefined` instead of `[]` when no errors |
| 4 | `cookie.js` | Cookie values not `encodeURIComponent`'d → semicolons corrupt cookie header |


PedroHenrique0713:

PR #17272, opened 2026-07-23
fix: escapeHtml null guard and handleResponse JSON.parse safety

PedroHenrique0713:
## Summary

Two bug fixes in ToolJet frontend.

### 1. `escapeHtml` renders `null` as text (`utils.js`)
`document.createTextNode(null)` renders the literal string `"null"`, producing misleading HTML when the input is null/undefined. Fix: return empty string for null/undefined input.

### 2. `handleResponse` crashes on non-JSON responses (`handle-response.js`)
`JSON.parse(text)` without try/catch in 2 response handlers. When the server returns non-JSON (HTML error pages, raw text), the unhandled rejection crashes the error handling path entirely. Fix: wrap in try/catch, propagate a useful error on parse failure.


PedroHenrique0713:

PR #17274, opened 2026-07-23
fix(useBatchedUpdateEffectArray): prevent effect re-running on every render

PedroHenrique0713:
## Summary

`useBatchedUpdateEffectArray` used `items.map(item => item.dep)` as its effect dependency array. Since `.map()` creates a new array reference on every render, React always sees the deps as changed, causing the effect to run on every single render — defeating the batching purpose entirely.

**Fix:** Use `JSON.stringify` to create a stable dependency key that only changes when actual item deps change.


PedroHenrique0713:

PR #17301, opened 2026-07-27
Feature: Wire combined unit + e2e coverage into CI

RndmCodeGuy20:
## 📝 What this does
Wires the existing (but previously unused) coverage tooling into CI: unit and e2e Jest coverage now merge into one report, with a compact collapsible summary posted to the PR and the full HTML report attached as a downloadable artifact. Along the way, found and fixed a bug where e2e coverage was silently collecting zero files.

## 🔀 Changes
- Fixed e2e coverage collecting zero files — jest's rootDir pointed at `server/test`, and the v8 coverage provider hard-filters coverage to files under rootDir, dropping every `src/` and `ee/` file before it even started counting
- Added a `coverage-report` CI job that merges unit + e2e coverage, uploads the full HTML report as an artifact, and posts a collapsed summary to the PR — kept out of the CI Gate pass/fail logic (informational only)
- Replaced the undeclared `nyc` dependency (network-fetched every run) with a small committed merge script using istanbul libraries already in the tree
- Fixed duplicate HTML rendering from the `lcov` reporter and a ~12MB bloat in jest's `--json` results caused by an embedded coverage map
- Carries a small pending fix from earlier work on this branch (test credentials in unit tests, `.gitignore` entry for `AUDIT_LEDGER.md`)

## 🧪 How to test
- [ ] Apply the `run-ci` label to trigger the full pipeline
- [ ] Confirm the PR comment shows a collapsed "Coverage" block with real percentages
- [ ] Download the `coverage-report-html` artifact and open `index.html` — annotated source should render for both `src/` and `ee/` files
- [ ] Confirm the `CI Gate` status is unaffected even if the `coverage-report` job fails


server/package.json
```diff
@@ -221,6 +221,9 @@
     "eslint-plugin-jest": "^28.13.3",
     "eslint-plugin-prettier": "^5.4.1",
     "globals": "^16.2.0",
+    "istanbul-lib-coverage": "^3.2.2",
```
akshaysasidrn: License check trips here — istanbul packages are BSD-3-Clause, and the allowlist only takes exact MIT/Apache-2.0. These exact packages already ship as jest transitives (hoisted to `node_modules/` root today), so declaring them adds zero new legal exposure. Options, best first:

1. Add BSD-3-Clause to the allowlist — real policy fix, it's permissive.
2. Drop these three lines and lean on jest's transitives — works today (verified hoisting), but it's a phantom dependency: a jest major bump or lockfile shuffle can nest them, and since the merge step is `continue-on-error`, the resulting "Cannot find module" would silently drop coverage from PR comments. Same failure-hiding shape as the old `npx nyc 2>/dev/null` bug this PR fixed. If forced down this path, leave a comment in merge-coverage.mjs.
3. Per-package exemption in the check — worst, invites a pile of exemptions.

Declaring is the right instinct either way — undeclared-import-via-hoisting is the exact anti-pattern the nyc removal killed.

server/scripts/merge-coverage.sh
```diff
@@ -8,42 +8,20 @@
...
-  --temp-dir "$MERGE_DIR/merged" \
-  --reporter=html --reporter=lcov --reporter=json \
-  --report-dir="$COMBINED_DIR" \
-  --exclude='test/**' \
-  --exclude='migrations/**' \
-  --exclude='data-migrations/**' \
-  --exclude='dist/**'
-
-rm -rf "$MERGE_DIR"
+node scripts/merge-coverage.mjs --out "$COMBINED_DIR" --reporters json,html,lcovonly,json-summary coverage-unit coverage-e2e
```
akshaysasidrn: This whole file is now: existence check + this one node call. And merge-coverage.mjs already does the same check itself ("No coverage-final.json found… Run the test suites with --coverage first", exit 1). Suggest deleting the file and pointing `test:cov:merge` straight at the node script — one less indirection layer.

server/scripts/run-e2e.sh
```diff
@@ -211,15 +213,12 @@ fi
...
   printf "\033[1m━━━ Merging coverage ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━\033[0m\n\n"
-  mkdir -p .coverage/merged
-  npx nyc merge .coverage .coverage/merged/coverage-final.json 2>/dev/null
-  npx nyc report \
-    --temp-dir .coverage/merged \
-    --reporter=html --reporter=lcov --reporter=json \
-    --report-dir=coverage-e2e 2>/dev/null
-  cp .coverage/merged/coverage-final.json coverage-e2e/coverage-final.json 2>/dev/null
-  printf "\033[32mCoverage report written to coverage-e2e/\033[0m\n"
+  node scripts/merge-coverage.mjs --out coverage-e2e --reporters json,html,lcovonly,json-summary .coverage
```
akshaysasidrn: In CI, only `coverage-e2e/coverage-final.json` survives this merge — the second stage (`test:cov:merge`) reads just the json. The HTML tree (~1300 annotated files), lcov, and summary rendered here get thrown away with the runner, so CI pays for the full HTML render twice per run.

Suggest making the reporter list a flag: CI passes `json` only, local standalone `test:e2e:cov` keeps the rich default (that's the audience that actually opens the e2e-only report).

.github/workflows/ci.yml
```diff
@@ -697,8 +688,15 @@ jobs:
             export E2E_STEP_URL="$UNIT_STEP_URL"
             export SCOPE=$(cat /tmp/ci-results/tj-scope.txt 2>/dev/null || echo "not recorded")
           else
-            export UNIT_STEP_URL=$(job_url "server-unit")
-            export E2E_STEP_URL=$(job_url "server-e2e")
+            export UNIT_STEP_URL=$(job_url "Test Suite · server")
+            export E2E_STEP_URL="$UNIT_STEP_URL"
+            artifacts_json=$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/artifacts?per_page=100" 2>/dev/null || echo '{}')
```
akshaysasidrn: This whole lookup (api call + jq + fallback branch) can go — `upload-artifact@v4+` outputs `artifact-url` directly. Give the coverage upload step an `id:`, expose it as a `test-server` job output, and read `needs.test-server.outputs.…` here — same channel the new `unit`/`e2e` outcomes already use.

scripts/render-coverage.mjs
```diff
@@ -0,0 +1,103 @@
...
+function areaFor(filePath) {
+  const marker = '/server/';
+  const idx = filePath.lastIndexOf(marker);
+  const rel = idx >= 0 ? filePath.slice(idx + marker.length) : filePath;
+  const parts = rel.split('/').filter(Boolean);
+  return parts.slice(0, 2).join('/') || rel;
+}
+
+function fmtPct(covered, total) {
+  if (total === 0) return '100%';
```
akshaysasidrn: 0/0 shown as 100% — a zero-line area reads as fully covered. Return `—` instead.

.github/workflows/ci.yml
```diff
@@ -360,15 +284,65 @@ jobs:
...
       - name: Upload e2e results for the PR comment
         if: ${{ !cancelled() }}
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: full-e2e-results
           path: /tmp/tj-e2e-json/
           if-no-files-found: ignore
           retention-days: 1
+      - name: Merge coverage
```
akshaysasidrn, line 323: Edge case: unit fails → e2e step skipped → this still runs (`!cancelled()`) and merges unit-only data, but the PR comment will say "unit + e2e merged". You already export `steps.e2e.outcome` as a job output — pass it into the render env and label honestly ("unit only — e2e skipped").

akshaysasidrn (commented):
Verified against run 30327615499: the merged summary artifact shows **949 `src/` + 357 `ee/` files, 40.42% lines (63,131/156,154)** — the e2e `rootDir` fix genuinely unblocked e2e coverage (was zero files before), and the consolidated numbers are sane. The merge approach is sound: shared `collectCoverageFrom` denominator, same-workspace paths so istanbul dedupes per-file, per-statement merge so a line covered by either suite counts.

Inline: 1 blocker (license check on the istanbul deps), rest are cuts/nits. Also noting the pre-push revert and README no-op cleanup already landed in the latest commits — good. Remember to retarget once `test/external-api-helper-port` merges.

For a later iteration: coverage **delta vs base** in the PR comment is what makes the number actionable — absolute % on a legacy codebase is hard to judge; `40.4% (+0.3)` isn't.

PR #17311, opened 2026-07-28
Fix: Prevent inactive FX from resolving

parthy007:
[Issue](https://github.com/ToolJet/tj-ee/issues/5255)


frontend/src/AppBuilder/_stores/transientStore.js
```diff
@@ -16,6 +16,24 @@ export default create(
...
+        get().fxExpressionStash.set(key, expression);
+      },
+
+      // Read-and-delete, so a restored expression cannot be re-applied by a later fx toggle.
+      takeFxExpression: (key) => {
+        if (!key) return undefined;
+        const stash = get().fxExpressionStash;
+        const expression = stash.get(key);
+        stash.delete(key);
+        return expression;
```
kavinvenkatachalam: Can we move it out of the store and use simple session storage?

frontend/src/AppBuilder/RightSideBar/Inspector/Components/PopoverMenu/components/OptionDetailsPopover.jsx
```diff
@@ -22,6 +22,9 @@ const OptionDetailsPopover = forwardRef(
     const iconVisibility = item?.iconVisibility;
     const { dataCy: dataCyPrefix, popoverFields, popoverClassName, popoverTitle } = config;
 
+    // No componentId here, so two Menu-style widgets share a stash entry per option index.
+    const getFxStashKey = (property) => `${componentType}-options-${index}-${property}`;
```
kavinvenkatachalam: This can cause a collision when two TreeSelect widgets on the same canvas share a stash slot at the same option index. If you toggle fx off for widget A’s option 0 (“visible”) and then toggle fx on for widget B’s option 0 (“visible”), widget B receives A’s stashed expression.

Fix this by using the componentId, either obtained from selectedComponent or passed as a prop.

frontend/src/AppBuilder/RightSideBar/Inspector/Components/TreeSelect/components/TreeSelectItemPopover.jsx
```diff
@@ -41,6 +41,9 @@ const TreeSelectItemPopover = forwardRef(
       onDeleteItem(item.value, parentValue);
     };
 
+    // No componentId here, so two TreeSelects sharing an item value share a stash entry.
+    const getFxStashKey = (property) => `treeSelect-${parentValue ?? 'root'}-${item.value}-${property}`;
```
kavinvenkatachalam: The same data leak problem will occur here as well.

kavinvenkatachalam (changes requested):

frontend/src/AppBuilder/_stores/transientStore.js
parthy007: The stash is only an undo for an accidental fx-off, so "gone on reload" is the lifetime we actually want. SessionStorage survives reloads, and entries are only cleared when the user toggles fx back on.

frontend/src/AppBuilder/RightSideBar/Inspector/Components/Steps.jsx
```diff
@@ -241,7 +242,8 @@ export function Steps({ componentMeta, darkMode, ...restProps }) {
               paramLabel={'Disable'}
               paramName={'disable'}
               onChange={(value) => handleLabelChange('disabled', { value }, index)}
-              onFxPress={(active) => handleOnFxPress(active, index, 'disabled')}
+              onFxToggle={(active, newValue) => handleOnFxPress(active, index, 'disabled', newValue)}
```
kavinvenkatachalam, line 250: We added the extra parameter but haven’t updated the function definition to accept it.

frontend/src/AppBuilder/RightSideBar/Inspector/Components/Tags.jsx
```diff
@@ -293,7 +293,8 @@ export const Tags = ({
               paramLabel={'Tag visibility'}
               onChange={(value) => handleValueChange(item, { value }, 'visible', index)}
               paramName={'visible'}
-              onFxPress={(active) => handleOnFxPress(active, index, 'visible')}
+              onFxToggle={(active, newValue) => handleOnFxPress(active, index, 'visible', newValue)}
```
kavinvenkatachalam, line 303: Same here, we added the extra parameter but haven’t updated the function definition to accept it.

frontend/src/AppBuilder/RightSideBar/Inspector/Components/Table/ProgramaticallyHandleProperties.jsx
```diff
@@ -162,6 +162,7 @@ export const ProgramaticallyHandleProperties = ({
           callbackFunction(index, 'fxActiveFields', resultFxActiveFields);
         }}
         fxActive={calcFxActiveState(props, property)}
+        fxStashKey={`${component?.id}-columns-${index}-${property}`}
```
kavinvenkatachalam: `ProgramaticallyHandleProperties` got a `fxStashKey` but was not migrated to `onFxToggle`, so it falls into the legacy branch at `SingleLineCodeEditor.jsx:734-735`:

```js
onFxPress(false);        // write 1
onChange(frozenValue);   // write 2  <- new in this PR
```

Before this PR that branch was only `onFxPress(false)`, so there was a single write and nothing to conflict. The gate is `persistsFxState` (`SingleLineCodeEditor.jsx:665`), which is true whenever `onFxPress` is passed — so every unmigrated call site opted into the two-write path automatically.

Both of those props resolve to the **same function** here (`ProgramaticallyHandleProperties.jsx:155-165`):

```jsx
onChange={(value) => callbackFunction(index, property, value)}
onFxPress={(active) => {
  const resultFxActiveFields = calculateFxActiveFields(active, props, property);
  callbackFunction(index, 'fxActiveFields', resultFxActiveFields);
}}
```

and `callbackFunction` (`useListItemManager.js:112`, and `useActionButtonManager.js:43` for action buttons) does not do a targeted key update — it rebuilds and persists the **entire array** from a render-time snapshot:

```js
const updateProperty = (index, property, value) => {
  let newItems = [...items];                                   // `items` = useMemo over props
  newItems[index] = { ...newItems[index], [property]: value };
  paramUpdated({ name: 'columns' }, 'value', newItems, ...);   // saves ALL columns
};
```

React doesn't re-render mid-handler, so both calls copy the *same stale* `items`:

```
original: { columnVisibility: '{{queries.q1.data.show}}', fxActiveFields: ['columnVisibility'] }

write 1: copy original, set fxActiveFields = []
         -> saves { columnVisibility: '{{queries...}}', fxActiveFields: [] }                    OK

write 2: copy original AGAIN (unchanged, no re-render), set columnVisibility = '{{true}}'
         -> saves { columnVisibility: '{{true}}', fxActiveFields: ['columnVisibility'] }        write 1 reverted
```

Write 2 lands last and wins, carrying the old `fxActiveFields` with it.

**Effect:** click fx off on any Table column property and the fx button pops straight back on. The literal is frozen, but the toggle doesn't persist.

**Steps to reproduce:** Table → any column → Visibility → fx on, enter `{{true}}` → fx off. Button re-activates.

**Fix:** migrate it to `onFxToggle` so `fxActiveFields` and the property value are written in one rebuild. `Inspector.jsx:349` (`paramsUpdated`) already exists for multi-field writes if that helps.

kavinvenkatachalam (changes requested):
@parthy007 requested some code changes and added another nitpick below -

Most stash keys use the array index:

```js
`${componentId}-options-${index}-${property}`
```

Toggle fx off on option 0, drag it to position 2, toggle fx on → the lookup hits "position 0" and restores whichever option now sits there, or nothing.

`TreeSelectItemPopover.jsx:47` already gets this right with `item.value`. The rest have stable ids available and should use them:

| Location                                  | Currently           | Stable id                                     |
| ----------------------------------------- | ------------------- | --------------------------------------------- |
| `OptionDetailsPopover.jsx:27`             | `options-${index}`  | `item.value`                                  |
| `Select.jsx:370,392,413`                  | `options-${index}`  | `item.value`                                  |
| `Tags.jsx:297`                            | `options-${index}`  | `item.value`                                  |
| `Steps.jsx:225,246`                       | `steps-${index}`    | `item.id`                                     |
| `TabComponent.jsx:370,390,412`            | `tabItems-${index}` | `item.id` (the handler already matches on it) |
| `ProgramaticallyHandleProperties.jsx:165` | `columns-${index}`  | column `name`                                 |

---

PR #17325, opened 2026-07-29
Feat/sso env config

Pratush613:
frontend : [PR](https://github.com/ToolJet/ee-frontend/pull/565)
server : [PR](https://github.com/ToolJet/ee-server/pull/729)

# Summary

This PR introduces support for configuring **OIDC, SAML, and LDAP SSO providers entirely through environment variables**, eliminating the need to store sensitive credentials in the database.

Instead of persisting secrets, the database now stores only minimal metadata such as `use_env_config` and other bookkeeping fields.

This implementation also includes:

* Enterprise-only licensing enforcement.
* Automatic boot-time enablement for valid environment-based configurations.
* Self-healing recovery when environment configurations become invalid and are later corrected.

This improves security by keeping credentials outside the database while enabling automated deployments and infrastructure-as-code workflows.

---

# What's Included

## 1. Instance-Level OIDC (Single Provider)

Supports configuring a single instance-level OIDC provider using standard environment variables.

### Required

* `OIDC_CLIENT_ID`
* `OIDC_WELL_KNOWN_URL`
* `OIDC_NAME`
* `OIDC_GRANT_TYPE`

  * Supported values:

    * `authorization_code`
    * `pkce`
  * Any other value is rejected and treated as an invalid configuration.

### Conditionally Required

* `OIDC_CLIENT_SECRET`

  * Required unless `OIDC_GRANT_TYPE=pkce`.

### Optional

* `OIDC_CUSTOM_SCOPES`
* `OIDC_CODE_VERIFIER` (automatically generated for PKCE if omitted)
* `OIDC_ENABLE_GROUP_SYNC`
* `OIDC_CLAIM_NAME`
* `OIDC_GROUP_MAPPING`

If `OIDC_ENABLE_GROUP_SYNC` is omitted but both `OIDC_CLAIM_NAME` and `OIDC_GROUP_MAPPING` are configured, group synchronization is automatically enabled.

### Workspace-Specific Group Sync Overrides

* `OIDC_GROUP_SYNC_{WORKSPACE}_CLAIM_NAME`
* `OIDC_GROUP_SYNC_{WORKSPACE}_MAPPING`

Where `{WORKSPACE}` is the normalized workspace name (uppercase, apostrophes removed, spaces replaced with underscores).

**Notes**

* Instance-level OIDC supports only one provider.
* Indexed provider keys such as `OIDC_1_CLIENT_ID` are **not** supported.

---

## 2. Workspace-Level OIDC (Multiple Providers)

Supports configuring multiple OIDC providers per workspace using `WORKSPACE_OIDC_CONFIG`.

### Behavior

* Configuration must be a plain JSON object keyed by workspace.
* Workspace lookup first matches the normalized workspace name and then falls back to the workspace slug.
* Bare keys (`OIDC_CLIENT_ID`, etc.) are supported for single-provider workspaces.
* Indexed keys (`OIDC_1_*`, `OIDC_2_*`, etc.) are supported for multiple providers.
* Required and optional fields are identical to the instance-level implementation.
* Grant-type validation and group-sync inference follow the same rules as instance-level OIDC.
* Provider slots are protected against collisions. Attempting to reuse an occupied slot raises a configuration error instead of silently overwriting another provider.

---

## 3. Workspace-Level SAML

Configured using `WORKSPACE_SAML_CONFIG`.

### Required

* `SAML_IDP_METADATA`

  * Validates:

    * Well-formed XML
    * Signing certificate
    * HTTP-Redirect SSO endpoint
* `SAML_NAME`

### Optional

* `SAML_ENABLE_GROUP_SYNC`
* `SAML_GROUP_SYNC_GROUP_ATTRIBUTE`

If `SAML_GROUP_SYNC_GROUP_ATTRIBUTE` is configured but the toggle is omitted, group synchronization is automatically enabled.

---

## 4. Workspace-Level LDAP

Configured using `WORKSPACE_LDAP_CONFIG`.

### Required

* `LDAP_HOST_NAME`
* `LDAP_PORT`
* `LDAP_BASE_DN`
* `LDAP_NAME`

### Optional

* `LDAP_SSL`
* `LDAP_SSL_CERTIFICATE`
* `LDAP_CLIENT_KEY`
* `LDAP_CLIENT_CERTIFICATE`
* `LDAP_SERVER_CERTIFICATE`
* `LDAP_ENABLE_GROUP_SYNC`

Unlike OIDC and SAML, LDAP does not infer group synchronization automatically and defaults to `false` when omitted.

---

# Enterprise Licensing

Every enable path, whether manual or automatic, requires:

* A valid Enterprise license.
* `WORKSPACE_ENV` entitlement.
* The corresponding SSO entitlement (OIDC, SAML, or LDAP).

---

# Boot-Time Auto Enable

On every server startup:

* Valid environment configurations are automatically enabled.
* Auto-enable only occurs for providers that have never been manually modified.
* Once a provider has been manually enabled or disabled, that decision is permanently preserved, including across future license renewals.

---

# Self-Healing

Automatically managed providers recover from configuration failures without manual intervention.

If an environment-managed provider becomes invalid because of:

* Invalid configuration
* Unreachable identity provider
* Licensing changes

its database record is removed.

When the configuration becomes valid again, the provider is automatically recreated and enabled during the next server startup.

Self-healing is driven from the database rather than the currently parsed `.env`, ensuring removed workspace configurations are correctly disabled instead of remaining stale.

Manually managed providers are never deleted and always preserve administrator intent.

This implementation also aligns SAML and LDAP boot-time behavior with the existing OIDC implementation.

---

# Validation

Environment-based configuration follows the same validation flow as manually configured providers.

### OIDC

* Performs `Issuer.discover(wellKnownUrl)`
* Validates supported grant types (`authorization_code` and `pkce`)

### SAML

* Validates metadata structure
* Validates signing certificate
* Validates HTTP-Redirect endpoint

### LDAP

* Validates configuration structure

# Example Environment Configuration

## Instance-Level OIDC

```env
OIDC_CLIENT_ID=your-client-id
OIDC_CLIENT_SECRET=your-client-secret
OIDC_WELL_KNOWN_URL=https://idp.example.com/.well-known/openid-configuration
OIDC_NAME=CorpOIDC
OIDC_GRANT_TYPE=authorization_code
OIDC_ENABLE_GROUP_SYNC=true

OIDC_GROUP_SYNC_MY_WORKSPACE_CLAIM_NAME=groups
OIDC_GROUP_SYNC_MY_WORKSPACE_MAPPING={"engineering":"admin"}
```

---

## Workspace-Level OIDC

```env
WORKSPACE_OIDC_CONFIG='{
  "my-workspace": [
    {
      "OIDC_1_NAME": "first",
      "OIDC_1_CLIENT_ID": "id-1",
      "OIDC_1_CLIENT_SECRET": "secret-1",
      "OIDC_1_WELL_KNOWN_URL": "https://idp1.example.com/.well-known/openid-configuration",
      "OIDC_1_GRANT_TYPE": "authorization_code"
    },
    {
      "OIDC_2_NAME": "second",
      "OIDC_2_CLIENT_ID": "id-2",
      "OIDC_2_CLIENT_SECRET": "secret-2",
      "OIDC_2_WELL_KNOWN_URL": "https://idp2.example.com/.well-known/openid-configuration",
      "OIDC_2_GRANT_TYPE": "pkce"
    }
  ]
}'
```

---

## Workspace-Level SAML

```env
WORKSPACE_SAML_CONFIG='{
  "my-workspace": {
    "SAML_IDP_METADATA": "<EntityDescriptor...>",
    "SAML_NAME": "CorpSAML",
    "SAML_ENABLE_GROUP_SYNC": "true",
    "SAML_GROUP_SYNC_GROUP_ATTRIBUTE": "groups"
  }
}'
```

---

## Workspace-Level LDAP

```env
WORKSPACE_LDAP_CONFIG='{
  "my-workspace": {
    "LDAP_HOST_NAME": "ldap.example.com",
    "LDAP_PORT": "389",
    "LDAP_BASE_DN": "dc=example,dc=com",
    "LDAP_NAME": "CorpLDAP",
    "LDAP_SSL": "false",
    "LDAP_ENABLE_GROUP_SYNC": "true"
  }
}'
```

---

# Test Plan

An automated end-to-end test suite (9 self-contained tests) covers:

* Boot-time auto-enable
* Self-healing
* Manual enable/disable persistence
* Enterprise licensing
* OIDC
* SAML
* LDAP

### Setup

* [ ] Verify an Enterprise license is active.
* [ ] Confirm workspace slug and normalized name match the environment configuration.

### Instance-Level OIDC

* [ ] Valid configuration auto-enables after restart.
* [ ] Manual disable persists across restarts.
* [ ] Invalid Well-Known URL disables the provider automatically.
* [ ] Fixing the configuration re-enables it.
* [ ] Previously manually disabled providers remain disabled.
* [ ] Invalid `OIDC_GRANT_TYPE` values are rejected.

### Workspace-Level OIDC

* [ ] Single-provider configuration auto-enables.
* [ ] Multiple providers auto-enable independently.
* [ ] Correcting invalid providers enables them independently.
* [ ] Manual disable preserves provider slot assignments.
* [ ] Slot conflicts surface the expected configuration error.

### SAML / LDAP

* [ ] Valid configurations auto-enable.
* [ ] Invalid SAML metadata prevents enablement.
* [ ] LDAP enables with valid configuration.
* [ ] SAML group sync defaults correctly.

### Self-Healing

* [ ] Automatically managed providers recover after configuration issues are fixed.
* [ ] Removing an entire workspace configuration disables the provider correctly.

### Licensing

* [ ] Enterprise licensing is required for both manual and automatic enablement.

### GUI Compatibility

* [ ] Existing GUI-configured providers remain unaffected.
* [ ] Switching between GUI and environment configuration preserves existing values.
* [ ] Confirmation dialog behaves correctly.

---

# Regression Areas

Please verify the following continue behaving as expected:

* Git Sync workspace resolution
* Multi-provider OIDC creation flow
* Enterprise licensing checks
* LDAP authentication flow
* SAML and LDAP boot-time behavior
* Password domain allow/restrict validation


server/test/modules/login-configs/e2e/sso-env-config.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+    await ssoConfigsRepository.delete({ sso, organizationId: orgId });
+  };
+
+  beforeAll(async () => {
+    ({ app } = await initTestApp({ edition: 'ee', plan: 'enterprise', freshApp: true }));
+    ssoConfigsRepository = getEntityRepository(SSOConfigs);
+
+    baselineWorkspaceOidcConfig = requireEnv('WORKSPACE_OIDC_CONFIG');
+    baselineWorkspaceSamlConfig = requireEnv('WORKSPACE_SAML_CONFIG');
+    baselineWorkspaceLdapConfig = requireEnv('WORKSPACE_LDAP_CONFIG');
```
akshaysasidrn: These `requireEnv` calls throw in `beforeAll` unless `WORKSPACE_OIDC_CONFIG` / `WORKSPACE_SAML_CONFIG` / `WORKSPACE_LDAP_CONFIG` are set, but `.env.test` is gitignored and CI never creates one, so the whole suite is red for anyone but the machine it was written on. Could the spec set its own fixtures instead?

```ts
beforeAll(() => {
  process.env.WORKSPACE_OIDC_CONFIG = JSON.stringify([{ 'test-ws': [ ... ] }]);
  ...
});
afterAll(() => { /* restore */ });
```

Same pattern the workspace-OIDC tests further down already use for their overrides, so it's just extending that to the baseline.

server/src/modules/login-configs/constants/feature.ts
```diff
@@ -21,5 +22,9 @@ export const FEATURES: FeaturesConfig = {
     [FEATURE_KEY.UPDATE_INSTANCE_SSO]: {},
     [FEATURE_KEY.UPDATE_INSTANCE_GENERAL_CONFIGS]: {},
     [FEATURE_KEY.GET_INSTANCE_SSO]: {},
+    [FEATURE_KEY.SAVE_OIDC_ENV_CONFIGS]: { license: LICENSE_FIELD.OIDC },
+    [FEATURE_KEY.SAVE_INSTANCE_OIDC_ENV_CONFIGS]: { license: LICENSE_FIELD.OIDC },
+    [FEATURE_KEY.SAVE_SAML_ENV_CONFIGS]: { license: LICENSE_FIELD.SAML },
+    [FEATURE_KEY.SAVE_LDAP_ENV_CONFIGS]: { license: LICENSE_FIELD.LDAP },
```
akshaysasidrn: These gates check the provider license but not the env-mapping one. The PRD puts the feature under both:

> Should be under the existing `workspaceEnv` toggle. Available on Enterprise plan only.

`LICENSE_FIELD.WORKSPACE_ENV` exists but isn't declared here, so a tenant licensed for OIDC without `workspaceEnv` passes this guard. The EE service re-checks it internally, so nothing leaks today, but the guard is advertising a weaker gate than the feature actually has. Either add `WORKSPACE_ENV` here or drop a one-liner saying the second check lives in the EE service, so the next reader doesn't have to trace it.

server/src/modules/login-configs/repository.ts
```diff
@@ -56,4 +56,42 @@ export class SSOConfigsRepository extends Repository<SSOConfigs> {
...
+    if (!existing) {
+      // Nothing to disable if the row doesn't exist yet — creating one just to immediately mark
+      // it disabled would leave a useless, permanently-disabled orphan provider behind.
+      if (!useEnvConfig) return null;
+      return this.save(
+        this.create({
+          organizationId,
+          sso,
+          configScope: ConfigScope.ORGANIZATION,
+          configs: {} as any,
```
akshaysasidrn: Two things on this signature:

1. `sso: SSOType = SSOType.SAML` — a call site that forgets the third argument silently writes a SAML row. The EE SAML service already has call sites that omit it (flagged there). Making the parameter required turns that bug class into a compile error for the cost of typing `SSOType.SAML` twice.

2. `configs: {} as any` — the column is non-nullable json, so the cast is understandable, but `{} as SSOConfigs['configs']` says the same thing without opening the `any` door.

server/src/modules/login-configs/dto/index.ts
```diff
@@ -69,3 +69,29 @@ export class InstanceConfigsUpdateDto {
...
+
+export class UpdateOidcEnvConfigDTO {
+  @IsBoolean()
+  useEnvConfig: boolean;
+
+  // Optional — absent when toggling env-config on a brand-new, never-saved provider draft;
+  // the toggle creates the row itself in that case (see toggleOidcEnvConfig).
+  @IsOptional()
+  @IsString()
+  configId?: string;
```
akshaysasidrn: `configId` is a UUID primary key, so `@IsUUID()` over `@IsString()` gets validation for free. Also: this DTO and the SAML/LDAP/instance ones below differ only in name; one `UpdateEnvConfigDTO` would cover all four routes.

server/src/modules/login-configs/service.ts
```diff
@@ -127,9 +127,11 @@ export class LoginConfigsService implements ILoginConfigsService {
       }
     } else {
       // Other SSO types (single config per organization)
+      // `configs` is omitted from the write payload when not provided (e.g. an env-config-managed
+      // SAML config toggling `enabled` only) so it doesn't null out the existing stored configs.
       ssoConfig = await this.ssoConfigsRepository.createOrUpdateSSOConfig({
         sso: type,
-        configs,
+        ...(configs !== undefined && { configs }),
```
akshaysasidrn, line 132: This guard protects the non-OIDC create path from an undefined `configs`, but the OIDC create paths above still pass `configs` straight through. An env-managed OIDC provider created without configs would insert `undefined` into a non-nullable json column. If OIDC can't reach that state, a one-line comment saying why would save the next reader the trace; if it can, the same guard belongs there.

akshaysasidrn (changes requested):

server/src/modules/login-configs/constants/feature.ts
Pratush613: This guard intentionally only checks the provider license — license here is a single LICENSE_FIELD, so it can't also require WORKSPACE_ENV without reworking FeatureConfig to accept multiple fields, which felt out of scope here. The EE service (LoginConfigsService's toggle*EnvConfig methods) re-checks WORKSPACE_ENV and licenseType === 'enterprise' before actually enabling env-config, so nothing leaks today. Leaving this as-is for now.

PR #17339, opened 2026-07-30
Fix: Persist event handler order in builder by sorting handlers on index

parthy007:
[Issue](https://github.com/ToolJet/tj-ee/issues/5317)

## Problem

Handler order reverted after drag-and-drop and became inconsistent after refresh for both **component** and **query** events.

Each handler stores an `index`, which represents its trigger order. The backend persisted this correctly, but the read path never used it.

```mermaid
flowchart LR
    A["Drag row"] --> B["New index saved (ok)"]
    B --> C["Merge by id into existing array slots"]
    C --> D["List renders array order, not index"]
    D --> E["Snaps back"]
    F["Refresh"] --> G["SELECT with no ORDER BY"]
    G --> H["Order undefined"]
```

As a result:
- The configured order was saved correctly.
- The UI continued rendering the previous array order.
- After refresh, the order depended on the database response since it wasn't explicitly sorted.

Execution already sorts handlers by `index`, so events fired in the correct order while the Event Manager displayed the old order.

---

## Changes

| Area | Change |
|------|--------|
| **EventManager.jsx** | Sort handlers by `index` before rendering so the UI matches execution order. |
| **reorderEvents** | Apply the reordered list to local state immediately after drag-and-drop instead of waiting for the PUT request. |
| **reorderEvents** | Persist only handlers whose `index` changed instead of updating the entire group. |
| **duplicateHandler** | Assign `max(index) + 1` instead of `events.length` to avoid index collisions after deletions. |
| **Draggable** | Use `event.id` as the React key instead of the array position so row state remains attached to the correct handler. |
| **eventsSlice.js** | Fixed `updateAppVersionEventHandlers`, which read from `getModuleEvents('canvas')` but wrote to `module[moduleId]`, causing a module's events to be replaced with canvas events in the module editor. |

---

## Backend Changes

None. Frontend-only changes.


frontend/src/AppBuilder/_stores/slices/eventsSlice.js
```diff
@@ -213,7 +213,8 @@ export const createEventsSlice = (set, get) => ({
       get().eventsSlice.updateEventsField('actionsUpdatedLoader', false, moduleId);
       get().eventsSlice.updateEventsField('eventsUpdatedLoader', false, moduleId);
       set((state) => {
-        const eventsInState = state.eventsSlice.getModuleEvents('canvas');
+        // Was hardcoded to 'canvas', overwriting module events with canvas events in the module editor
+        const eventsInState = state.eventsSlice.getModuleEvents(moduleId);
```
sviat-barbutsa: Nice fix. I think the module-editor path still falls back to canvas, though.
EventManager gets moduleId from useModuleContext(), but calls useEvents() and useEventActions() without passing it. 
Both hooks default to "canvas", and useEventActions forwards that captured value here. So during a module edit, this line still reads canvas events and the module-isolation fix does not take effect.

Could you, please,  pass moduleId into those hooks, along with the related loader selectors, and add a small regression test proving that reordering module events does not change canvas events?

sviat-barbutsa (commented):
I left an inline comment. Please check it when you have a chance. Thanks!

frontend/src/AppBuilder/RightSideBar/Inspector/EventManager.jsx
devanshu052000, line 410: The PR surfaces a latent server-side index-corruption bug — server/src/modules/apps/services/event.service.ts:369-381

`updateEventsOrderOnDelete` decrements index for every event sharing `sourceId`, ignoring `target` and `event.ref`:

```
const allEvents = await this.findAllEventsWithSourceId(sourceId);   // sourceId only
const eventsToUpdate = allEvents.filter((event) => event.index > deletedIndex);
```

A Table's component-target handlers and its table_column / table_action handlers share one sourceId, same for Navigation per-menu-item ref events. Deleting a handler in one panel shifts indices in the other. It's also not scoped by appVersionId.

While the UI ignored index this was invisible. After this PR it becomes a visible reorder-after-refresh — the exact symptom the PR is fixing, just for a narrower case.

Also, in frontend `deleteAppVersionEventHandler` only calls `removeEvent`, so surviving events keep stale indices in the store.

**Suggested change -** 

We can remove the decrementing of index logic on event deletion from backend as it is one of the source for creation of duplicate event indices and now the newly generated indices in the frontend are always unique so even on direct deletion of the event, the rest of the events will maintain the order of execeution and display.

frontend/src/AppBuilder/_stores/slices/eventsSlice.js
devanshu052000: **[nitpick]** This function only runs when an Event manager is rendered which means module id will always be 'canvas' so basically this is a no-op. Correct in principle but can remove the comment

devanshu052000 (changes requested):

frontend/src/AppBuilder/RightSideBar/Inspector/EventManager.jsx
```diff
@@ -1121,8 +1125,15 @@ export const EventManager = ({
...
+
+    // Reflect the new order right away, the store merge only catches up once the PUT resolves
+    setEvents(reorderedEvents);
+
     updateAppVersionEventHandlers(
-      reorderedEvents.map((event) => ({
+      changedEvents.map((event) => ({
         event_id: event.id,
         diff: event,
       })),
```
devanshu052000: `setEvents(reorderedEvents)` is now applied before the PUT, but reorderEvents doesn't await and `updateAppVersionEventHandlers` has no try/catch. If the request fails, the list keeps showing the new order while the DB keeps the old one — silently wrong until refresh. The unhandled rejection also leaves actionsUpdatedLoader / eventsUpdatedLoader stuck true (pre-existing, but the optimistic update is what makes it user-visible). Wrap the call and restore the pre-drag events on error.

devanshu052000 (changes requested):

frontend/src/AppBuilder/_stores/slices/eventsSlice.js
parthy007: I looked into it and I don't think it changes behavior today. Editor takes `moduleId = 'canvas'` (`AppBuilder.jsx:28`) and no caller passes anything else, so the builder is always on canvas, including when editing a module, since a module opens as its own app rather than nested inside a host. Non-canvas module ids only appear in the Viewer (`Viewer.jsx:166`) for modules embedded via the `ModuleViewer` widget, and the Viewer doesn't render the inspector, so `EventManager` never sees one.

So there's no canvas/module crossover for a regression test to catch here. I've kept the `moduleId` change in the slice since it's the right shape and matches the rest of the actions in that file, but I've dropped the comment that claimed it fixed a module-editor bug. Happy to revisit if the builder ever renders under a non-canvas module id.

devanshu052000 (dismissed):

johnsoncherian (approved):

frontend/src/AppBuilder/_stores/slices/eventsSlice.js
sviat-barbutsa: Hi @parthy007! Thanks for checking and explaining the current mounting path! After your explanations I see that's your truth. The problem was: I traced the hook `defaults` but missed that `EventManager` only renders in the builder, where moduleId remains canvas. Removing the module-editor claim makes sense. Thanks for the clarification.

PR #17345, opened 2026-07-30
[Feat] : Custom component  library core

manishkushare:
Resolves: https://github.com/ToolJet/tj-ee/issues/5127

ee-frontend: https://github.com/ToolJet/ee-frontend/pull/593
ee-server: https://github.com/ToolJet/ee-server/pull/733


kavinvenkatachalam (changes requested):
- Can we add FE tests for this?

 - Also, can we add `sandbox="allow-scripts"` to the iframe? Right now the shell is same-origin, so component code can read cookies/localStorage and call the API as the logged-in user. Sandboxing gives it an opaque origin and cuts all of that off. **Note**: it won't work as a one-line change — once the origin is opaque, the vendor React modules and the bundle/CSS endpoints become cross-origin, so they'll need `Access-Control-Allow-Origin: *`. And we shouldn't add `allow-same-origin` alongside it, since that combination defeats the sandbox.

johnsoncherian (approved):

kavinvenkatachalam (approved):

johnsoncherian (approved):

PR #17359, opened 2026-07-31
Fix: Module Related Fixes

RndmCodeGuy20:
## 📝 What this does
Fixes pinned module references failing to resolve ("Module version not found") in git-sync-enabled workspaces, caused by two stale assumptions in the resolver from before branch_id became mandatory on every app_versions row.

## 🔀 Changes
- Derived `isGitSyncEnabled` from the org's actual git-sync config instead of `!!defaultBranch`, which is now always true
- Replaced `branchId: IsNull()` filters in pinned-lookup tiers with the resolved default branch id, since branch_id is no longer nullable
- Added e2e coverage for pinned resolution by name, UUID, and draft-only tiers in both non-git-sync and git-sync-enabled workspaces

## 🧪 How to test
- [ ] Pin a module to a specific version in a git-sync-enabled workspace, push, then pull into a fresh workspace on the same repo — module should load instead of showing "Module version not found"
- [ ] `task test:e2e-filter -- module-ref-resolution`


shah21 (approved):
Module related changes are looking good to me

PR #17387, opened 2026-08-04
Fix: resolve legacy name-pinned module draft in git-sync workspaces

RndmCodeGuy20:
## 📝 What this does
Fixes module resolution 404s for modules pinned by version name or UUID (legacy format, pre-dating git-sync/branching on the workspace) when that pinned version is still an unreleased draft on the default branch — without relaxing resolution for genuinely git-native drafts.

## 🔀 Changes
- Removed the git-sync-only gate blocking the default-branch draft fallback in the name-based (Tier 0) module version lookup
- Scoped that fallback (and an equivalent one added to the UUID-based Tier 1 lookup) to rows with `isSynced: false` — legacy versions that predate git-sync on the workspace — so actively synced drafts still require a PUBLISHED match
- Added regression coverage for a legacy name-pinned draft in a git-sync-enabled workspace, plus a guardrail test proving a genuinely synced draft still 404s by name

## 🧪 How to test
- [ ] In a git-sync/branching-enabled workspace, pin an app's ModuleViewer to a module version by name (not UUID) while that version is still a DRAFT that predates git-sync being enabled
- [ ] Open the app and confirm the module fetches successfully instead of 404ing with "Module version not found"
- [ ] Confirm published/saved module version pins still resolve as before
- [ ] Confirm a module version drafted *after* git-sync was enabled still does not resolve by name unless published

### DIFF
--- a/server/src/modules/versions/module-ref.util.ts
+++ b/server/src/modules/versions/module-ref.util.ts
@@ -143,11 +143,13 @@ export async function resolveModuleRef(
   }
 
   // Tier 0 — versionName lookup (non-UUID ref, cross-workspace stable).
-  // git-sync: default-branch PUBLISHED only — branched rows are always DRAFTs by constraint
-  //   chk_app_versions_branched_implies_draft, so checking PUBLISHED first prevents a
-  //   same-named branch DRAFT shadowing the real release.
-  // non-git-sync: also check default-branch any-status — DRAFTs are valid specific pins,
-  //   and a plain workspace's only version may still be an unpublished DRAFT.
+  // Check PUBLISHED first regardless of git-sync: a released version by that name should
+  // always win over a same-named unreleased draft. Then fall back to any status (DRAFT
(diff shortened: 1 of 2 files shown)


server/src/modules/versions/module-ref.util.ts
```diff
@@ -161,18 +163,16 @@ export async function resolveModuleRef(
...
-          appId: moduleApp.id,
-          name: versionName,
-          branchId: defaultBranch.id,
-          versionType: AppVersionType.VERSION,
-          isStub: false,
-        },
-      });
-      if (defaultBranchDraft) return defaultBranchDraft;
-    }
+    const defaultBranchDraft = await manager.findOne(AppVersion, {
```
shah21, line 167: Please make sure the version pinning logic and rendering work correctly, especially when there is no explicit pinning to a published version for Git-synced, non-Git-synced, and unsynced apps.

If there are any side effects, guard the check based on whether the app is unsynced or not.

PR #17401, opened 2026-08-04
fix: prevent eye icon from overlapping API key text in encrypted input fields

### LINKED ISSUE #17283: OpenAI: Eye icon overlaps masked API key text in datasource configuration
The eye icon should be properly aligned with sufficient padding so it does not overlap the masked API key text. The entire masked value should remain visible without UI obstruction

The eye (visibility) icon overlaps the masked API key text at the end of the input field.

2. Navigate to datesource page
3. Add Open AI datasource
4. Enter API Key and verify the eye icon

### Screenshots or Screencast

Image

waterWang:
### Problem
The eye icon in encrypted/password input fields (e.g., OpenAI API key) overlaps the masked text, making it partially unreadable.

### Root cause
1. The parent container `.tj-app-input-wrapper` is missing `position: relative`, so the absolute-positioned eye icon positions relative to the wrong ancestor.
2. The `<input>` element lacks `padding-right`, so the input text extends underneath the icon.

### Fix
1. Added `position: relative` to `.tj-app-input-wrapper` so the absolute-positioned icon is correctly scoped to the wrapper.
2. Added `input { padding-right: 28px; }` inside `.tj-app-input-wrapper` to ensure the text doesn't overlap with the icon.

Closes #17283

### DIFF
--- a/frontend/src/_styles/theme.scss
+++ b/frontend/src/_styles/theme.scss
@@ -9,7 +9,6 @@
 @import "./typography.scss";
 @import "./designtheme.scss";
 @import "./dropdown-custom.scss";
-@import "./editor/react-select-search.scss";
 @import "./ui-operations.scss";
 @import "./license.scss";
 @import "react-loading-skeleton/dist/skeleton.css";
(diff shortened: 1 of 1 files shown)


RudraPratik30:

PR #17407, opened 2026-08-04
fix: validate expression whitespace and filter empty path segments in Inspector [Fixes #17240]

### LINKED ISSUE #17240: [Workspace Constants] Invalid expression with whitespace resolves successfully and Inspector copies incorrect variable path

- `{{constants. null_value}} `should display a syntax/validation error since the expression contains invalid whitespace after `.`.
- The Inspector should copy the variable path as:`constants.null_value` without the trailing `.`.

- The expression `{{constants. null_value}} `(constant with spaces) resolves to the constant's value instead of displaying an error.
- When copying the variable path from the Inspector, the copied path is `constants.null_value. `(contains an extra trailing .), resulting in an invalid expression if pasted directly.

2. Navigate to the workspace constant page 
3. Add a constant with a valid key and value
4. Navigate to the dashboard and create an app
5. Click on Inspector 
6. Use the added workspace constant and check the expression 

### Screenshots or Screencast
Inspector expression: 
Image

Constant expression:
Image

Image

waterWang:
## Changes

Two fixes for workspace constants expression handling:

1. **Expression validation**: Updated `validateMultilineCode` to detect invalid whitespace immediately after a dot (e.g., `{{constants. null_value}}`). This now correctly returns a validation error instead of silently resolving.

2. **Inspector copy path**: Added `.filter(Boolean)` to `formatPathForCopy` to remove empty path segments caused by trailing dots. This ensures the copied path is `constants.null_value` instead of `constants.null_value.`.

Fixes #17240

### DIFF
--- a/frontend/src/AppBuilder/LeftSidebar/LeftSidebarInspector/utils.js
+++ b/frontend/src/AppBuilder/LeftSidebar/LeftSidebarInspector/utils.js
@@ -108,7 +108,7 @@ export const copyToClipboard = (data, includeQuotes = true) => {
 export const formatPathForCopy = (path) => {
   const isArray = path.includes('.');
   if (isArray) {
-    const pathArray = path.split('.');
+    const pathArray = path.split('.').filter(Boolean);
     const newPath = pathArray.map((item) => {
       if (!isNaN(item) && item !== '') {
         return `[${item}]`;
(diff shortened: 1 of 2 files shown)


RudraPratik30:

PR #17411, opened 2026-08-04
fix: save CodeHinter value on unmount to prevent data loss on popover rootClose

### LINKED ISSUE #12019: [Dropdown/Multiselect component] : When editing option properties (label), the values do not get saved when clicking on the next option or add new option
### Environment
ToolJet Cloud

When editing the option properties of a dropdown or multiselect component, the edited values of the options should be saved when user directly clicks on the next option to continue editing

As of now the user has to click somewhere inside the option popup first and then click on the next option for the edited value to be saved (The option values or new options are not saved otherwise)

1. Add a dropdown or a multiselect component
2. Click on any one of the options in the property panel
3. Edited the option value
4. Click on the next option or add a new option
5. Go back to the previously edited option and check if latest changes have been saved

### Screenshots or Screencast
https://jam.dev/c/e7e1fb91-b9fb-4b66-b32d-444062da8a3a

waterWang:
## Summary

This PR fixes #12019 where editing option properties (label, value, etc.) in the Dropdown/Multiselect component property panel does not save the edited value when the user clicks directly on the next option (or adds a new option).

## Root Cause

The SingleLineCodeEditor (CodeHinter) only persists its value via onChange on blur. When the OptionDetailsPopover closes via Bootstrap rootClose mechanism (e.g., clicking the next option), the CodeMirror editor blur event does not fire reliably because the active element at the time of close is the newly clicked element, not the CodeMirror editor.

## Fix

Added a useEffect cleanup in the EditorInput component that saves the latest editor value (tracked via a ref) when the component unmounts. This ensures that any pending edits in the CodeHinter are persisted to the parent state regardless of how the popover is closed.

## Changes

- frontend/src/AppBuilder/CodeEditor/SingleLineCodeEditor.jsx: Added latestValueRef to track the current editor value, and a cleanup effect that calls onBlurUpdate with the latest value on unmount.

### DIFF
--- a/frontend/src/AppBuilder/CodeEditor/SingleLineCodeEditor.jsx
+++ b/frontend/src/AppBuilder/CodeEditor/SingleLineCodeEditor.jsx
@@ -264,7 +264,7 @@ const SingleLineCodeEditor = ({ componentName, fieldMeta = {}, componentId, modu
 };
 
 const EditorInput = ({
-  currentValue,
+  currentValue: currentValueProp,
   setCurrentValue,
   setFocus,
   validationType,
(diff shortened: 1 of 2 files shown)


RudraPratik30:

PR #17417, opened 2026-08-05
Fix: Allow nesting component expression resolution

parthy007:
[Issue](https://github.com/ToolJet/tj-ee/issues/5329)


kavinvenkatachalam (approved):

frontend/src/AppBuilder/_stores/slices/componentsSlice.js
```diff
@@ -48,6 +48,23 @@ const queryRerunTimers = new Map();
...
+    } else if (current && typeof current === 'object') {
+      for (const entry of Object.values(current)) pending.push(entry);
+    }
+  }
+  return false;
+};
+
+// Arrays are objects too, so entries that are themselves arrays do not count as per-key resolvable values
+const hasObjectEntry = (items) => items.some((item) => item && typeof item === 'object' && !Array.isArray(item));
+
```
devanshu052000: Can we shift this code outside of `Query re-run on dependency change` section?

devanshu052000 (changes requested):

devanshu052000 (changes requested):
**High priority** — nested dependency edges are never cleaned up, so deleted child items come back

Before this change, nested `children` was handed to `updateResolvedValues` as a whole array, so no dependency edges existed for nested content at all. Now we register one edge per nested leaf, but nothing ever removes them — `setComponentProperty`'s array branch returns at `componentsSlice.js:2231`, before the only `removeDependency` call at line 2274.

Repro:

1. Create a query `q` returning `{ name: "Alpha" }`, run it
2. Drop a Navigation widget. Set Group 1 → Option 4's label to `{{queries.q.data.name}}` → shows "Alpha"
3. Delete Option 4 → canvas is correct, one child left
4. Change `q` to return `{ name: "Beta" }`, run it

Option 4 reappears as a menu item labelled "Beta". The stale edge survives the delete, and `applyDependencyUpdate` does `lodashSet(entity, ['properties','menuItems',2,'children',1,'label'], value)`, which recreates the index. It renders because `isItemVisible` sees `visible === undefined` → `undefined !== true` → visible. The ghost has no `id`, so selection misbehaves, and it sticks around until the next `menuItems` edit.

Top-level items already ghost the same way on `develop`, so the bug class predates this PR — but the nested case is new, and group children get added/deleted/reordered far more often than top-level items.

For the fix: `removeNode` won't work as-is because it matches `${path}.` and misses array notation (`menuItems[0].label`). Needs a sibling in `DependencyClass` that also matches `${path}[`, called from the array branch of `setComponentProperty` before `checkValueAndResolve` re-adds the edges.

devanshu052000:
**Low priority** — per-leaf cost on large trees

Flagging for later. Once a single binding exists anywhere in a subtree, `hasDynamicValue` passes and then every key of every node goes through `updateResolvedValues` → `findNearestSubcontainerAncestor` + `parsePropertyPath` + `lodashSet` — on every page load and every property edit, and once per row inside a ListView. On top of that, `cloneDeep` runs again at each recursion level in addition to the parent's clone, so cloning is O(nodes × depth).

Large inspector-built TreeSelect trees are the case I'd watch. Might be worth profiling ~1.5k nodes with one binding; if it shows up, hoisting the clone (pass the already-cloned subtree into the recursive call instead of re-cloning) plus a node-count cap would cover it. Happy to leave as-is if the numbers are small.

parthy007:
> **Low priority** — per-leaf cost on large trees
> 
> Flagging for later. Once a single binding exists anywhere in a subtree, `hasDynamicValue` passes and then every key of every node goes through `updateResolvedValues` → `findNearestSubcontainerAncestor` + `parsePropertyPath` + `lodashSet` — on every page load and every property edit, and once per row inside a ListView. On top of that, `cloneDeep` runs again at each recursion level in addition to the parent's clone, so cloning is O(nodes × depth).
> 
> Large inspector-built TreeSelect trees are the case I'd watch. Might be worth profiling ~1.5k nodes with one binding; if it shows up, hoisting the clone (pass the already-cloned subtree into the recursive call instead of re-cloning) plus a node-count cap would cover it. Happy to leave as-is if the numbers are small.

Profiled the shape you suggested (~1.5k nodes, depth 4, one binding).

Small correction on `O(nodes × depth)`: `hasDynamicValue` keeps static subtrees out of the recursion, so clones follow the depth of the *dynamic path*, not node count — that tree produced 3 clones, not one per level per node.

Hoisted the clone: nested calls now reuse the parent's cloned subtree (trailing optional param, other call sites untouched). 3 clones → 1, ~8.9ms → ~3.0ms. Verified output identical and the definition still unmutated.

Left the per-leaf writes (~11ms) alone — batching them means changing `updateResolvedValues` / `setAllValueToComponent`, which every widget's array properties use. Too much blast radius for a rare case. Skipping non-dynamic keys isn't viable either: it leaves nested values as raw `{ value: … }` objects, the exact mismatch this PR fixes.

parthy007:
> **High priority** — nested dependency edges are never cleaned up, so deleted child items come back
> 
> Before this change, nested `children` was handed to `updateResolvedValues` as a whole array, so no dependency edges existed for nested content at all. Now we register one edge per nested leaf, but nothing ever removes them — `setComponentProperty`'s array branch returns at `componentsSlice.js:2231`, before the only `removeDependency` call at line 2274.
> 
> Repro:
> 
> 1. Create a query `q` returning `{ name: "Alpha" }`, run it
> 2. Drop a Navigation widget. Set Group 1 → Option 4's label to `{{queries.q.data.name}}` → shows "Alpha"
> 3. Delete Option 4 → canvas is correct, one child left
> 4. Change `q` to return `{ name: "Beta" }`, run it
> 
> Option 4 reappears as a menu item labelled "Beta". The stale edge survives the delete, and `applyDependencyUpdate` does `lodashSet(entity, ['properties','menuItems',2,'children',1,'label'], value)`, which recreates the index. It renders because `isItemVisible` sees `visible === undefined` → `undefined !== true` → visible. The ghost has no `id`, so selection misbehaves, and it sticks around until the next `menuItems` edit.
> 
> Top-level items already ghost the same way on `develop`, so the bug class predates this PR — but the nested case is new, and group children get added/deleted/reordered far more often than top-level items.
> 
> For the fix: `removeNode` won't work as-is because it matches `${path}.` and misses array notation (`menuItems[0].label`). Needs a sibling in `DependencyClass` that also matches `${path}[`, called from the array branch of `setComponentProperty` before `checkValueAndResolve` re-adds the edges.

Done

johnsoncherian (approved):

johnsoncherian (approved):

PR #17424, opened 2026-08-06
Feat/app page groups limit

Siddharthpl:
need to impose limits on number of pages and page groups in app builder
add page and add group button will be disabled once the limit is reached
need to show tooltip indicating that the limit has reached and the user need to upgrade for more pages and more groups

Design - https://www.figma.com/design/TCyub5quh1RNp7xfSwVzE4/Licensing?node-id=9580-63970


Siddharthpl:
Merge in another branch

PR #17450, opened 2026-08-08
fix(ability): merge orgVariableCRUD from group grants

### LINKED ISSUE #17392: orgVariableCRUD permission never merges from group grants — always false
## What

`ee/ability/service.ts` — `resourceActionsPermission`'s reduce over group permissions merges every other flag via `acc.X || group.X`, but `orgVariableCRUD` only ever reads `acc.orgVariableCRUD` (never ORs in the group's own value):

```ts
orgVariableCRUD: acc.orgVariableCRUD,
```

So no matter what any group actually grants, the resolved `orgVariableCRUD` permission is permanently `false`.

## Repro (test added, documents current behavior)

`server/test/modules/ability/unit/ee-service.spec.ts` — `it('BUG: orgVariableCRUD never merges from the group and is always false', ...)`. Mocks a single group with `orgVariableCRUD: true` and asserts the resolved permission is `false` — i.e. the test currently passes against the *buggy* behavior, flagged as a bug rather than treated as intended.

## Fix direction

Change the reduce line to `orgVariableCRUD: acc.orgVariableCRUD || group.orgVariableCRUD` to match every other flag in the same reduce.

## Found while

Reconciling the abilities/guards test suite in #17391.

eeshsaxena:
### What

`server/src/modules/ability/service.ts` — in `resourceActionsPermission`, the reduce over group permissions ORs every flag from each group (`acc.X || group.X`) except `orgVariableCRUD`, which read only `acc.orgVariableCRUD` and never merged the group's own value:

```ts
orgVariableCRUD: acc.orgVariableCRUD,
```

So the resolved `orgVariableCRUD` permission stayed at its default (`false`) regardless of what any group actually granted.

### Fix

Match the sibling flags so a group's grant is merged in:

```ts
orgVariableCRUD: acc.orgVariableCRUD || group.orgVariableCRUD,
```

Fixes #17392

### DIFF
--- a/server/src/modules/ability/service.ts
+++ b/server/src/modules/ability/service.ts
@@ -70,7 +70,7 @@ export class AbilityService extends IAbilityService {
           folderDelete: acc.folderDelete || group.folderDelete,
           orgConstantCRUD: acc.orgConstantCRUD || group.orgConstantCRUD,
           tjdbCRUD: acc.tjdbCRUD || group.tjdbCRUD,
-          orgVariableCRUD: acc.orgVariableCRUD,
+          orgVariableCRUD: acc.orgVariableCRUD || group.orgVariableCRUD,
           workflowCreate: acc.workflowCreate || group.workflowCreate,
           workflowDelete: acc.workflowDelete || group.workflowDelete,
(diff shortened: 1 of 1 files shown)


eeshsaxena:

PR #17451, opened 2026-08-08
fix(versions): findAppFromVersion returns null instead of throwing on no match

### LINKED ISSUE #17395: findAppFromVersion() throws instead of returning null — guards' 404 checks are dead code, callers 500
## What

`VersionRepository.findAppFromVersion(id, organizationId)` (`server/src/modules/versions/repository.ts`) resolves the app via `manager.findOneOrFail(AppVersion, { where: { id, app: { organizationId } }, relations: ['app'] })` — `findOneOrFail` throws `EntityNotFoundError` on no match instead of returning `null`.

Every caller assumes the null-returning contract:

```ts
const app = await this.versionRepository.findAppFromVersion(versionId, user.organizationId);
if (!app) throw new NotFoundException('App not found');
```

(`server/src/modules/versions/guards/validate-app-version.guard.ts`) That `if (!app)` branch is dead code — it can never execute, because the method throws before returning. A `versionId` that doesn't resolve for the caller's org (either it belongs to another organization, or it simply doesn't exist) surfaces as an unhandled `500` with a raw TypeORM error message, instead of the intended `404`.

## Confirmed

Live request: cross-org `GET /api/data-queries/:versionId` → `500` with body:
```json
{"statusCode":500,"message":"Could not find any entity of type \"AppVersion\" matching: {...}"}
```

## Scope — other callers of the same method, same bug shape

```
server/src/modules/data-queries/guards/validate-query-app.guard.ts
server/src/modules/versions/guards/validate-app-version.guard.ts
server/src/modules/workflows/guards/workflow-access.guard.ts
server/ee/workflows/ability/app/index.ts
```
All four call `findAppFromVersion` and check for falsy `app` before throwing their own `NotFoundException` — same dead code, same 500-instead-of-404 exposure per caller.

## Important: this is NOT a tenant-isolation gap

The isolation boundary itself holds — another organization's user genuinely cannot resolve the version or its queries. Only the status code/error shape is wrong (leaks a raw TypeORM error message via 500 instead of a clean 404).

## Repro (test added, currently skipped)

`server/test/modules/data-queries/e2e/data-queries.spec.ts` — `it.skip("does not expose another organization app version's queries — 404, not the query list", ...)`.

## Fix direction

Change `findAppFromVersion` to `findOne` (returning `App | null`) so the existing `if (!app) throw new NotFoundException(...)` guard clauses in all four callers actually run. Low-risk, single-method fix; no caller needs to change.

## Found while

Writing cross-tenant isolation e2e cases in #17391.

eeshsaxena:
### What

`VersionRepository.findAppFromVersion(id, organizationId)` (`server/src/modules/versions/repository.ts`) resolves the app via `findOneOrFail`, which throws `EntityNotFoundError` on no match instead of returning `null`. But the declared return type is `Promise<App>` and every caller assumes the null-returning contract:

```ts
const app = await this.versionRepository.findAppFromVersion(versionId, user.organizationId);
if (!app) throw new NotFoundException(...);
```

So those `if (!app)` 404 guards are dead code, and a missing/foreign version surfaces as an uncaught `EntityNotFoundError` (500) rather than a clean 404.

### Fix

Use `findOne`, guard the null before dereferencing `appVersion.app`, and widen the return type to `Promise<App | null>` so the existing caller guards do their job:

```ts
const appVersion = await m.findOne(AppVersion, { where: { id, app: { organizationId } }, relations: ['app'] });
if (!appVersion) return null;
const app = appVersion.app;
```

Fixes #17395

### DIFF
--- a/server/src/modules/versions/repository.ts
+++ b/server/src/modules/versions/repository.ts
@@ -170,7 +170,7 @@ export class VersionRepository extends Repository<AppVersion> {
     }
 
     const m = manager ?? this.manager;
-    const appVersion = await m.findOneOrFail(AppVersion, {
+    const appVersion = await m.findOne(AppVersion, {
       where: { id },
       relations: [
         'app',
@@ -255,12 +255,13 @@ export class VersionRepository extends Repository<AppVersion> {
     return m.delete(AppVersion, { id: versionId });
   }
 
-  async findAppFromVersion(id: string, organizationId: string, manager?: EntityManager): Promise<App> {
+  async findAppFromVersion(id: string, organizationId: string, manager?: EntityManager): Promise<App | null> {
     const m = manager ?? this.manager;
-    const appVersion = await m.findOneOrFail(AppVersion, {
+    const appVersion = await m.findOne(AppVersion, {
       where: { id, app: { organizationId } },
       relations: ['app'],
     });
+    if (!appVersion) return null;
     const app = appVersion.app;
     // Workflows keep metadata on apps.*; non-workflows must overlay from the
     // canonical version row resolved by git-sync state. Forward the loaded


eeshsaxena:

PR #17474, opened 2026-08-11
[Table] Add default sort option on columns

jhamadhav:
Issue: https://github.com/ToolJet/tj-ee/issues/2082

## Fix:
Adds a "Default sort column" and "Default sort order" (Ascending / Descending / Auto) setting to the Table widget, under Search, sort and filter in the properties panel, right below "Enable column sorting" and its client/server type toggle — addressing

## Behavior implemented:
Default sort column is an fx-enabled dropdown listing the table's current column names, so users can either pick a column directly or resolve one via an expression.
Default sort order offers Ascending, Descending, or Auto. Auto is a deliberate no-op — it applies no default sort at all.
The two new fields only appear when sorting is enabled and set to client-side. If sort type is switched to server-side, both fields are hidden and their values are cleared back to empty, since a default sort only makes sense for client-side sorting.
The configured default is applied automatically once the table renders, and re-applied whenever the setting itself is changed (e.g. while editing the app) — but it does not fight the user's own interaction: clicking a column header or calling the setSort action later isn't overridden by unrelated re-renders, since the effect only reacts to the default-sort setting changing, not to incidental table updates (pagination, resizing, etc.).

## Areas to Test

- Set a default column + Ascending/Descending on a fresh table — confirm it loads sorted, on both the builder canvas and preview, and after a full page refresh.
- Change the default sort setting while the app is open (not just on a brand-new table) — confirm it reapplies.
- Set direction to Auto — confirm no sort is imposed and existing/manual sorting is left alone.
- Switch sort type from client-side to server-side — confirm both new fields disappear and their stored values are cleared.
- Click a column header (or call the `setSort` action) after the default sort has applied — confirm it isn't reverted by an unrelated table re-render (e.g. changing page, resizing a column).
- Regression: existing tables created before this change should not error when their properties panel is opened.

## Screenshots:

image

### DIFF
--- a/frontend/src/AppBuilder/CodeBuilder/Elements/Select.jsx
+++ b/frontend/src/AppBuilder/CodeBuilder/Elements/Select.jsx
@@ -79,6 +79,7 @@ const selectCustomStyles = (width) => {
 };
 
 export const Select = ({ value, onChange, meta, width = '144px' }) => {
+  const resolvedWidth = meta?.fullWidth ? '100%' : width;
   return (
     <div
       className="row fx-container"
@@ -94,7 +95,7 @@ export const Select = ({ value, onChange, meta, width = '144px' }) => {
           onChange={onChange}
           width={224}
           height={32}
-          styles={selectCustomStyles(width)}
+          styles={selectCustomStyles(resolvedWidth)}
           useCustomStyles={true}
           customClassPrefix="inspector-select"
           components={{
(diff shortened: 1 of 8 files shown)


jhamadhav:
@DevangAgarwal93 Please review and suggest if improvements are required in the implemented behaviour.

jhamadhav:
Feature fix moved to common PR: https://github.com/ToolJet/ToolJet/pull/17759

Issue #5319, opened 2023-01-12
Datepicker in Tables Broken


julienmarie:
### Version Information

v2.0.1  CE / 
### Environment
Docker
### What is the expected behaviour?
The datepicker in a table should display nicely

### What is the current behaviour?

It is broken.

### How to reproduce the issue?

1 - Create a table
2 - Add a date picker field

### Screenshots or Screencast

https://user-images.githubusercontent.com/189369/212057015-d245f94a-dceb-4300-af0d-a7f0e19d3f5a.mp4

### Please provide any traces or logs that could help here.

N/A

### Any possible solutions?

Fixing the CSS :)

### Can you identify the location in the driver source code where the problem exists?

N/A

### If the bug is confirmed, would you be willing to submit a PR?

It's my first day trying tooljet. I don't have yet any knowledge of the software.

PR #17479, opened 2026-08-12
[App Theme] Expose themes in inspector and code suggestions #5319

jhamadhav:
Issue: https://github.com/ToolJet/tj-ee/issues/5319

## Summary

Previously `globals.theme` in the App Builder inspector only exposed `{ name: 'light' | 'dark' }`. This change resolves the app's full theme definition (brand/text/border/systemStatus/surface colors, plus theme metadata) against the current mode and exposes it under `globals.theme`, so app builders can bind to actual theme values (e.g. `{{globals.theme.brand.colors.primary}}`) instead of only the mode name.

## What's Implemented

- `globals.theme` now exposes, in addition to the existing `name` (mode: `'light'`/`'dark'`):
  - `themeName` — the theme record's own name (e.g. `"ToolJet"`).
  - `isBasic` / `isDefault` — the theme's own metadata flags.
  - The full resolved color/style tree: `brand`, `text`, `border`, `systemStatus`, `surface` — each `{ light, dark }` pair in the underlying theme definition is resolved down to a single value for the current mode; mode-independent leaves (`text.font`, `border.radius.*`) pass through unchanged.
- Licensed orgs with custom-theme access get their selected theme (`globalSettings.theme`) resolved; unlicensed orgs always fall back to `baseTheme` (ToolJet's built-in default), matching existing `useThemeAccess` gating.
- Re-resolves and re-exposes automatically when: app mode changes, dark/light is toggled, or the **selected theme itself** changes (e.g. via the theme dropdown) — previously only mode/dark-mode toggles triggered a refresh.
- Autocomplete (`{{ }}` hint) depth for `globals.*` was bumped from 3 to 4 so nested theme paths (e.g. `globals.theme.brand.colors.primary`) are suggested; other globals (`currentUser`, `environment`, etc.) incidentally gain one more level of hint depth too, with no functional effect (hints don't gate what's resolvable, they're autocomplete-only).
- Removed an unused `setResolvedGlobals` store selector in `CreateVersionModal.jsx` found unused during this work.

### Resolved `globals.theme` shape (example — default `baseTheme`, light mode)

```json
{
  "name": "light",
  "themeName": "ToolJet",
  "isBasic": true,
  "isDefault": true,
  "brand": {
    "colors": {
      "primary": "#4368E3",
      "secondary": "#6A727C",
      "tertiary": "#1E823B"
    }
  },
  "text": {
    "font": "IBM Plex Sans",
    "colors": {
      "primary": "#1B1F24",
      "placeholder": "#6A727C"
    }
  },
  "border": {
    "radius": {
      "default": 6,
      "small": 0,
      "large": 0
    },
    "colors": {
      "default": "#CCD1D5",
      "weak": "#E4E7EB"
    }
  },
  "systemStatus": {
    "colors": {
      "success": "#1E823B",
      "error": "#D72D39",
      "warning": "#BF4F03"
    }
  },
  "surface": {
    "colors": {
      "appBackground": "#F6F6F6",
      "surface1": "#FFFFFF",
      "surface2": "#F6F8FA",
      "surface3": "#E4E7EB"
    }
  }
}
```

Switching `mode` to `"dark"` resolves the same tree against each leaf's `dark` value instead; switching to a licensed custom theme swaps `themeName`/`isBasic`/`isDefault` and every color to that theme's own definition.

## Where It's Implemented

- `frontend/src/AppBuilder/_stores/utils.js` — added `resolveThemeForMode(definition, mode)`, a recursive resolver that walks a theme definition tree and collapses every `{ light, dark }` leaf to the value for the given mode. Also added `name`/`isDefault`/`isBasic` metadata fields to `baseTheme` so the unlicensed fallback theme has the same shape as a real theme record.
- `frontend/src/AppBuilder/_stores/slices/resolvedSlice.js` — added `updateExposedTheme(mode, moduleId)`, a single store action that picks the licensed/selected theme or `baseTheme`, resolves it via `resolveThemeForMode`, and calls `setResolvedGlobals('theme', { name, themeName, isBasic, isDefault, ...resolved }, moduleId)`. Centralizes logic previously duplicated across call sites.
- `frontend/src/AppBuilder/_hooks/useAppData.js` — the two effects that previously called `setResolvedGlobals('theme', { name: exposedTheme })` now call `updateExposedTheme(exposedTheme, moduleId)`. The mode-computation effect's dependency array now also includes `selectedTheme` and `themeAccess`, so a theme change (not just a mode toggle) re-triggers the resolve/expose step.
- `frontend/src/AppBuilder/LeftSidebar/GlobalSettings/AppModeToggle.jsx` and `frontend/src/_components/DarkModeToggle.jsx` — both toggle handlers now call `updateExposedTheme(mode)` instead of manually building `{ name: mode }`.
- `frontend/src/AppBuilder/_stores/slices/codeHinterSlice.js` — `buildGlobalHints`'s `traverseObjectToHints(globals, 'globals', 3)` depth bumped to `4` so nested theme paths surface in autocomplete.
- `frontend/src/AppBuilder/Header/CreateVersionModal.jsx` — removed an unused `setResolvedGlobals` selector (dead code, unrelated leftover found while auditing call sites).

## Areas to Review

- Depth bump in `codeHinterSlice.js` is scoped only to `buildGlobalHints`'s call; other hint builders (`queries`, `components`, `variables`, `page`) keep their own explicit `3` and are unaffected. Confirmed `rebuildGlobalHints` has no callers anywhere in the codebase today — global hints are only built once via `initSuggestions`, gated to `mode === 'edit'`, so this never runs for published/viewer apps and isn't recomputed per keystroke.
- `resolveThemeForMode` assumes theme definitions are non-circular, JSON-derived plain objects (true for both `baseTheme` and API-sourced theme records) — recursion depth is bounded by the actual (≈4-level) definition shape.
- Naming choice: the new theme-name key is `themeName`, not `name`, to avoid clashing with the pre-existing `globals.theme.name` (mode). Worth confirming this naming reads clearly to end users in the inspector.
- One `updateExposedTheme` call site (`useAppData.js`, inside the app-load flow) intentionally omits `moduleId` (defaults to `'canvas'`), matching the pre-existing behavior of the code it replaced — not changed as part of this work, but worth a second look if module-scoped theming was ever intended there.

## Areas to Test

- Toggling Auto/Light/Dark and the sun/moon icon in both editor and preview — confirm `globals.theme` in the inspector shows `name` plus all resolved category values matching the current mode, updating live without a refresh.
- Switching the selected custom theme (licensed org) via the theme dropdown — confirm `globals.theme` re-resolves to the new theme's colors and `themeName`/`isBasic`/`isDefault` update accordingly.
- Unlicensed org (no `customThemes` feature access) — confirm `globals.theme` always reflects `baseTheme` (`themeName: "ToolJet"`, `isBasic: true`, `isDefault: true`) regardless of any org-level custom theme selection.
- `{{globals.theme.brand.colors.primary}}` and similar nested paths autocomplete correctly in code editor fields.
- Existing apps/bindings using only `{{globals.theme.name}}` are unaffected (purely additive change).

## Screenshots
image

### DIFF
--- a/frontend/src/AppBuilder/_hooks/useAppData.js
+++ b/frontend/src/AppBuilder/_hooks/useAppData.js
@@ -80,6 +80,7 @@ const useAppData = (
   const setSecrets = useStore((state) => state.setSecrets);
   const setQueryMapping = useStore((state) => state.setQueryMapping);
   const setResolvedGlobals = useStore((state) => state.setResolvedGlobals);
+  const updateExposedTheme = useStore((state) => state.updateExposedTheme);
   const setResolvedPageConstants = useStore((state) => state.setResolvedPageConstants);
   const updateFeatureAccess = useStore((state) => state.updateFeatureAccess);
   const computePageSettings = useStore((state) => state.computePageSettings);
@@ -264,8 +265,8 @@ const useAppData = (
   useEffect(() => {
     const exposedTheme =
       appMode && appMode !== 'auto' ? appMode : localStorage.getItem('darkMode') === 'true' ? 'dark' : 'light';
-    setResolvedGlobals('theme', { name: exposedTheme }, moduleId);
-  }, [appMode, darkMode, moduleId]);
+    updateExposedTheme(exposedTheme, moduleId);
+  }, [appMode, darkMode, moduleId, selectedTheme, themeAccess]);
 
   useEffect(() => {
     if (!currentSession) {
@@ -891,7 +892,7 @@ const useAppData = (
         try {
           const exposedTheme =
             appMode && appMode !== 'auto' ? appMode : localStorage.getItem('darkMode') === 'true' ? 'dark' : 'light';
-          setResolvedGlobals('theme', { name: exposedTheme });
+          updateExposedTheme(exposedTheme);
         } catch (error) {
           console.log('error', error);
         }
--- a/frontend/src/AppBuilder/_hooks/useAppData.js
+++ b/frontend/src/AppBuilder/_hooks/useAppData.js
@@ -81,6 +81,7 @@ const useAppData = (
   const setQueryMapping = useStore((state) => state.setQueryMapping);
   const setResolvedGlobals = useStore((state) => state.setResolvedGlobals);
   const updateExposedTheme = useStore((state) => state.updateExposedTheme);
+  const getActiveTheme = useStore((state) => state.getActiveTheme);
   const setResolvedPageConstants = useStore((state) => state.setResolvedPageConstants);
   const updateFeatureAccess = useStore((state) => state.updateFeatureAccess);
   const computePageSettings = useStore((state) => state.computePageSettings);
@@ -263,9 +264,7 @@ const useAppData = (
   }, [moduleMode]);
 
   useEffect(() => {
-    const exposedTheme =
-      appMode && appMode !== 'auto' ? appMode : localStorage.getItem('darkMode') === 'true' ? 'dark' : 'light';
-    updateExposedTheme(exposedTheme, moduleId);
+    updateExposedTheme(undefined, moduleId);
   }, [appMode, darkMode, moduleId, selectedTheme, themeAccess]);
 
   useEffect(() => {
@@ -734,7 +733,7 @@ const useAppData = (
   useEffect(() => {
     const root = document.documentElement;
     const mode = appMode && appMode !== 'auto' ? appMode : darkMode ? 'dark' : 'light';
-    const themeObj = !themeAccess ? baseTheme?.definition : selectedTheme?.definition || {};
+    const themeObj = getActiveTheme().definition || {};
     Object.keys(themeObj).forEach((category) => {
       const categoryObj = themeObj[category];
       Object.keys(categoryObj).forEach((property) => {
@@ -890,9 +889,7 @@ const useAppData = (
         }
 
         try {
-          const exposedTheme =
-            appMode && appMode !== 'auto' ? appMode : localStorage.getItem('darkMode') === 'true' ? 'dark' : 'light';
-          updateExposedTheme(exposedTheme);
+          updateExposedTheme();
         } catch (error) {
           console.log('error', error);
         }
--- a/frontend/src/AppBuilder/_stores/slices/resolvedSlice.js
+++ b/frontend/src/AppBuilder/_stores/slices/resolvedSlice.js
@@ -1,4 +1,4 @@
-import { resolveDynamicValues, resolveThemeForMode, baseTheme } from '../utils';
+import { resolveDynamicValues, resolveThemeForMode } from '../utils';
 import { extractAndReplaceReferencesFromString } from '@/AppBuilder/_stores/ast';
 import { componentTypeDefinitionMap } from '@/AppBuilder/WidgetManager';
 import { createBatchManager } from '@/AppBuilder/_stores/batchManager';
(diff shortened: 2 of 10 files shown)


frontend/src/AppBuilder/_hooks/useAppData.js
```diff
@@ -229,8 +230,8 @@ const useAppData = (
   useEffect(() => {
     const exposedTheme =
       appMode && appMode !== 'auto' ? appMode : localStorage.getItem('darkMode') === 'true' ? 'dark' : 'light';
-    setResolvedGlobals('theme', { name: exposedTheme }, moduleId);
-  }, [appMode, darkMode, moduleId]);
+    updateExposedTheme(exposedTheme, moduleId);
+  }, [appMode, darkMode, moduleId, selectedTheme, themeAccess]);
```
devanshu052000: The `appMode !== 'auto' ? appMode : localStorage.getItem('darkMode') === 'true' ? 'dark' : 'light'` ternary is repeated at `useAppData.js:231`, `useAppData.js:834`, `DarkModeToggle.jsx:32`, and in a variant at `AppModeToggle.jsx:32-35`. The store already holds `globalSettings.appMode` in appSlice initial state). Since `updateExposedTheme` is now a store action, it should derive the mode itself by default with an optional override for the toggle handlers that fire before the mode is persisted — that removes three localStorage reads and the possibility of a call site drifting from the others.

frontend/src/AppBuilder/_stores/slices/resolvedSlice.js
```diff
@@ -131,6 +131,19 @@ export const createResolvedSlice = (set, get) => {
       );
       get().updateDependencyValues(`globals.${objKey}`, moduleId);
     },
+
+    updateExposedTheme: (mode, moduleId = 'canvas') => {
+      const state = get();
+      const hasThemeAccess = state.isLicenseValid() && state.isFeatureAccessible('customThemes');
+      const theme = (hasThemeAccess && state.globalSettings?.theme) || baseTheme;
+      const resolvedTheme = resolveThemeForMode(theme.definition, mode);
```
devanshu052000: the theme-picking logic already exists twice:

- `AppBuilder/_hooks/useThemeAccess.js` — reads license.featureAccess directly
- `useAppData.js:702` — !themeAccess ? baseTheme?.definition : selectedTheme?.definition || {}
- new `resolvedSlice.js:137-138` — isLicenseValid() && isFeatureAccessible('customThemes')

They already disagree on one edge: for a theme record with a missing definition, line 702 yields {} (no CSS vars), while the new code keeps the record and exposes a correct themeName with an empty color tree instead of falling back to `baseTheme`. 

Suggest 
```
const theme = (hasThemeAccess && state.globalSettings?.theme?.definition) ? state.globalSettings.theme : baseTheme;
``` 
and, better, one store getter (getActiveTheme()) consumed by all three — that also guarantees the CSS variables and the exposed values can never describe different themes.

frontend/src/AppBuilder/_stores/utils.js
```diff
@@ -764,6 +767,23 @@ export const baseTheme = {
...
+    }
+    return Object.keys(node).reduce((resolved, key) => {
+      resolved[key] = resolveNode(node[key]);
+      return resolved;
+    }, {});
+  };
+
+  return resolveNode(definition);
+};
+
```
devanshu052000, line 785: **`globals.theme` is exposed with snake_case keys in every viewer / released / public-app path**

`resolveThemeForMode` passes the definition's keys through verbatim, but that object doesn't have the same casing in the editor and the viewer:

- `useAppData.js:270` runs `appData = convertAllKeysToSnakeCase(appData)` for released apps, public links, preview-for-version and the module viewer (every non-editor path).
- `appDataCaseConversion.ts` only skips recursion for `pages` / `events` / `dataQueries` / `data_queries`, so `global_settings.theme.definition` is deep-converted.
- `useAppData.js:417` then does a shallow `mapKeys(..., camelCase)`, which only restores top-level global-settings keys — the `theme` subtree stays snake_case.

Result for a licensed org with a custom theme, in the released app:

| Editor (works) | Released app / public link |
| --- | --- |
| `globals.theme.systemStatus.colors.error` | `undefined` (key is `system_status`) |
| `globals.theme.surface.colors.appBackground` | `undefined` (key is `app_background`) |
| `globals.theme.isBasic` / `isDefault` | `undefined` (keys are `is_basic` / `is_default`) |

Single-word keys (`brand`, `text`, `border`, `surface1..3`) survive, so it fails silently on exactly these paths. Unlicensed orgs fall back to the JS `baseTheme` constant and stay camelCase — so the same released app exposes a different shape depending on license state.

The existing code already defends against this: the CSS-variable loop at `useAppData.js:702-720` runs `camelCase(type)` / `camelCase(category)` on every key.

Fix — normalize keys in the resolver (`_` is already imported at line 8 of this file):

```js
return Object.keys(node).reduce((resolved, key) => {
  resolved[_.camelCase(key)] = resolveNode(node[key]);
  return resolved;
}, {});
```

and, since the two metadata flags are read off the record outside this resolver, in `resolvedSlice.js`:

```js
isBasic: theme.isBasic ?? theme.is_basic,
isDefault: theme.isDefault ?? theme.is_default,
```

devanshu052000 (changes requested):

frontend/src/AppBuilder/_hooks/useAppData.js
jhamadhav: fixed.

frontend/src/AppBuilder/_stores/slices/resolvedSlice.js
jhamadhav: done.

frontend/src/AppBuilder/_stores/utils.js
jhamadhav: done.

devanshu052000 (dismissed):

kavinvenkatachalam (dismissed):

PR #17501, opened 2026-08-13
fix(app): clear the metadata-refresh interval on unmount

### LINKED ISSUE #17498: frontend: App.jsx setInterval for metadata refresh never cleared - accumulates on remount, continues after logout
﻿## Bug Description

App.jsx starts a setInterval for metadata refresh but never stores the return value and never clears it. If the component is unmounted and remounted, a new interval is added each time without cancelling the previous one. Stale intervals continue making authenticated HTTP requests after the user logs out.

## Affected file

rontend/src/App/App.jsx, line 133:

`js
setInterval(this.fetchMetadata, 1000 * 60 * 60 * 1);
// return value discarded - interval can never be cleared
`

For contrast, ViewerApp.jsx line 112 correctly stores and clears the interval:
`js
// ViewerApp.jsx - correct pattern
this._metadataInterval = setInterval(this.fetchMetadata, interval);
// componentWillUnmount:
clearInterval(this._metadataInterval);
`

## Failure scenario

1. User logs in. App mounts. Interval 1 starts, making metadata requests every hour.
2. HMR reload in development (or future routing change) unmounts and remounts App. Interval 2 starts.
3. Interval 1 is still running - now two concurrent hourly requests fire.
4. Over repeated HMR cycles, N intervals accumulate.
5. User logs out. Stale intervals continue calling etchMetadata, triggering 401 responses and unnecessary network activity for the session lifetime.

## Fix

`js
componentDidMount() {
  this._metadataInterval = setInterval(this.fetchMetadata, 1000 * 60 * 60 * 1);
  // ... rest of componentDidMount
}

componentWillUnmount() {
  clearInterval(this._metadataInterval);
}
`

## Environment

ToolJet main branch (2026-08-13), React (class component).

eeshsaxena:
Fixes #17498.

`App.componentDidMount` starts an hourly metadata refresh but throws the timer id away:

```js
setInterval(this.fetchMetadata, 1000 * 60 * 60 * 1);
```

Nothing ever clears it, and `App` has no `componentWillUnmount`. So when `App` unmounts and remounts a fresh interval is added each time on top of the old ones, and the stale intervals keep calling `fetchMetadata` (an authenticated request) even after the user logs out.

The viewer side already does this correctly by storing the id and clearing it on unmount. This does the same: keep the id on the instance and clear it in `componentWillUnmount`.

The nearby telemetry interval a few lines down is unaffected since it already clears itself from inside its own callback.

### DIFF
--- a/frontend/src/App/App.jsx
+++ b/frontend/src/App/App.jsx
@@ -130,7 +130,7 @@ class AppComponent extends React.Component {
         showBanner: false, // show banner when required for ee or cloud
       });
     }
-    setInterval(this.fetchMetadata, 1000 * 60 * 60 * 1);
+    this.metadataRefreshInterval = setInterval(this.fetchMetadata, 1000 * 60 * 60 * 1);
     this.updateMargin(); // Set initial margin
     let counter = 0;
     let interval;
@@ -146,6 +146,14 @@ class AppComponent extends React.Component {
       }
     }, 1000);
   }
+
+  componentWillUnmount() {
+    // Without this the hourly metadata refresh keeps running after the
+    // component unmounts (e.g. on logout) and on remount a second interval
+    // is added, so authenticated fetchMetadata calls accumulate.
+    clearInterval(this.metadataRefreshInterval);
+  }
+
   // check if its getting routed from editor
   checkPreviousRoute = (route) => {
     if (route.includes('/apps')) {
--- a/frontend/src/App/App.jsx
+++ b/frontend/src/App/App.jsx
@@ -133,16 +133,15 @@ class AppComponent extends React.Component {
     this.metadataRefreshInterval = setInterval(this.fetchMetadata, 1000 * 60 * 60 * 1);
     this.updateMargin(); // Set initial margin
     let counter = 0;
-    let interval;
 
-    interval = setInterval(async () => {
+    this.userCheckInterval = setInterval(async () => {
       ++counter;
       const current_user = authenticationService.currentSessionValue?.current_user;
       if (current_user?.id) {
         this.initTelemetryAndSupport(current_user); //Call when currentuser is available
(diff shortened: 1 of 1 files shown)


harsh4vardhan:
Correct fix. Storing the interval ID on the instance and clearing it in `componentWillUnmount` is exactly the right pattern, and it matches `ViewerApp.jsx`. Two small things to consider before merge:

1. Does `App.jsx` have any other `setInterval`/`setTimeout` calls that are also not cleaned up? A quick search didn't surface any from the diff context, but worth a confirm.
2. A test that mounts, unmounts, then remounts `App` and asserts `fetchMetadata` is called exactly once per mount (not accumulating) would prevent this from regressing. The fix is simple enough that the test would be equally simple.

eeshsaxena:
Thanks for the review.

1. The other interval a few lines down (the per-second `authenticationService` poll) does self-clear once a user is available or after ~10 iterations, so it's bounded rather than an open-ended leak. But on a quick unmount/remount the old instance's poll can still run for a few seconds against a stale `this`, so I stored it on the instance and clear it in `componentWillUnmount` too.
2. A test that mounts `App` pulls in `authorizeWorkspace`, routing and `localStorage`, so a faithful mount/unmount/remount test needs a fair bit of mocking. I left it out here rather than ship something flaky, but happy to add a focused one if you'd prefer it in this PR.

eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`fix(app): clear the metadata-refresh interval on unmount`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining ToolJet!

eeshsaxena:

PR #17551, opened 2026-08-17
fix(grpcv2): reject queries for services removed from the data source (#17537)

### LINKED ISSUE #17537: gRPC: Able to run a query for a service that was removed from the configuration
The query should not be executable once the associated gRPC service has been removed from the configuration. An appropriate error should be displayed indicating that the service is no longer available/configured.

The query can still be executed even though the associated gRPC service has been removed from the configuration.

2. Navigate to the data source page 
3. Configure a gRPC service.
4. Create an app and create a query for the service.
5. Remove the service from the configuration.
6. Try to run the existing query.
7. Observe that the query still executes successfully.

### Screenshots or Screencast
https://jam.dev/c/c05b72fb-b4a8-4055-845d-60a4898db5a9

yzxcj797:
Closes #17537.

## Root cause

Every gRPC client builder validates the requested service against the **live source**: `buildReflectionClient` checks the server's reflection list, `buildProtoFileClient` the proto file, `buildFilesystemClient` the filesystem scan. None of them ever consults the datasource **configuration**. So after a service is removed from the data source's Services selection, a saved query targeting it still resolves — the server still exposes it, or the proto files still contain it — and executes successfully, exactly as reported.

## Fix

`run()` now calls a `validateServiceConfigured()` gate before any client work: when `selected_services` is a non-empty array on the source options, the query's `service` must be in it, or the run fails with an error naming the configured alternatives:

> `Service grep.Old is no longer configured on this data source. Update the query to use one of the configured services: grep.New`

The error surfaces through the existing `GrpcOperationError -> QueryError('Query could not be completed', ...)` path, so the UI shows it like any other query failure. An **unset or empty** selection (legacy or unrestricted datasource) stays permissive — the gate only enforces a selection the configuration actually made.

## Tests

First real tests in `plugins/packages/grpcv2/__tests__` (the file was an `it.todo` stub):

- a removed service is rejected with the no-longer-configured message — before any client/network work;
- a still-selected service passes the gate (the run then fails at client creation against a dead URL, proving it got through);
- an unset selection stays permissive (same downstream failure).

### DIFF
--- a/plugins/packages/grpcv2/__tests__/index.js
+++ b/plugins/packages/grpcv2/__tests__/index.js
@@ -1,7 +1,38 @@
 'use strict';
 
-const grpcv2 = require('../lib');
+const Grpcv2QueryService = require('../lib').default;
 
 describe('grpcv2', () => {
-    it.todo('needs tests');
+  describe('#17537: a query may only target services still configured on the data source', () => {
+    const service = new Grpcv2QueryService();
+
+    const sourceOptions = (selected_services) => ({
+      proto_files: 'server_reflection',
+      url: 'localhost:50051',
+      ...(selected_services !== undefined ? { selected_services } : {}),
+    });
+    const queryOptions = { service: 'grep.Old', method: 'Search', raw_message: '{}' };
+
+    it('rejects a service removed from the configuration before any client work', async () => {
+      await expect(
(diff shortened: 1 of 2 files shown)


RudraPratik30:

PR #17552, opened 2026-08-17
Feature: Serply plugin

### LINKED ISSUE #17582: Add a Serply plugin (SERP / web search data source)

ToolJet ships data-source plugins for many external services, but none for a SERP or web-search API. A builder who wants live web search results (organic results, news, scholar, and similar) inside a ToolJet app currently has to fall back on a generic REST call and parse the raw response by hand.

Add a Serply plugin as an optional marketplace data source. Serply is a SERP API (https://serply.io, docs at https://serply.io/docs). The plugin exposes a small set of query operations that return normalized JSON, so a builder can drop search results straight into a table or transform without writing HTTP glue.

It follows the existing plugin structure and is fully optional: it does nothing unless a user adds the plugin and supplies their own Serply API key. No existing behavior changes for anyone who does not use it.

Using the REST API plugin against a raw Serply endpoint works, but it leaves every builder to build the request and reshape the response themselves. A first-class plugin gives typed operations and a consistent result shape, matching how other data sources are exposed in ToolJet.

Yes. I have a working implementation already open as #17552 and am happy to align it with this issue and your review.

Disclosure: I work with Serply.

googio:
Closes #17582

## What

Adds a Serply marketplace plugin under `marketplace/plugins/serply` with four
operations: Web Search and News Search (Google results as JSON), plus Reddit
Post and Reddit Comments for pulling a post and its comment thread by ID.
The datasource takes an encrypted API key and an optional proxy location
(US, EU, or GB) for routing searches through a region. The plugin is
registered in `server/src/assets/marketplace/plugins.json`.

## Why

The marketplace has no SERP or web search datasource today. Apps that need
live search results or community content (lead research, SEO dashboards,
market monitoring, brand tracking) currently have to hand-build a REST API
source for each provider. This gives them a one-click datasource instead.

## Notes

- Modeled on the prometheus plugin: a `QueryService` class, `got` for HTTP,
  and per-operation key prefixes in `operations.json`.
- No new dependencies. `got` resolves to 11.8.6, already hoisted in
  `marketplace/package-lock.json`, so the lockfile is untouched.
- The API key is `encrypted: true` in source options, and test connection
  runs a single search against the API.
- `__tests__/index.js` carries the same `it.todo` placeholder the other
  marketplace plugins use.
- Docs page added under `docs/docs/marketplace/plugins/serply.md` with a
  sidebar entry, following the format of the other plugin pages. It has no
  screenshots yet; happy to add them, or to adjust scope or naming.

## Testing

```
cd marketplace && npm install
npm run build                        # all 39 plugin workspaces build, exit 0
npx eslint plugins/serply/lib/*.ts   # clean
```

Loaded the built bundle and ran every operation against the live API with a
real key:

```
web search: ok | 7 results | first: https://tooljet.com/
news search: ok | entries: 100
reddit post: ok | title: Showcase Thread | comments: 77
reddit comments: ok | payload keys: cached,data
testConnection: {"status":"ok"}
testConnection bad key: {"status":"failed","message":"Response code 401 (Unauthorized)"}
missing id: QueryError raised
```

Disclosure: I work with Serply. Happy to adjust scope, naming, or drop this
entirely if it isn't a direction you want for the project.

### DIFF
--- a/docs/docs/marketplace/plugins/serply.md
+++ b/docs/docs/marketplace/plugins/serply.md
@@ -1,0 +1,125 @@
+---
+id: marketplace-plugin-serply
+title: Serply
+---
+
+Integrating Serply with ToolJet lets applications fetch live Google web and news search results and pull Reddit posts and comment threads. This is useful for building lead research tools, SEO dashboards, market monitoring, and brand tracking apps.
+
+## Connection
+
+To connect with Serply, you will need the **API Key**, which can be generated from **[Serply](https://serply.io)**.
+
+You can optionally set a **Proxy location** (for example US, EU, or GB) to route searches through a specific region. See the **[Serply API documentation](https://serply.io/docs)** for the full list of supported regions and response fields.
+
+## Supported Operations
+
+### Web Search
+
+Returns Google web search results for a query as JSON.
+
+**Required Parameters**
+
+- **Query**: The search query string.
+
+**Optional Parameters**
+
+- **Number of results**: How many results to return, between 10 and 100. Defaults to 10.
+
+<details id="tj-dropdown">
+<summary>**Response Example**</summary>
+
+```json
+{
+  "results": [
+    {
+      "title": "ToolJet",
+      "link": "https://tooljet.com/",
+      "description": "ToolJet helps you build enterprise apps, AI agents and workflows in minutes, not months.",
+      "position": 1
+    }
+  ]
+}
+```
+
+</details>
+
+### News Search
+
+Returns Google news results for a query as JSON.
+
+**Required Parameters**
+
+- **Query**: The news search query string.
+
+**Optional Parameters**
+
+- **Number of results**: How many results to return, between 10 and 100. Defaults to 10.
+
+<details id="tj-dropdown">
+<summary>**Response Example**</summary>
+
+```json
+{
+  "entries": [
+    {
+      "title": "Ferrari's first ever electric car sold for record $40m",
+      "link": "https://news.google.com/rss/articles/...",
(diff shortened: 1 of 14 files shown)


RudraPratik30:
Hi @googio, thanks for your interest in contributing to ToolJet. Could you please link the concerned issue here?

googio:
Thanks @RudraPratik30. I have opened a feature request for this: #17582. I also left a comment there to express interest and request assignment, per the contributing guide.

Happy to align this PR with your process, including renaming the branch to the `feature/<issue-id>-...` convention and adding `Closes #17582` to the commit if that helps. Just let me know what you would prefer.

Disclosure: I work with Serply, and this plugin is optional (it does nothing unless a user adds it with their own API key).

googio:
@RudraPratik30 following up on the issue link you asked for. I opened #17582 the same day and commented there asking to be assigned; the issue is still open and still unassigned a month later.

Where this stands: the branch is mergeable against `develop`, the CLA check is green, and the only other check is the Socket dependency report, which is neutral because the scan exceeded its 10,000-dependency limit rather than because of anything in the diff. The plugin adds no new runtime dependency.

Two questions, either answer is fine:

1. Is a search plugin something ToolJet wants at all? If the answer is no, I will close this and #17582 so neither sits open.
2. If yes, is this PR the implementation you want? @PremveerSingh195 asked to take #17582 on 08-20. I have no claim on the issue, so if the team would rather assign it there, say the word and I will close this to keep the issue clean.

If you do want this one, I am happy to rebase onto current `develop` and work through review whenever someone has time.

Disclosure, as stated in the PR description: I work with Serply, so treat this as a vendor contribution and hold it to that standard.

PR #17557, opened 2026-08-18
fix: unblock npm install on rspack 2 and pin css-loader 7 module options

johnsoncherian:
## Problem

`npm install` fails with `ERESOLVE` on `feat/rspack`. Four webpack loaders/plugins are pinned to versions whose `peerOptional @rspack/core` range is `0.x || 1.x`, which conflicts with the `@rspack/core@2.x` this branch introduced. npm surfaced them one at a time, so each fix revealed the next.

This is not just a local annoyance: CI (`ci.yml` runs `npm ci`; the cloud-frontend and deploy workflows run plain `npm install`) passes no `--legacy-peer-deps`, so the branch would fail there too. I deliberately avoided an `.npmrc` escape hatch for that reason — the versions are fixed instead.

## Changes

Each package is bumped to the first release that accepts `@rspack/core` 2.x:

| package | was | now |
|---|---|---|
| `css-loader` | `^6.7.3` | `^7.1.4` (no 6.x supports rspack 2) |
| `html-webpack-plugin` | `^5.5.0` | `^5.6.8` |
| `postcss-loader` | `^8.1.0` | `^8.2.1` |
| `sass-loader` | `^16.0.5` | `^16.0.8` |

The lockfile pinned older patches, so the ranges had to be raised to force the upgrade.

`css-loader` 7 is a major bump that changes two CSS Modules defaults, so both `rspack.config.js` and `webpack.config.js` now pin them back to v6 behaviour:

```js
modules: { auto: true, namedExport: false, exportLocalsConvention: 'as-is' },
```

- **`namedExport`** now defaults to `true`, which drops the default export the app relies on — `src/_ui/SortableTree/TreeItemWrapper.jsx` and `src/TooljetDatabase/Menu/CellEditMenu/index.jsx` both do `import styles from './x.module.scss'`.
- **`exportLocalsConvention`** then falls back to `'camel-case-only'` (v6 used as-is), which lowercases PascalCase keys — `styles.Wrapper`, `styles.TreeItem` and `styles.ChildIndentation` in SortableTree would silently become `undefined`.
- **`auto: true`** has to be explicit: when `modules` is passed as an object, css-loader leaves `auto` undefined, and the module check only short-circuits for a boolean/regexp/function. Undefined means *every* stylesheet is treated as a CSS module and all global class names get hashed — which strips styling from the entire app.

## Verification

- `npm install` completes cleanly.
- `npm run build` (rspack, production) compiles with no errors — only the pre-existing browserslist-age and entrypoint-size warnings.
- Global classes emit verbatim in the built CSS (`.canvas-container`, `.widget-button`, `.query-manager`, `.theme-dark`).
- CSS Modules still hash, and every class the JSX uses resolves — `styles.Wrapper` compiles to the hashed `.SOwePrsfbHEGmuzD` in the bundle. The only unreferenced hashes are `.Text`, `.Collapse` and `.collapsed`, which `TreeItem.module.css` defines but no component uses (unchanged from before).

Note: `webpack.config.js` (used by `build:cloud`) got the same css-loader options for consistency, but that path was not build-tested.


kavinvenkatachalam:
Cherry‑picked the commit into the PR: https://github.com/ToolJet/ToolJet/pull/17563. Closing this one.

PR #17561, opened 2026-08-18
Feature: Stream server logs to OTEL, readable log levels in prod

adishM98:
## 📝 What this does
Streams server-side logs to the same OTLP pipeline already used for frontend error logs, and fixes production JSON logs encoding level as a bare number instead of a readable word. Stacked on #17541 (`feat/frontend-metrics-lts`).

Went through two review passes after the first commit. Biggest outcome: the server built **two separate pino instances** (nestjs-pino for HTTP/bootstrap logs, `TransactionLogger` for `appLogger.*`), and only one was ever wired to OTLP. Both now share one construction site instead of duplicating the wiring.

## 🏗️ Architecture

```mermaid
flowchart TD
    subgraph origins["three log origins, one instance now — previously two"]
        HTTP["HTTP request/response\nautoLogging"]
        BOOT["main.ts bootstrap"]
        APP["appLogger.* — 15 files\nTransactionLogger"]
    end

    HTTP -->|"loader.ts: pinoHttp.logger"| BASE
    BOOT --> BASE
    APP -->|"redact() first"| BASE

    BASE["base-logger.ts\nbuildBaseLogger() — memoized singleton"]

    BASE -->|"stdout level, unchanged"| STDOUT["stdout / docker logs"]
    BASE -->|"OTEL_LOGS_LEVEL, default info\nonly if ENABLE_OTEL && OTEL_EXPORTER_OTLP_LOGS"| OLS["OtelLogStream\nwhole write() guarded"]

    OLS -->|"getServerLogger() — lazy"| EMIT["logger.emit()\ntrace context filled natively"]
    EMIT --> BATCH["BatchLogRecordProcessor"]
    BATCH -->|"OTLP"| COLLECTOR["otel-collector"] --> LOKI["Loki"]

    SIGTERM["SIGTERM — two independent handlers"] -->|"await shutdownOtelLogs()"| BATCH

    style BASE stroke:#166f63,stroke-width:2px
    style OLS stroke:#166f63,stroke-width:2px
```

Gated entirely on `ENABLE_OTEL` **and** `OTEL_EXPORTER_OTLP_LOGS` — with either unset, `buildBaseLogger()` returns a single-destination instance and no multistream is built, so `docker logs` volume is unchanged. The one thing that *does* change for everyone is the level encoding — see below.

## ⚠️ Breaking change — log level encoding

`formatters.level` applies unconditionally, not only when OTEL is on. Production JSON logs change for
**every** deployment, including those with `ENABLE_OTEL` unset:

```diff
- {"level":30,"time":1770...,"msg":"app opened"}
+ {"level":"info","time":1770...,"msg":"app opened"}
```

Deliberate — a self-hosted admin tailing `docker logs` has no way to know `30` means info. But any
downstream pipeline rule matching numeric levels (e.g. Datadog/ELK `level >= 50`) silently stops
matching. Needs a CHANGELOG line:

> `ORM_LOGGING` aside, server logs now encode level as a word (`"info"`) rather than a number (`30`).
> Update any log pipeline rule that matches on numeric levels.

## Design considerations

- **One pino instance, not two.** Half the server's logs (HTTP + bootstrap) never reached OTLP before, and the level-mapping logic had already drifted between the two (`'trace'` vs `'debug'` for the same `NODE_ENV=development`). One shared `base-logger.ts` fixes both.
- **New `OTEL_LOGS_LEVEL` var, kept separate from `OTEL_LOG_LEVEL`.** The latter already gates the OTel SDK's own internal diagnostics — unrelated concern, would've collided in meaning.
- **Redaction happens in `enrichLogData()`, before `JSON.stringify`.** By the time a param reaches `OtelLogStream`, it's already flattened to a string — no object shape left for any redactor to match. Protects stdout too, not just OTLP.
- **`OtelLogStream.write()` is fully guarded, not just the `JSON.parse`.** `pino.multistream` doesn't isolate destinations — an unguarded throw here would surface as a request failure at the original `appLogger.*` call site.
- **Both shutdown paths (`tracing.ts`'s SIGTERM handler and `main.ts`'s `gracefulShutdown`) now await `shutdownOtelLogs()`**, bounded by a timeout, so the last few seconds of logs aren't lost on restart.
- **Audited every logger construction site.** `CustomDomainCacheService` and `CustomDomainStatusScheduler` were building their own `Logger` instead of using the shared one — `cache.service.ts` mattered functionally (its `onModuleInit()` errors could fire before the global override was set), so both now inject `TransactionLogger`. Left `tooljet-database.validator.ts` alone — it's a `class-validator` constraint, not Nest-DI-managed, no timing bug there, not worth wiring DI into `class-validator` for one log line.

## 🔀 Changes
- Added `base-logger.ts`, the single pino construction site shared by HTTP/bootstrap logs and `appLogger.*`
- Added `LOG_LEVEL` (app verbosity) and `OTEL_LOGS_LEVEL` (Loki ingest volume) env vars. `OTEL_LOGS_LEVEL` is fully independent of the similarly-named `OTEL_LOG_LEVEL` (SDK diagnostics). `LOG_LEVEL` **takes priority over** `ORM_LOGGING` but still falls back to it when unset, so existing deployments keep their current verbosity — decoupling the two entirely is a follow-up
- Fixed numeric log levels in prod JSON, OTLP turning on without an endpoint configured, a throw in the OTLP path being able to fail a request, and the log queue not flushing on shutdown
- Added redaction for sensitive fields (passwords, tokens, secrets) before they leave the process, with a circular-reference guard so a bad object can't crash the caller
- Fixed two more logger construction sites (`CustomDomainCacheService`, `CustomDomainStatusScheduler`) to use the shared logger instead of their own

## 🚀 Rollout notes

Two things to set deliberately when this reaches an environment with a collector:

- **`SLOW_QUERY_LOGGING_THRESHOLD=30`.** It defaults to `1`, so TypeORM reports nearly every query,
  and the sub-30ms branch of `logQuerySlow` emits at `info` — which the OTLP stream accepts. That is
  a Loki record per query, dozens per request. `BatchLogRecordProcessor` drops on queue overflow
  (2048), and the drop is indiscriminate, so the flood would evict error records too.
- **`LOG_LEVEL`.** Left unset it resolves to `warn`, while `OTEL_LOGS_LEVEL` defaults to `info` —
  so HTTP request logs reach Loki but are *invisible* in `docker logs`. Setting `LOG_LEVEL=info`
  makes both streams show the same thing. Either is defensible; it should be a decision, not a
  default.

Kill switch is `OTEL_EXPORTER_OTLP_LOGS` — unset it and log shipping stops while traces and metrics
keep working. Do not use `ENABLE_OTEL=false` for that; it takes all telemetry down with it.

## 🧪 How to test
- [x] `npm test` — 24/24 passing across 3 spec files
- [x] Typechecked against installed `pino@10.3.1` — zero errors
- [x] Verified live: throwing `emit()` no longer escapes `OtelLogStream.write()`
- [x] Verified live: both log origins land on the same pino instance and reach OTLP
- [x] End-to-end against local Loki + otel-collector, `ENABLE_OTEL=true` — server logs confirmed in Loki


server/src/modules/app/loader.ts
```diff
@@ -76,28 +77,23 @@ export class AppModuleLoader {
...
-            level,
+            // Shared with TransactionLogger (service.ts) — buildBaseLogger() is memoized,
+            // so this resolves to the exact same pino instance either way. Previously built
+            // twice here, with its own independent copy of the level-mapping logic that had
+            // already drifted from service.ts's ('debug' vs 'trace' for the same
+            // NODE_ENV=development case) — one construction site now, so HTTP request logs
+            // and bootstrap lines get the OTLP stream identically to appLogger.* calls,
+            // instead of never reaching it at all.
+            logger: buildBaseLogger(),
             autoLogging,
```
akshaysasidrn: Why would we need 2 logger and autoLogging? Check their difference.

server/src/helpers/log-redaction.helper.ts
```diff
@@ -0,0 +1,52 @@
...
+ * to appLogger.log(msg, someObject) gets JSON.stringify'd into the message string before
+ * pino (or the OTel log stream downstream of it) ever sees it. This has to run before
+ * that flattening, not after, or there's no object structure left to redact.
+ */
+export function redactSensitiveKeys(value: unknown, seen = new WeakSet<object>()): unknown {
+  if (Array.isArray(value)) {
+    return value.map((item) => redactSensitiveKeys(item, seen));
+  }
+  if (value && typeof value === 'object') {
+    if (seen.has(value)) return '[CIRCULAR]';
```
akshaysasidrn: Is adding circular for redaction a standard practice?

server/src/main.ts
```diff
@@ -155,14 +163,19 @@ async function bootstrap() {
...
-      process.exit(0);
     } catch (error) {
       logger.error('❌ Error during application shutdown:', error);
-      process.exit(1);
+      exitCode = 1;
     }
+    // Flush the OTel log queue regardless of how app.close() went — tracing.ts's own
+    // SIGTERM handler shuts down traces/metrics independently of this one, and neither
+    // used to wait for the log processor's own up-to-5s batch to drain before exiting.
+    await shutdownOtelLogs();
```
akshaysasidrn: this should be in try block?

server/test/modules/logging/unit/base-logger.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  it('is a singleton — the same instance every call, even across different consumers', () => {
+    process.env.NODE_ENV = 'test';
+    const first = buildBaseLogger();
+    const second = buildBaseLogger();
+
+    expect(first).toBe(second);
+  });
+
+  describe('I5 — one instance means HTTP/bootstrap logs reach OTLP too, not just TransactionLogger', () => {
```
akshaysasidrn: Dont put agent issue context like I5* on code.
tests should be readable without any context.

server/test/modules/logging/unit/base-logger.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      const emitted: unknown[] = [];
+      const originalGetServerLogger = otelLogs.getServerLogger;
+      otelLogs.getServerLogger = () => ({ emit: (record: unknown) => emitted.push(record) });
+
+      try {
+        const logger = build({
+          NODE_ENV: 'production',
+          ORM_LOGGING: 'error', // stdout stays quiet — split verbosity must survive unification
+          ENABLE_OTEL: 'true',
+          OTEL_EXPORTER_OTLP_LOGS: 'http://localhost:4318/v1/logs',
```
akshaysasidrn: Will this break if this endpoint is not live?

PR #17573, opened 2026-08-18
fix(ui): keep NumberInput clear button hidden when the value is empty

### LINKED ISSUE #17550: bug: number input clear button is always visible despite FX visibility condition

The clear button should respect the configured FX condition on the number input component. When the primary number field is connected to a value field, the clear button should only be visible when that value field is not empty.

The clear button is always visible on the number input component, even when the FX condition should evaluate to false and hide it.

1. Add a Number Input widget in the ToolJet app builder.
2. Configure the primary number field to be connected to a value field.
3. Set the clear button visibility logic so it should only appear when the value field is not empty.
4. Observe the rendered number input component.
5. Notice that the clear button remains visible regardless of whether the value field is empty.

Image

yzxcj797:
## Summary

Fixes #17550.

The clear button on the NumberInput widget stayed visible even when the input looked empty and the FX on the toggle evaluated to hiding it.

## Root cause

Not the FX evaluation — the value normalization feeding `BaseInput`:

```js
// NumberInput.jsx (before)
value: Number(parseFloat(props.properties.value).toFixed(props.properties.decimalPlaces))
```

When the bound value field is empty (`''` / `null` / `undefined` — e.g. the FX-bound source field has no content yet), `parseFloat` yields `NaN`, so the component state becomes `NaN`. In `BaseInput`:

```js
const hasValue = value !== '' && value !== null && value !== undefined;      // NaN passes all three → true
const shouldShowClearBtn = showClearBtn && hasValue && !disable && !loading; // → visible
```

`<input type="number">` refuses to render `NaN` and shows an empty field, so the result is exactly what the issue reports: **an input that looks empty with a visible clear button**, regardless of the configured FX.

The same coercion ran again in `handleBlur`, so clearing the input and blurring brought `NaN` right back.

## Fix

A shared `toDisplayValue(rawValue, decimalPlaces)` helper maps non-finite parses to `''` — the empty sentinel `TextInput` and `handleClear` already use — and is applied at init and in `handleBlur`. Numeric values keep the existing `decimalPlaces` rounding.

## Tests

`frontend/src/AppBuilder/Widgets/__tests__/NumberInput.test.jsx` (jest + @testing-library/react, `BaseInput` mocked to mirror its real `shouldShowClearBtn` gating):

- empty / `undefined` / `null` value → `''` reaches `BaseInput`, clear button hidden;
- numeric value → clear button visible;
- `decimalPlaces` rounding preserved;
- blur on an emptied input keeps `''` (not `NaN`), clear button hidden.

Couldn't run the frontend suite locally (Windows checkout without `node_modules`); relying on CI for the jest run.

### DIFF
--- a/frontend/src/AppBuilder/Widgets/NumberInput.jsx
+++ b/frontend/src/AppBuilder/Widgets/NumberInput.jsx
@@ -4,6 +4,15 @@ import { useInput } from './BaseComponents/hooks/useInput';
 import { cn } from '@/lib/utils';
 import SolidIcon from '@/_ui/Icon/SolidIcons';
 
+// parseFloat('') / parseFloat(null) is NaN, and a NaN value must never reach
+// BaseInput: <input type="number"> renders NaN as an empty field while
+// `hasValue` (NaN !== '' && NaN !== null && NaN !== undefined) stays true,
+// so the clear button shows on an input that looks empty (#17550).
+const toDisplayValue = (rawValue, decimalPlaces) => {
+  const parsed = parseFloat(rawValue);
+  return Number.isFinite(parsed) ? Number(parsed.toFixed(decimalPlaces)) : '';
+};
+
 export const NumberInput = (props) => {
   const inputRef = useRef(null);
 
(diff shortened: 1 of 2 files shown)


RudraPratik30:

PR #17586, opened 2026-08-18
fix(widgets): number input clear button no longer shows on an empty field (#17550)

### LINKED ISSUE #17550: bug: number input clear button is always visible despite FX visibility condition

yzxcj797:
Closes #17550.

## Root cause

`NumberInput` parsed its value through `Number(parseFloat(v).toFixed(dp))`, which yields **NaN** for an empty or unparsable value. `BaseInput`'s emptiness gate compares against `''`/`null`/`undefined` — **NaN equals none of them** — so `hasValue` was `true` on an empty field, and `shouldShowClearBtn` rendered the button no matter what the visibility condition evaluated to. Verified directly: `Number(parseFloat(undefined).toFixed(2))` → `NaN`, and the old `hasValue` expression returns `true` for it.

## Fix — two complementary layers

1. **`NumberInput`** coerces an unparsable value to `null` — the same value `handleChange` already sets when the field is cleared — so the component's value state stays consistent end to end.
2. **`BaseInput`'s `hasValue`** also treats `NaN` as empty (`!Number.isNaN(value)`), protecting every numeric caller of the shared component from the same shape.

With an empty field the clear button is now hidden; with a value (FX-driven or typed) it shows — exactly the reported expectation. The FX condition on `showClearBtn` itself was never broken: the shared gate below it was.

### DIFF
--- a/frontend/src/AppBuilder/Widgets/BaseComponents/BaseInput.jsx
+++ b/frontend/src/AppBuilder/Widgets/BaseComponents/BaseInput.jsx
@@ -84,7 +84,8 @@ export const BaseInput = ({
   const _width = getLabelWidthOfInput(widthType, width);
   const defaultAlignment = alignment === 'side' || alignment === 'top' ? alignment : 'side';
   const hasLabel = (label?.length > 0 && width > 0) || (auto && width == 0 && label && label?.length != 0);
-  const hasValue = value !== '' && value !== null && value !== undefined;
+  const hasValue =
+    value !== '' && value !== null && value !== undefined && !Number.isNaN(value);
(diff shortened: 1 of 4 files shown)


RudraPratik30:

PR #17685, opened 2026-08-26
Nav Enhancement: itemId and group expand behaviour updated

jhamadhav:
# nav_enhancement: Nav widget parity with Tabs and Page/Navigation

## Summary

Three enhancements to the Nav (Navigation) widget, bringing it closer in behavior to the Tabs widget and the app's Page/Navigation menu: editable/accessible item ids, auto-hiding empty groups, and accordion-style group collapse on selection.

## What's Implemented

- **Item ids are now editable and addressable.** Nav items already had internal `id`s, but they weren't user-editable or usable from actions. The item edit popover now has an "Id" field (validated for emptiness/uniqueness/no dynamic bindings). Two new actions, `Set item visibility` and `Set item disable`, take an item id and mutate that specific item at runtime.
- **A group auto-hides when it has no visible, enabled children.** Mirrors Page/Navigation's rule: a group is rendered only if it isn't itself hidden AND at least one child is both visible and enabled.
- **Selecting an item collapses other open groups (vertical/accordion mode).** Selecting an item inside a group collapses every other group, leaving that item's parent group open. Selecting a top-level item collapses all groups. Manually clicking a group header to expand/collapse it is unaffected — that stays independent per group.
- **Id validation now rejects `{{ }}` dynamic bindings.** Item/tab ids are compared with plain equality everywhere at runtime (dnd, event refs, action params) and are never resolved, so a binding can never work as an id — it's now rejected outright instead of being silently accepted and breaking downstream lookups. Applied to both the Nav widget and the Tabs widget (which had the identical gap) via a shared validator.

## Areas to Test

- Add/edit/delete nav items and groups; edit an item's id and confirm click events still fire correctly after the rename.
- Try setting a duplicate id, an empty id, and an id containing `{{ }}` in both the Nav item popover and the Tabs tab-id field — all three should be rejected with a clear message.
- Call `setItemVisibility`/`setItemDisable` from another component's action and confirm the targeted item (including a child inside a group) updates without affecting siblings.
- Hide all children of a group (individually, or via the new actions) and confirm the group itself disappears from both horizontal and vertical layouts, including the overflow "More" menu.
- In vertical orientation: select an item inside Group A while Group B is open — confirm Group B collapses and Group A stays open; then select a top-level item and confirm all groups collapse.
- Regression: horizontal orientation's dropdown-style groups (unaffected by the collapse logic) still open/close normally via click and outside-click.

### DIFF
--- a/frontend/src/AppBuilder/RightSideBar/Inspector/Components/Navigation/Navigation.jsx
+++ b/frontend/src/AppBuilder/RightSideBar/Inspector/Components/Navigation/Navigation.jsx
@@ -34,6 +34,7 @@ export const Navigation = ({ componentMeta, darkMode, ...restProps }) => {
     handleAddItemToGroup,
     handleReorder,
     getResolvedValue,
+    validateItemId,
   } = useMenuItemsManager(component, paramUpdated);
 
   // Property organization
@@ -60,6 +61,7 @@ export const Navigation = ({ componentMeta, darkMode, ...restProps }) => {
         onMouseLeave={() => setHoveredItemIndex(null)}
         onDeleteItem={handleDeleteItem}
         onItemChange={handleItemChange}
+        validateItemId={validateItemId}
         onAddItem={handleAddItem}
         onAddGroup={handleAddGroup}
         onAddItemToGroup={handleAddItemToGroup}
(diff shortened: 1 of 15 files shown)


jhamadhav:
Changes moved to PR: https://github.com/ToolJet/ToolJet/pull/17590

PR #17688, opened 2026-08-26
Fix: don't hard-fail startup when RDS cert download has no internet access

adishM98:
📝 What this does

Firewalled/offline deployments crashed on boot — the entrypoint downloaded the AWS RDS cert bundle at container startup and `exit 1`'d on failure, even for installs that don't use RDS. Now the bundle is baked into the image at build time (build stage has internet) as a guaranteed fallback, and the entrypoint only attempts an opportunistic runtime refresh to pick up AWS rotations — never blocks startup either way.

🏗️ Architecture

```mermaid
sequenceDiagram
    participant Build as Image build
    participant Entry as Container entrypoint
    participant AWS as AWS RDS truststore

    Build->>AWS: wget global-bundle.pem
    alt reachable
        AWS-->>Build: bundle baked into image
    else unreachable
        Build->>Build: warn, continue build
    end

    Entry->>AWS: wget global-bundle.pem.tmp
    alt reachable
        AWS-->>Entry: mv .tmp -> bundle (refreshed)
    else unreachable (firewalled)
        Entry->>Entry: keep baked-in bundle
    end
    Entry->>Entry: export NODE_EXTRA_CA_CERTS if bundle exists
    Entry->>Entry: exec app (never exits 1 on cert failure)
```

Design considerations

- **Why bake at build time instead of just catching the error?** A caught error alone leaves zero-internet deployments with no RDS cert at all. Baking at build time (where internet is guaranteed) gives them a working fallback instead of silently missing SSL trust for RDS.
- **Why still attempt a runtime refresh?** AWS rotates this bundle occasionally (~1yr notice before old CAs retire). Refresh-if-reachable keeps online deployments current; offline ones fall back to whatever was baked in at the last image build — same staleness tradeoff as the OS `ca-certificates` package.
- **Why atomic `.tmp` + `mv` instead of overwriting in place?** A wget that dies mid-write must never leave a half-written/corrupt cert file in place of a good baked-in one.
- **Why not fail the image build if the bake-time download fails?** A flaky build host hitting AWS shouldn't kill the whole image build for what's an optional fallback — warns and continues, runtime refresh still gets a shot later.

🔀 Changes

- Baked AWS RDS global cert bundle into the image at build time as a fallback
- Entrypoint now treats the runtime download as a refresh, not a requirement — failure warns and keeps the baked-in bundle instead of `exit 1`
- `NODE_EXTRA_CA_CERTS` only exported when a bundle actually exists (build-time or refreshed)

🧪 How to test

- [x] Build image normally, confirm `/home/appuser/certs/global-bundle.pem` present before container start
- [x] Start container with outbound internet blocked, confirm boot succeeds and logs the refresh-failed warning instead of exiting
- [x] Start container with internet available, confirm bundle gets refreshed (`mv` log line, mtime updates)
- [x] Connect to an actual RDS instance over SSL with internet blocked, confirm baked-in cert is sufficient


slancerk (dismissed):
All test passed ✅.

PR #17691, opened 2026-08-26
Radio v2: copy issue fixed

jhamadhav:
# radiov2-paste-options-bug: RadioButtonV2 (and DropdownV2/MultiselectV2) regain a default option on copy-paste

Issue: (no tracked ticket)

## Issue Description

Editing a RadioButton v2 widget down to 2 options, then copying and pasting it, produces a pasted copy with 3 options — the widget's default option reappears even though it was removed from the original instance.

## Root Cause

`pasteComponents` in `frontend/src/AppBuilder/AppCanvas/copyPasteWidgetsUtils.js:346` builds the pasted component's data with:

```js
const componentData = _.merge({}, componentMeta, component.component);
```

`componentMeta` is the widget's static default definition (from the widget registry), and `component.component` is the actual copied instance. Lodash's `_.merge` deep-merges arrays **by index**, not by replacement:

```js
_.merge({}, { arr: [1,2,3] }, { arr: [10,20] })  // => { arr: [10, 20, 3] }
```

RadioButtonV2's default `definition.properties.options.value` is a 3-entry array (`frontend/src/AppBuilder/WidgetManager/widgets/radioButtonV2.js:315-339`). When the instance being pasted has only 2 options, the merge overwrites indices 0 and 1 with the real data but has no source element for index 2, so the default's 3rd entry survives into the merged result.

The copy path is unaffected — it reads live store data and serializes it whole to the clipboard with no loss. The bug is entirely in this paste-side merge.

`DropdownV2` and `MultiselectV2` share the same `options` array-of-objects shape and are equally affected. `RadioButton v1` is unaffected because it stores options as a template string (e.g. `'{{[true,false]}}'`), which `_.merge` treats as an atomic value rather than recursing into.

Notably, an identical bug for the `Table` widget's `columns` array was already worked around a few lines below (`copyPasteWidgetsUtils.js:347-352`), but `options` was never added to that override.

## Fix

Generalized the existing Table-only workaround in `copyPasteWidgetsUtils.js` (right after the `_.merge` call) to also override any `options` array property with the pasted value directly (deep-cloned), bypassing the index-merge:

```js
const pastedOptions = component.component?.definition?.properties?.options;
if (Array.isArray(pastedOptions?.value) && componentData?.definition?.properties?.options) {
  componentData.definition.properties.options = deepClone(pastedOptions);
}
```

This keys off the presence of an `options.value` array rather than hardcoding component names, so it fixes `RadioButtonV2`, `DropdownV2`, and `MultiselectV2` in one change.

## Areas to Test

- Copy/paste and cut/paste a RadioButtonV2 with fewer options than the default (e.g. 2 instead of 3).
- Same for DropdownV2 and MultiselectV2.
- Copy/paste with *more* options than the default, to confirm no options are dropped.
- Regression: Table widget column copy/paste still works correctly (shares the surrounding code path).
- Regression: RadioButton v1 (template-string based options) still pastes correctly.

### DIFF
--- a/frontend/src/AppBuilder/AppCanvas/__tests__/integration/pastePreservesArrayOptions.spec.js
+++ b/frontend/src/AppBuilder/AppCanvas/__tests__/integration/pastePreservesArrayOptions.spec.js
@@ -1,0 +1,72 @@
+/**
+ * Regression: pasting a widget whose `options` array (RadioButtonV2,
+ * DropdownV2, MultiselectV2) was edited to fewer entries than the widget's
+ * default brings back the missing default option(s). Cause: pasteComponents()
+ * merges the pasted instance onto the widget's default config with
+ * `_.merge`, which merges arrays by index instead of replacing them.
+ */
+import { http, HttpResponse } from 'msw';
+import config from 'config';
+import '@/test/setupMsw';
+import { server } from '@/test/msw/server';
+import useStore from '@/AppBuilder/_stores/store';
+import { componentTypeDefinitionMap } from '@/AppBuilder/WidgetManager';
+import { pasteComponents } from '@/AppBuilder/AppCanvas/copyPasteWidgetsUtils';
+import { seedApp } from '@/test/app-builder';
+import { deepClone } from '@/_helpers/utilities/utils.helpers';
+
+const state = () => useStore.getState();
+
+describe('paste preserves an edited array-valued options property', () => {
+  beforeEach(() => {
+    // Paste always fires a real batch-save call; stub it so only the merge bug can fail this test.
+    server.use(
+      http.put(`${config.apiUrl}/v2/apps/:appId/versions/:versionId/components/batch`, () => HttpResponse.json({}))
+    );
+  });
+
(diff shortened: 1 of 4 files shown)


jhamadhav:
Test added for bug fix.

Before fix:
```
FAIL src/AppBuilder/AppCanvas/__tests__/integration/pastePreservesArrayOptions.spec.js
  paste preserves an edited array-valued options property
    ✕ a RadioButtonV2 edited down to 2 options still has 2 options after paste (55 ms)

  ● paste preserves an edited array-valued options property › a RadioButtonV2 edited down to 2 options still has 2 options after paste

    expect(received).toHaveLength(expected)

    Expected length: 2
    Received length: 3
    Received array:  [{"label": "option1", "value": "1", ...}, {"label": "option2", "value": "2", ...}, {"label": "option3", "value": "3", ...}]

      67 |
      68 |     expect(pasted).toBeDefined();
    > 69 |     expect(pasted.component.definition.properties.options.value).toHaveLength(2);
         |                                                                  ^

Tests: 1 failed, 1 total
```

nakulnagargade (dismissed):

PR #17708, opened 2026-08-27
Page Switch batch issue fixed

jhamadhav:
# page_switch_batch_issue: Page-switch exposed-value batch leak freezes app state

## Issue Description

Customer reported that app state stops updating reliably in the App Builder editor:

- A Table's `selectedRow` doesn't update consistently after clicking a row.
- The component Inspector doesn't always show all of a component's fields — some are missing intermittently.
- The "Open Modal" / "Close Modal" event action dropdown doesn't always work.

The issue was intermittent. It initially seemed tied to high memory usage (80-85%) on the customer's machine, but it came back later even under 50% memory usage. A full reload temporarily fixed it.

### How a page switch is supposed to work

When you switch pages, ToolJet doesn't write every widget's data to the store one change at a time. Instead, it briefly "holds" all of it in memory, and only writes everything to the store in one go once the new page has fully finished loading (all its lazy-loaded components have resolved). This is just a performance optimization — apply a burst of changes together, and recompute anything that depends on them exactly once, instead of once per tiny update.

This "hold" period is tracked with a simple counter:
- Starting the hold adds 1 to the counter.
- Finishing (the new page has fully loaded) brings the counter back down by 1.
- While the counter is above 0, nothing gets written to the store — everything just sits in memory, waiting.

### What goes wrong

The bug happens when you switch pages quickly, before the previous page has finished loading.

Every switch adds 1 to the counter when it starts. But there's only **one** thing in the whole app watching "has the page finished loading?" — and it only ever watches the page you're *currently* on. If you switch away again before that watcher fires, it gets thrown away with the old page. It never fires, so its matching "-1" never happens either.

**Example:**
1. Switch to Page B → counter goes to 1. Page B starts loading.
2. Before Page B finishes, switch to Page A → counter goes to 2. Page B's "loading" watcher is discarded — abandoned mid-flight.
3. Only Page A eventually finishes loading, bringing the counter down to 1 — never back to 0.

Once the counter gets stuck above 0, it never recovers. From that moment on, **every write from every widget, anywhere in the app, on any page, gets held in memory and never actually applied.** There's no error, no warning, nothing visibly wrong — the app just silently stops updating state until you reload the whole page.

This explains everything the customer saw:
- **`selectedRow` never updated** — the click's write was being held, not applied.
- **Fields missing from the Inspector** — many of a widget's fields are set through a series of small updates right after it mounts. If the counter was already stuck, those updates got held and never landed.
- **Open Modal did nothing** — the modal's data was one of the writes stuck in limbo, so the code trying to open it found nothing there and failed silently.

### Why the safety check didn't catch this

There's a guard whose whole job is to block a second page switch while one is still in progress — that should have prevented this entirely. It failed for two separate reasons:

1. It was being turned off the moment a switch started its clean-up step — long before the switch had actually finished.
2. A second, unrelated piece of code was *also* turning it off almost immediately, for an entirely different reason (it just wanted to remember that a switch had happened, to decide whether to skip some startup queries afterward).

Because of these two bugs, the guard never actually stopped a second switch from starting — so nothing was in place to prevent the counter from stacking up.

## Fix

1. **`frontend/src/AppBuilder/_stores/slices/appSlice.js`** — the guard is no longer turned off during clean-up. It now only clears once the switch's held changes have actually been written to the store — not just once the switch's own code has finished running.
2. **`frontend/src/AppBuilder/_hooks/useAppData.js`** — the unrelated piece of code that was also turning off the guard now only *reads* it; it no longer changes it. Only the page-switch logic is allowed to release the guard.

### Known limitation

There's a second, separate way the "page finished loading" signal could theoretically fail to fire (a loading-tracker that's sensitive to timing under heavy load), which wasn't addressed here. Also, the counter itself is shared across four different features (app load, page switch, table/form row growth, hot reload) with no way to tell which one opened it — a more robust long-term fix would make each of them track its own handle instead of sharing one counter.

## Areas to Test

- Rapid/double-click page switching — `selectedRow`, Inspector field completeness, and Open/Close Modal should all stay correct afterward.
- Clicking the same page switch twice in the same instant still shows a "please wait" message and doesn't change the destination.
- Page switching combined with a table/form whose row count is still growing.
- A normal single page switch still behaves the same as before (startup queries/events on the new page still fire correctly).
- Full regression pass on `frontend/src/AppBuilder/_stores/__tests__/integration/pageSwitch.spec.js`, `exposedValueBatch.spec.js`, and `Widgets/NewTable/__tests__/integration/tableSelectedRow.spec.js`.

### DIFF
--- a/frontend/src/AppBuilder/AppCanvas/AppCanvas.jsx
+++ b/frontend/src/AppBuilder/AppCanvas/AppCanvas.jsx
@@ -285,7 +285,10 @@ export const AppCanvas = ({ appId, switchDarkMode, darkMode }) => {
                 >
                   {environmentLoadingState !== 'loading' && !isCanvasReloading && (
                     <SuspenseCountProvider
-                      key={currentPageId}
+                      // Also keyed on pageKey — a same-page "switch page" CSA changes pageKey
+                      // but not currentPageId, so without this the provider wouldn't remount
+                      // and the batch that switch opens would never flush.
+                      key={`${currentPageId}-${pageKey}`}
                       disabled={pageLoader}
                       onAllResolved={handleAllSuspenseResolved}
                       deferCheck={isModuleMode || appType === 'module'}
--- a/frontend/src/AppBuilder/AppCanvas/__tests__/SuspenseTracker.spec.jsx
+++ b/frontend/src/AppBuilder/AppCanvas/__tests__/SuspenseTracker.spec.jsx
@@ -1,0 +1,101 @@
+/**
+ * Unit coverage for SuspenseCountProvider's remount/key contract, which
+ * AppCanvas.jsx's page-switch fix depends on (AppCanvas.jsx:288).
+ *
+ * onAllResolved only ever fires once PER PROVIDER INSTANCE — hasResolved is a
+ * one-shot latch (SuspenseTracker.jsx:19-24) that never resets on its own.
+ * React only creates a fresh instance (fresh latch) when the `key` prop
+ * changes. AppCanvas.jsx used to key this provider on `currentPageId` alone.
+ * A "switch page" CSA targeting the CURRENT page mints a fresh `pageKey`
+ * (appSlice.js's isSamePage branch) but leaves `currentPageId` unchanged, so
+ * the provider never remounted, onAllResolved never fired again, and the
+ * exposed-value batch that switch opened never flushed. Fixed by keying on
+ * `${currentPageId}-${pageKey}` instead.
+ *
+ * This spec tests the underlying mechanism directly rather than through
+ * AppCanvas.jsx itself (a large, heavily-wired component) — it reproduces
+ * the same before/after key choice with a minimal harness.
+ */
+import React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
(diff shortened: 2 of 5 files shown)


kavinvenkatachalam (changes requested):
@jhamadhav One gap before merging.

The Switch Page CSA can target the `current` page (`eventsSlice.js:1047`, no same-page check). That runs the full `doSwitch`, so it opens a batch — but the flush never comes, because `SuspenseCountProvider` is keyed on `currentPageId` (`AppCanvas.jsx:288`), which doesn't change on a same-page switch. Provider never remounts, `onAllResolved` never re-fires.

Before this PR that just leaked the batch. Now `pageSwitchInProgress` also stays true, so page navigation is dead until reload.

Steps to reproduce: add button on Page A → On click → Switch page → Page A. Click it, then click a table row — `selectedRow` won't update. Then try switching pages.

Likely one line fix would solve the issue: `key={${currentPageId}-${pageKey}}`.

We can add test case for this scenario as well

jhamadhav:
@kavinvenkatachalam Replicated and fix added along with test.

kavinvenkatachalam (approved):

johnsoncherian (approved):

PR #17714, opened 2026-08-28
Fix: Prevent deletion of ModuleContainer component

vjaris42:
## 📝 What this does
Prevents the parent ModuleContainer from being deleted when editing a module — both in the UI and via API (MCP/AI).

## 🔀 Changes
- Hid the delete button in ConfigHandle when the component is a ModuleContainer
- Filtered ModuleContainer out of the deleteComponents store action so it's skipped even via keyboard shortcuts
- Added a backend guard in component.service.ts that throws BadRequestException when attempting to delete a ModuleContainer via API

## 🧪 How to test
- [ ] Open a module in the editor, verify the ModuleContainer has no delete (trash) button
- [ ] Select the ModuleContainer and press Backspace — confirm nothing happens and the dialog closes
- [ ] Call `DELETE /v2/apps/:id/versions/:versionId/components` with a ModuleContainer ID — confirm a 400 error is returned


kavinvenkatachalam (dismissed):

kavinvenkatachalam (changes requested):
@vjaris42 This spec never imports `componentsSlice.js` — it hand-writes a copy of `deleteComponents` inside `buildMockStore()` and asserts against the copy. Delete the `.filter()` you added at `componentsSlice.js:1563` and all 7 tests still pass, so it can't catch the regression it's named for. That's the mutation-verify rule in `frontend/src/test/README.md`.

Two side effects of the copy: the mock returns the filtered array and 5 tests assert on it, but the real `deleteComponents` returns nothing; and the mock's `getShouldFreeze()` drops the `isAppBeingEditedByAI` argument production passes.

`seedApp()` from `@/test/app-builder` makes the real store cheap to set up — see `slices/__tests__/integration/validateWidget.spec.js`. And since it'd then touch the real store, it belongs in `slices/__tests__/integration/`.

```js
import useStore from '@/AppBuilder/_stores/store';
import { seedApp, componentDefinition } from '@/test/app-builder';

const state = () => useStore.getState();

beforeEach(() => {
  seedApp({
    mc1: componentDefinition('mc1', 'moduleContainer1', 'ModuleContainer'),
    b1: componentDefinition('b1', 'button1', 'Button'),
  });
});

test('a ModuleContainer in the selection is skipped; siblings still delete', () => {
  state().deleteComponents(['mc1', 'b1']);
  const ids = Object.keys(state().getCurrentPageComponents('canvas'));
  expect(ids).toContain('mc1');
  expect(ids).not.toContain('b1');
});

test('selecting only the ModuleContainer deletes nothing and dismisses the dialog', () => {
  state().setSelectedComponents(['mc1']);
  state().deleteComponents([]);
  expect(Object.keys(state().getCurrentPageComponents('canvas'))).toContain('mc1');
  expect(state().showWidgetDeleteConfirmation).toBe(false);
});
```

Then mutation-verify: remove the `.filter()`, confirm both go red, restore.

PR #17776, opened 2026-09-02
Fix/ai reattach generation on refresh

swathihameed:
This PR wires the AI reattach work into the app builder, bumping both ee submodules and adding the client for the new status endpoint.

**Generation-status client**

- Added `getConversationStatus(conversationId)` to `frontend/src/_services/ai.service.js`, hitting `GET ai/conversation/:conversationId/status`

**Building overlay no longer disables the left sidebar**

- `frontend/src/AppBuilder/AgentBuildingOverlay.jsx`: dropped `.left-sidebar` from the disable list, leaving `['.right-sidebar', '[class*="query-manager"]']`
- The left sidebar is chat, debugger and history — none of it edits the app, and disabling it removed the only way to see what a build was doing. It also made a stranded indicator unrecoverable, since the chat itself was out of reach
- Fixed a docstring describing a z-11 layer that no longer exists

### DIFF
--- a/frontend/src/AppBuilder/AgentBuildingOverlay.jsx
+++ b/frontend/src/AppBuilder/AgentBuildingOverlay.jsx
@@ -6,19 +6,15 @@ import useStore from '@/AppBuilder/_stores/store';
 
 /**
  * "Agent is building" overlay for the editor. While an AI build or modification is in progress
- * (ai.isLoading in the generate conversation), it disables the canvas and the left icon toolbar and
+ * (ai.isLoading in the generate conversation), it disables the canvas and the editing panels and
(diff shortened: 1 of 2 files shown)


kavinvenkatachalam (dismissed):

akshaysasidrn (commented):
Wiring only, looks fine. Keeping the left sidebar usable during a build is the right call; the chat is the only place to see what's happening.

Good to approve once ToolJet/ee-server#810 and ToolJet/ee-frontend#623 land and the submodule pointers are bumped.

One knock-on for #623, noted there: with the sidebar live mid-build, "new chat" becomes reachable and hits the `isEditorFreezed` release path.

PR #17786, opened 2026-09-02
Integrate AI builder recovery across ToolJet

shaurya-sharma064:
## Fixes
- Wires the active-generation status call into ToolJet.
- Updates the EE frontend and server revisions for refresh-safe AI builds.
- Keeps the left sidebar available while a build runs so chat and history remain reachable.

## Performance impact
- Only active AI builds perform status polling.
- No impact on normal editor or runtime traffic.

## Integrated source
- #17776

Deploy the tooljet-agent integration before or together with this ToolJet integration.


plugins/packages/common/lib/query.error.ts
```diff
@@ -1,19 +1,111 @@
...
+      if (['body', 'responsebody', 'headers', 'requestobject'].includes(field)) continue;
+      visit(child, depth + 1);
+    }
+  };
+  values.forEach((value) => visit(value));
+  return { codes, statuses };
+}
+
+/** Classifies only structured driver/API fields. Human-readable messages are never inspected. */
+export function classifyQueryErrorCategory(data: unknown, description?: unknown): QueryErrorCategory {
```
akshaysasidrn: **Suggest reverting.** Error codes belong in the plugin that owns the driver, not in a central list.

Ten driver families share one namespace here, and the bare numbers collide:

```
13 → Mongo Unauthorized      → 'authentication'
13 → gRPC INTERNAL           → also 'authentication'   ✗
4  → gRPC DEADLINE_EXCEEDED  → 'timeout'
```

`structuredErrorFacts` also reads any `code`/`name`/`type`/`number`/`state` four levels deep, so a nested `{ code: 4 }` from a REST plugin reads as `timeout`.

A table can't reach everyone either: `restapi`, `grpc` and ~26 others throw with an empty `data`.

Better settled in the query-error standardisation work. One thing to carry into it: the frontend puts the whole failed result in `queries.<id>.error` and the code hinter lists its keys, so whatever we name becomes visible to builders and hard to change after.

akshaysasidrn (commented):
**`b46462a`: keep.** "This plugin has no connection test" is the missing piece, and it lets the MCP drop its regex-built catalog.

**`4f42806`: we'd suggest reverting.** Right goal, but error codes can't be hardcoded in core. They belong in the plugin that owns the driver. Plugins are contributed, so a central list means a core PR for every new plugin or driver code.

This fits the query-error categorisation and standardisation work, where the interfaces and types get decided properly. Fine to leave until then.

**Reverting costs users nothing.** Nothing reads `category` yet:

```
MySQL, wrong password              today     with 4f42806
  agent says "the query failed"    same      same
  no "open datasource settings"     ✗         ✗
```

That missing link is the MCP classifying only from `data.code`, while MySQL puts SQLSTATE in `data.sqlState`. Fixable in the MCP alone. MSSQL uses its own codes, which is the per-driver knowledge that has to sit in the plugin.

Separately: the two new server specs have no `@group` tag, so CI (`working`/`workflows`/`security`) skips them. No CI job runs the plugin test either.

PR #17859, opened 2026-09-08
Fix: Treat table search highlights as literal text

Creatixpy:
## 📝 What this does
Searching a table for `(` or `C++` can crash matching cells because the search highlighter interprets the term as a regular expression. Other terms, such as `(draft)`, duplicate displayed text. Treat search terms literally, matching the table's existing substring filter.

This isolates the table-highlighting bug reported by @PedroHenrique0713 in #17269, which was closed after going unreviewed. The bug is still present on current `lts-3.16` (61cb5d6f77) and `main`; this PR adds regression coverage for that specific behavior.

## 🔀 Changes
- Escaped regex metacharacters before splitting cell text, following the existing dropdown highlighting pattern.
- Returned unhighlighted text for missing search terms, including the `undefined` value produced when clearing search.
- Added tests for literal punctuation, repeated matches, case preservation, numeric values, empty or unmatched searches, and clearing an active search.

## 🧪 How to test
- [x] Run `cd frontend && npm test -- --runInBand --runTestsByPath src/AppBuilder/Widgets/NewTable/_components/HighLightSearch/__tests__/HighLightSearch.spec.jsx`: 26 passed. The original 23-test suite had 12 failures against the unchanged base; all three added missing/cleared-search tests failed before the corrective guard.
- [x] Run ESLint on both changed files and `git diff --check`.
- [ ] In a Table widget, use rows containing `Order (draft)`, `C++`, and `file [draft]`. Enable search and search for `(`, `C++`, and `[draft]`. Matching cells should render with only the literal matches highlighted and the original text intact.
- [ ] With a cell containing `undefined value`, search for `value` and then clear the search. The full cell text should remain visible without highlights or a rendering error.

Validation used React Testing Library with the repository's Jest configuration. A full ToolJet browser session and the full frontend suite were not run.

Reviewed with the help of AI.


PedroHenrique0713:
Thanks for picking this one up, and for writing the test. My PR touched the same line but shipped no coverage for it, which is a fair part of why it was easy to skip.

Re-checked today: `HighLightSearch.jsx` is byte identical on `main` (1794841c3c) and `lts-3.16` (7f82c1fba5), so this lands the same way on `main` if it is wanted there too.

Your `searchTerm == null` guard also covers a case mine did not. With `searchTerm` undefined, `searchTerm?.toLowerCase()` is `undefined`, and `String.prototype.includes(undefined)` coerces it to the literal `"undefined"`, so the early return is skipped and the cell highlights the word inside its own text:

```js
"undefined value".includes(undefined)                    // true
"undefined value".split(/(undefined)/gi)                 // ["", "undefined", " value"]
```

On the crash itself, measured on Node 24: `new RegExp("(C++)", "gi")` throws `SyntaxError: Nothing to repeat` and `"(draft"` throws `Unterminated group`, while `"(draft)"` does not throw at all, it silently duplicates the rendered text (`"Before (draft) after"` splits into `["Before (", "draft", "draft", ") after"]`). So the same bug is a crash for some terms and a wrong render for others.

#17269 was closed with five other findings unreviewed. All five are still live on `main` at 1794841c3c, so I am listing them here rather than letting them go down with that PR, in case anyone wants to isolate them the way you did:

- `frontend/src/_helpers/http-client.js:104`: the catch block runs `JSON.parse(text)` on the same body that just failed to parse, so a non-JSON error response (an HTML 502 page) throws a second time out of the catch.
- `frontend/src/_lib/generate-file.js:42`: `Object.keys(value[0])` throws `TypeError` when the array is empty, which is a valid query result, and it breaks PDF export.
- `frontend/src/_stores/dataQueriesStore.js:603-618` and `frontend/src/AppBuilder/_stores/slices/dataQuerySlice.js:667-687`: `sortByAttribute` calls `data.sort()` in place on the array it receives from store state, so sorting mutates the state array instead of returning a new one.
- `frontend/src/AppBuilder/Widgets/DropdownV2/utils.js:74-89`: same in place `arr.sort()` in `sortArray`, which is used by dropdown, multiselect and tag input.

One of the six I would not copy as it was written: the `componentsSlice.js` change added optional chaining to four `modules[moduleId].pages` accesses, but that file has around twenty accesses of the same shape. Four guards is a patch on the symptom, not a fix. What that path needs is a decision on what should happen when `moduleId` is stale, not more `?.`.

Worth noting for whoever reviews this: `escapeRegExp` already exists at `frontend/src/AppBuilder/Widgets/DropdownV2/utils.js:52-57`, and `highlightText` right below it already escapes, so the table widget was the one place left unguarded. Sharing that helper instead of keeping two copies would be the follow-up, but that is a maintainer call.

PR #17866, opened 2026-09-08
Fix: block edits to released app version from non-UI clients

vjaris42:
## 📝 What this does
Released app versions could still be edited by non-UI clients (MCP, PAT tokens, the public API) because the freeze was only enforced in the frontend. This adds a backend guard that blocks those edits at the API boundary.
- [ee-server](https://github.com/ToolJet/ee-server/pull/823)
- [ee-frontend](no changes)

## 🔀 Changes
- Added MutableAppVersionGuard that rejects mutations targeting an app's released version
- Wired it onto all component, page, event, and query create/update/delete routes (CE + EE)
- Preserved the is_user_switched_version escape hatch used by version switching
- Added unit tests for released vs draft versions and the escape hatch

## 🧪 How to test
- [ ] On a released app version, call a component/page/query/event write endpoint with a PAT — expect 400 "You cannot update a released version"
- [ ] Repeat on a draft version — expect success
- [ ] Run: cd server && NODE_ENV=test NODE_OPTIONS='--max-old-space-size=8192' npx jest --config jest.config.ts test/modules/app/unit/mutable-app-version.guard.spec.ts


kavinvenkatachalam (changes requested):
**1. The guard is opt-out**

```ts
if (request.body?.is_user_switched_version === true) return true;
```

Any PAT/MCP client can bypass the guard with one line of JSON — the exact clients this PR is meant to block. It also isn't needed on these routes:

- The UI hardcodes `false` (`componentsSlice.js:2531`, `pageMenuSlice.js:15` — never set to `true` anywhere)
- The flag's only real use is the version-save route (`appVersion.service.js:93` → `PUT /v2/apps/:id/versions/:versionId`), which this PR doesn't guard
- It's `@IsBoolean()` and **not** `@IsOptional()` on the component DTOs, so clients are effectively invited to set it

**Suggestion:** drop the escape hatch from this guard.

**2. The query "run" endpoint still writes to a released version**

`POST /data-queries/:id/versions/:versionId/run/:environmentId` didn't get the guard, but it persists:

```ts
// server/src/modules/data-queries/service.ts:251
if (mode !== 'view' && ability.can(...) && !isEmpty(options)) {
  await this.dataQueryRepository.updateOne(dataQueryId, { options });
```

Same threat model, still open. Either add the guard or gate the write on the version being mutable.

**3. `global_settings` / `page_settings` / version-update are unguarded**

In `versions/controller.v2.ts`, these all reach `versionsUtilService.updateVersion`, which writes `globalSettings`, `pageSettings`, `homePageId` and `showViewerNavigation` with no released-version check for non-workflow apps (`updateWorflowVersion`'s check is workflow-only *and* scoped to `definition`/`currentEnvironmentId`):

- `PUT :id/versions/:versionId`
- `PUT :id/versions/:versionId/global_settings`
- `PUT :id/versions/:versionId/page_settings`

MCP's `update_app_settings` / `update_pages` land here.

> ⚠️ Don't blanket-guard `PUT :id/versions/:versionId` — the UI legitimately renames and changes status/description on a released version. This one needs a field-scoped check like the workflow path, not the guard.

**4. Guard order — run it after `FeatureAbilityGuard`**

Currently a user with no edit permission gets `400 You cannot update a released version` instead of `403`. That leaks release state to unauthorized callers and muddies status codes. Authorize first, then apply the business rule.

**5. The fallback ordering is backwards**

```ts
const editingVersionId = app.appVersions?.[0]?.id ?? request.params?.versionId;
```

`appVersions` depends on which upstream guard loaded the app — `findAppFromVersion` (used by `ValidateAppVersionGuard` / `ValidateQueryAppGuard`) returns `appVersion.app` with the relation **not loaded**, and `findByDataQuery` without a `versionId` can hydrate an arbitrary version row.

Every wired route has `:versionId`, and that param *is* the mutation target — read it first, use `appVersions[0]` only as the fallback.

**6. Wiring is fragile**

20+ hand-edited `@UseGuards` lists across CE + EE, with EE re-declaring the same lists. The next mutation route will silently miss the guard. A composed decorator would make safe the default and shrink both diffs:

```ts
@AppVersionMutation() // ValidAppGuard, MutableAppVersionGuard, FeatureAbilityGuard, GitSyncEditGuard
```

**7. Tests don't cover the part that breaks**

The spec unit-tests a hand-mocked `ExecutionContext`, so it would still pass if every `@UseGuards` edit were reverted. Worth adding:

- [ ] One route-level (supertest) test hitting a released version
- [ ] A case where `appVersions[0]` differs from `params.versionId` (see 5)

kavinvenkatachalam (dismissed):

PR #17878, opened 2026-09-09
feat: app-scoped session from a workspace PAT

swathihameed:
## 📝 What this does
A workspace personal access token can now mint a session pinned to one app, so a headless browser can open that app for the AI builder's render check. The PAT scope interceptor no longer treats "has an appId" as proof of the embed flow; the JWT now carries the token's own kind, and the pair decides the session's limits.
- [ee-server](https://github.com/ToolJet/ee-server/pull/829)
- [ee-frontend](no changes)

## 🔌 API Reference
| Method | Route | Permission | Request | Response |
|--------|-------|------------|---------|----------|
| **POST** | `/api/personal-access-tokens/session` | workspace PAT (`Authorization: Bearer tj_pat_…`) | `{ appId?: uuid }` | `{ authToken }`; 400 on a malformed id, 404 when the app is not in the token's workspace |

## 🔀 Changes
- Session exchange accepts an optional `appId`; the app must belong to the token's workspace.
- JWT gains `patScope` (the token's kind) so the interceptor keys on `(patScope, appId)` instead of `appId` alone.
- App-pinned sessions get their own module list, narrowed by feature for `AUTH`, `ORGANIZATION_CONSTANT` and `VERSION`; read-only apart from query runs; pinned to `/apps/<uuid>` routes.
- Sessions minted before `patScope` existed pass through unchanged (transitional).
- Unit spec covers the three session kinds, the feature narrowing, and the rule that naming an app never widens what the token could reach.

## 🧪 How to test
- [ ] Mint a workspace PAT, exchange it with `{ appId }`, decode the JWT: `patScope: workspace`, `appId` set
- [ ] Exchange with an app from another workspace: 404
- [ ] Exchange with `appId: "not-a-uuid"`: 400
- [ ] Open the app's editor in a browser with that session: page renders, queries run
- [ ] Same session: `GET /api/organization-constants/decrypted?type=Secret` and `GET /api/switch/<workspace>` return 403; `PUT /api/apps/<id>` returns 403
- [ ] Exchange without `appId`: behaviour unchanged (automation ceiling)

### DIFF
--- a/server/test/modules/personal-access-tokens/e2e/pat-session.spec.ts
+++ b/server/test/modules/personal-access-tokens/e2e/pat-session.spec.ts
@@ -1,7 +1,9 @@
 import { INestApplication } from '@nestjs/common';
 import { createUser, initTestApp, login, closeTestApp, getDefaultDataSource } from 'test-helper';
+import { createApplication } from '../../../helpers/seed';
 import { UserPersonalAccessToken } from '@entities/user_personal_access_tokens.entity';
 import { OrganizationUser } from '@entities/organization_user.entity';
+import { User } from '@entities/user.entity';
 import * as request from 'supertest';
 
 /**
@@ -32,6 +34,24 @@ describe('Personal access token session exchange', () => {
       throw new Error(`createPat expected 201, got ${res.status}: ${JSON.stringify(res.body)}`);
     }
     return { token: res.body.token, id: res.body.id };
+  };
+
+  const createApp = async (name: string): Promise<string> => {
+    const owner = await getDefaultDataSource().getRepository(User).findOne({ where: { id: userId } });
+    const application = await createApplication(app, { name, user: owner });
+    return application.id;
+  };
+
+  // A real app owned by another workspace, not just an unknown id.
+  const createAppInOtherWorkspace = async (): Promise<string> => {
+    const { user: outsider } = await createUser(app, {
+      email: 'outsider@tooljet.io',
+      firstName: 'other',
+      lastName: 'workspace',
+      organizationName: 'Other workspace',
+    });
+    const application = await createApplication(app, { name: 'foreign-app', user: outsider });
+    return application.id;
   };
 
   const exchange = (token: string) =>
@@ -88,6 +108,40 @@ describe('Personal access token session exchange', () => {
       expect(payload.appId).toBeUndefined();
       // Attribution: PAT-driven writes must be distinguishable from a human's in audit logs.
       expect(payload.tj_api_source).toBe('personal_access_token');
+    });
+
+    it('should mint an app-scoped session for an app in the token workspace', async () => {
+      const { token } = await createPat('app-scoped');
+      const appId = await createApp('render-target');
+
+      const res = await exchange(token).send({ appId }).expect(201);
+
+      const payload = JSON.parse(Buffer.from(res.body.authToken.split('.')[1], 'base64').toString());
+      expect(payload.isPATLogin).toBe(true);
+      expect(payload.appId).toBe(appId);
+      expect(payload.scope).toBe('App');
+      expect(payload.organizationIds).toEqual([orgId]);
+    });
+
+    it('should refuse an app in another workspace', async () => {
+      const { token } = await createPat('foreign-app');
+      const foreignAppId = await createAppInOtherWorkspace();
+
+      await exchange(token).send({ appId: foreignAppId }).expect(404);
+    });
+
+    it('should refuse an app that does not exist', async () => {
+      const { token } = await createPat('missing-app');
+      await exchange(token).send({ appId: '00000000-0000-0000-0000-000000000000' }).expect(404);
+    });
+
+    it('should still mint a workspace session when no app is named', async () => {
+      const { token } = await createPat('no-app-named');
+      const res = await exchange(token).expect(201);
+
+      const payload = JSON.parse(Buffer.from(res.body.authToken.split('.')[1], 'base64').toString());
+      expect(payload.appId).toBeUndefined();
+      expect(payload.scope).toBeUndefined();
     });
 
     it('should reject an expired token', async () => {
--- a/server/src/modules/personal-access-tokens/interface/IService.ts
+++ b/server/src/modules/personal-access-tokens/interface/IService.ts
@@ -39,7 +39,9 @@ export interface IPersonalAccessTokensService {
   createPat(user: User, organizationId: string, name: string, expiresAt: Date): Promise<PatView & { token: string }>;
   listPats(userId: string): Promise<PatView[]>;
   deletePat(userId: string, id: string): Promise<void>;
-  createSessionFromPat(pat: UserPersonalAccessToken, response: Response): Promise<PatSession>;
+  /* `appId` mints an app-scoped session instead of a workspace one, for an app in the token's own
+     workspace. */
+  createSessionFromPat(pat: UserPersonalAccessToken, response: Response, appId?: string): Promise<PatSession>;
   /* Find-or-create the single service token for (user, workspace, name). Returns the ENTITY, with
      `user` populated, because the caller mints a session from it in-process via createSessionFromPat
      — no raw token is produced and none is needed. */
--- a/server/src/modules/personal-access-tokens/module.ts
+++ b/server/src/modules/personal-access-tokens/module.ts
@@ -3,6 +3,7 @@ import { SubModule } from '@modules/app/sub-module';
 import { SessionModule } from '@modules/session/module';
 import { UserRepository } from '@modules/users/repositories/repository';
 import { OrganizationRepository } from '@modules/organizations/repository';
+import { AppsRepository } from '@modules/apps/repository';
 import { FeatureAbilityFactory } from './ability';
 
 export class PersonalAccessTokensModule extends SubModule {
@@ -17,7 +18,7 @@ export class PersonalAccessTokensModule extends SubModule {
       module: PersonalAccessTokensModule,
       // SessionModule: the PAT -> session exchange mints a normal session via SessionUtilService.
       imports: [await SessionModule.register(configs)],
-      providers: [PersonalAccessTokensService, FeatureAbilityFactory, UserRepository, OrganizationRepository],
+      providers: [PersonalAccessTokensService, FeatureAbilityFactory, UserRepository, OrganizationRepository, AppsRepository],
       controllers: isMainImport ? [PersonalAccessTokensController] : [],
       exports: [PersonalAccessTokensService],
     };
--- a/server/src/modules/personal-access-tokens/service.ts
+++ b/server/src/modules/personal-access-tokens/service.ts
@@ -23,7 +23,11 @@ export class PersonalAccessTokensService implements IPersonalAccessTokensService
     throw new Error('Method not implemented.');
   }
 
-  async createSessionFromPat(pat: UserPersonalAccessToken, response: Response): Promise<PatSession> {
+  async createSessionFromPat(
+    pat: UserPersonalAccessToken,
+    response: Response,
+    appId?: string
+  ): Promise<PatSession> {
     throw new Error('Method not implemented.');
   }
 


server/test/modules/personal-access-tokens/e2e/pat-session.spec.ts
```diff
@@ -9,9 +9,11 @@ import {
   ensureAppEnvironments,
   createGroupPermission,
 } from 'test-helper';
+import { createApplication } from '../../../helpers/seed';
 import { UserPersonalAccessToken } from '@entities/user_personal_access_tokens.entity';
 import { User } from '@entities/user.entity';
 import { OrganizationUser } from '@entities/organization_user.entity';
+import { User } from '@entities/user.entity';
```
akshaysasidrn: The new e2e tests have not run: the file has a duplicate import.

`User` is imported twice (lines 14 and 16). `tsc` rejects that (TS2300). Jest does not, because `test/jest-e2e.config.ts` sets `diagnostics: false`. The PR notes say the e2e suite was not run locally, and CI has not run it either.

These tests are the only ones covering the checks this PR adds: app in another workspace (404), unknown app (404), bad id (400).

```suggestion
```

Could we paste the e2e run output on the PR?

server/src/modules/personal-access-tokens/interceptors/pat-scope.interceptor.ts
```diff
@@ -48,4 +101,93 @@ export class PatScopeInterceptor implements NestInterceptor {
...
+ * the verdict it WOULD have received. One boot of the player then yields the complete list of what
+ * it touches, instead of one module per run: a denial redirects the SPA to login and stops the
+ * boot, so collecting 403s the obvious way needs a run per module.
+ *
+ * Remove this, and the env var, once PAT_APP_VIEWER_MODULES is settled. It must never be set
+ * anywhere but a developer's machine — it disables the limit it is measuring.
+ */
+function auditOnly(): boolean {
+  return process.env.PAT_SCOPE_AUDIT === 'true';
+}
```
akshaysasidrn: One env var turns off every limit this PR adds, on any server.

With `PAT_SCOPE_AUDIT=true`, `intercept` logs what it would have done and lets the request through anyway (lines 81 to 90). Nothing limits this to development. Copy a developer `.env` to production and every app-pinned session gets the token owner's full role, plus every request URL goes to stdout.

**Impact.** Silent. Whoever holds an app-pinned session on that server can read every workspace secret, edit every app, and switch into other workspaces, with no error and no log line except the audit print.

The suggestion also trims the doc block to the one line a reader cannot get from the code.

```suggestion
// TEMPORARY dev-only audit: logs verdicts instead of denying; never set in prod; remove once list settles
function auditOnly(): boolean {
  return process.env.NODE_ENV !== 'production' && process.env.PAT_SCOPE_AUDIT === 'true';
}
```

**Also on these lines.** `logAudit` writes with `console.log`; the module's `Logger` keeps the audit lines in the same stream and format as the rest of the server output.

server/src/modules/personal-access-tokens/interceptors/pat-scope.interceptor.ts
```diff
@@ -48,4 +101,93 @@ export class PatScopeInterceptor implements NestInterceptor {
...
+  ): string | undefined {
+    /* Read-only. A render check looks at an app; it never changes one. */
+    if (request?.method && request.method !== 'GET' && !isViewerWriteException(request)) {
+      return `This session is read-only. App-scoped render sessions may only issue GET requests.`;
+    }
+
+    /* Actually pin it to the app. Without this the session is merely narrow, not scoped — the
+       reviewer's point. Covers /api/apps/<uuid>/... only; version- and query-keyed routes do not
+       carry the app id, and their containment is the module list above plus CASL. */
+    const requestedAppId = extractAppIdFromPath(request?.originalUrl || request?.url);
```
akshaysasidrn: The app pin only works when the app's uuid is in the URL. Slug routes and query runs slip past it.

`extractAppIdFromPath` only looks for `/apps/<uuid>`. These allowed routes find the app some other way, so the pin never checks them:

```
GET  /api/apps/slugs/:slug                                    any app in the workspace
GET  /api/apps/:id                                            ValidAppGuard accepts a slug here
POST /api/data-queries/:id/versions/:versionId/run/:envId     any query in the workspace, and
POST /api/data-queries/:id/run                                this is the one write we exempt
GET  /api/v2/apps/module/by-correlation/:coRelationId/version VERSION GET_ONE, no app id
```

Each of them already runs a guard that loads the app and sets `request.tj_app`, and the interceptor never looks at it:

```
request ──► ValidAppGuard / QueryAuthGuard ──► PatScopeInterceptor ──► handler
             sets request.tj_app                  ignores tj_app,
             (works for id, slug, query)          regexes the URL for /apps/<uuid>
                                                  slug or query URL: no match, no pin
```

Guards involved: `apps/guards/valid-app.guard.ts:48`, `app-auth.guard.ts:59`, `valid-slug.guard.ts:34`, `data-queries/guards/validate-query-app.guard.ts:57`, `query-auth.guard.ts:45`, `versions/guards/valid-module-by-correlation.guard.ts:54`. Guards run before interceptors, so by the time `denyViewer` runs, the app is already on the request.

**Impact.** An app-pinned session is meant to touch one app. The raw token travels in the iframe URL and the JWT sits in `window.name`; if either leaks, the holder can open any app in the workspace by slug and run any query in the workspace, including ones that write to the customer's database, in any environment.

Checking `tj_app` instead covers slug routes and query runs with no extra work. The regex at lines 140 to 151 stays as a fallback for routes with no app guard.

```suggestion
    /* Pin to the app. Guards already resolved it; the regex is the fallback for routes with no app guard. */
    const requestedAppId: string | undefined =
      request?.tj_app?.id ?? extractAppIdFromPath(request?.originalUrl || request?.url);
```

`request` needs `tj_app?: { id: string }` on its type at line 112. The unit spec's `run()` helper can pass `tj_app: { id: OTHER_APP_ID }` with a slug URL to cover the new path.

server/test/modules/personal-access-tokens/unit/pat-scopes.spec.ts
```diff
@@ -122,6 +129,37 @@ describe('PAT scope definition', () => {
...
+    ]) {
+      expect(patAppViewerCanAccess(MODULES.ORGANIZATION_CONSTANT, feature)).toBe(false);
+      expect(patCanAccess(MODULES.ORGANIZATION_CONSTANT, feature)).toBe(false);
+    }
+    expect(patAppViewerCanAccess(MODULES.AUTH, AUTH_FEATURE.SWITCH_WORKSPACE)).toBe(false);
+
+    for (const module of PAT_APP_VIEWER_MODULES) {
+      if (isViewerOnly(module, undefined) || VIEWER_ONLY.some(([m]) => m === module)) continue;
+      expect(patCanAccess(module)).toBe(true);
+    }
```
akshaysasidrn: This test does not check features. Add a bad feature to AUTH or ORGANIZATION_CONSTANT and it still passes.

Line 158 skips a module completely if it appears anywhere in `VIEWER_ONLY`. So for AUTH, ORGANIZATION_CONSTANT and CUSTOM_STYLES the loop checks nothing. Only the hand-written checks at lines 148 to 155 do.

Mutation, per `server/docs/testing.md` "When to delete a test": add `ORGANIZATION_CONSTANT_FEATURE.GET_PUBLIC` to the ORGANIZATION_CONSTANT feature set. The suite still passes. That one is harmless, but a dangerous one would pass the same way.

**Impact.** None today. The next time someone adds a feature to AUTH or ORGANIZATION_CONSTANT for the player, CI stays green and workspace secrets can be exposed again the way they were before this PR.

Looping over the feature map instead checks every feature. Needs `PAT_APP_VIEWER_FEATURES` exported. The suggestion keeps the explicit deny checks and drops the five-line comment at the top; the `it` name says it.

```suggestion
    const VIEWER_ONLY: Array<[MODULES, string | undefined]> = [
      [MODULES.AUTH, AUTH_FEATURE.AUTHORIZE],
      [MODULES.ORGANIZATION_CONSTANT, ORGANIZATION_CONSTANT_FEATURE.GET_FROM_APP],
      [MODULES.ORGANIZATION_CONSTANT, ORGANIZATION_CONSTANT_FEATURE.GET_FROM_ENVIRONMENT],
      [MODULES.CUSTOM_STYLES, undefined],
    ];
    const isViewerOnly = (module: MODULES, feature?: string) =>
      VIEWER_ONLY.some(([m, f]) => m === module && f === feature);

    for (const feature of [
      ORGANIZATION_CONSTANT_FEATURE.GET_DECRYPTED_CONSTANTS,
      ORGANIZATION_CONSTANT_FEATURE.GET_SECRETS,
    ]) {
      expect(patAppViewerCanAccess(MODULES.ORGANIZATION_CONSTANT, feature)).toBe(false);
      expect(patCanAccess(MODULES.ORGANIZATION_CONSTANT, feature)).toBe(false);
    }
    expect(patAppViewerCanAccess(MODULES.AUTH, AUTH_FEATURE.SWITCH_WORKSPACE)).toBe(false);

    for (const module of PAT_APP_VIEWER_MODULES) {
      const features = PAT_APP_VIEWER_FEATURES[module];
      if (!features) {
        if (!isViewerOnly(module)) expect(patCanAccess(module)).toBe(true);
        continue;
      }
      for (const feature of features) {
        if (!isViewerOnly(module, feature)) expect(patCanAccess(module, feature)).toBe(true);
      }
    }
```

server/src/modules/personal-access-tokens/interceptors/pat-scope.interceptor.ts
```diff
@@ -17,6 +23,18 @@ import { PAT_ALLOWED_BUNDLES, patCanAccess } from '@modules/personal-access-toke
...
+ * alone. Keying on appId was safe only while the embed flow was the only thing that set it; once
+ * a workspace token can pin a session to an app, "has an appId" stops meaning "is an embed
+ * session" and starts meaning nothing at all.
+ *
+ *   patScope 'app'                  embed viewer      unrestricted (a shipped feature; it runs a
+ *                                                     whole app viewer on a token bound to that
+ *                                                     app row, so the app binding is the scope)
+ *   patScope 'workspace' + appId    render check      PAT_APP_VIEWER_MODULES, pinned to that one
+ *                                                     app, read-only
+ *   patScope 'workspace', no appId  automation        PAT_ALLOWED_BUNDLES
```
akshaysasidrn: The module's `AGENTS.md` still says all app-scoped sessions are exempt. That is the bug this PR fixes.

Three rules in this PR cannot be read off the code and belong in `personal-access-tokens/AGENTS.md` (root `AGENTS.md` asks for the module doc to change in the same PR):

```
Session kind = (patScope, patAppId): app → embed, unrestricted;
workspace + appId → viewer, PAT_APP_VIEWER_MODULES, pinned, read-only;
workspace alone → PAT_ALLOWED_BUNDLES. Never key on appId alone.

PAT_APP_VIEWER_MODULES is not a PAT_BUNDLE: bundles partition modules and
exclude the credential surface; the viewer list overlaps APPS and includes AUTH.

An app-pinned session must never reach more than the same token without an
appId, except the listed viewer-only boot reads. A module with any GET on a
workspace-level path enters PAT_APP_VIEWER_MODULES only with a feature narrowing.
```

Once they are there, this block goes, and so do the two in `scopes.ts` (separate threads).

```suggestion
```

**Also on these lines.** Same rule, applied to the rest of the diff: each of these comments restates its test name, the code next to it, or history, and can go. `constants/scopes.ts` 21-23, 154, 155, 160, 161, 169-175, 179-185, 218-222, 227-230. `dto/create-pat-session.dto.ts` 4-10. This file 15, 53, 61-62, 105-109, 116, 153-156, 159-163. `interface/IService.ts` 42-43. `session/types/index.ts` 18-19. `session/util.service.ts` 113-114. `entities/user.entity.ts` 232-235. `e2e/pat-session.spec.ts` 61, 140-142. `unit/pat-scopes.spec.ts` 147, 218-221, 227-229, 235-236, 303-305, 320-321, 326-327, 332-335, 351-353, 359-360, 367-368, 381, 394-395, 414, 421-422, 429, 453, 458. Happy to open a thread per line if that is easier to action.

frontend/package.json
```diff
@@ -241,7 +241,7 @@
     "typescript": "^5.9.3",
     "webpack": "^5.75.0",
     "webpack-bundle-analyzer": "^4.10.2",
-    "webpack-cli": "^5.0.1",
+    "webpack-cli": "^5.1.4",
```
akshaysasidrn: This PR also carries a webpack-cli bump and two lockfile regenerations.

Commit 61fb51f1d3 is 4 commits ahead of lts-3.16 and not on it. It touches `frontend/package.json`, `package-lock.json`, `server/package-lock.json` (836/2504) and the `frontend/ee` pointer. None of it is about PAT sessions.

The three widget spec hunks (Cascader, Spinner, Timeline) come in through the lts-3.16 merge commit, not through that commit; worth a look that they match lts-3.16 after a rebase.

Could we drop 61fb51f1d3 from this branch and land it separately?

server/src/modules/personal-access-tokens/constants/scopes.ts
```diff
@@ -122,6 +126,116 @@ export const PAT_ALLOWED_BUNDLES: PAT_BUNDLE[] = [
...
+ * is read-only. It cannot roam the workspace and it cannot write.
+ *
+ * MEASURED, not guessed. Derived from a real editor boot with PAT_SCOPE_AUDIT=true
+ * (.agent-work/pat-viewer-audit.py): 23 requests, every one logged with the module its route
+ * carries. Re-measure rather than extend by reasoning.
+ *
+ * The EDITOR, not the viewer: an AI-built app has not been released, and /applications/<id>
+ * therefore renders "App URL Unavailable" without reaching the app at all. render_lint.app_url
+ * already opens the editor route for this reason.
+ */
```
akshaysasidrn, line 137: This block is a design note, and one line of it points at a file that is not in the repo.

The one thing here a reader cannot get from the code is "not a bundle, on purpose"; that is a module rule and goes to `AGENTS.md` (thread on the interceptor). The rest is how the list was built, what an earlier version had, and a path under `.agent-work` that is not tracked. Line 137 also says the surface includes SESSION and PROFILE, and line 184 says they were dropped.

```suggestion
```

server/src/modules/personal-access-tokens/constants/scopes.ts
```diff
@@ -122,6 +126,116 @@ export const PAT_ALLOWED_BUNDLES: PAT_BUNDLE[] = [
...
+  // Reads by app and by environment. NEVER get_decrypted or get_secrets: a render check looks at
+  // what the app painted and has no business decrypting anything.
+  [MODULES.ORGANIZATION_CONSTANT]: new Set<string>([
+    ORGANIZATION_CONSTANT_FEATURE.GET_FROM_APP,
+    ORGANIZATION_CONSTANT_FEATURE.GET_FROM_ENVIRONMENT,
+  ]),
+
+  // The single app-definition fetch a versioned boot makes.
+  [MODULES.VERSION]: new Set<string>([VERSION_FEATURE.GET_ONE]),
+};
```
akshaysasidrn, line 171: The rule here is real, but it is a module rule, not a note on this map.

"Naming an app must never widen the session" goes to `AGENTS.md` (thread on the interceptor). The ORGANIZATION_CONSTANT story is history. The inline comments at 203, 207 and 214 restate the sets and the test names ("never decrypts", "cannot switch workspace").

```suggestion
const PAT_APP_VIEWER_FEATURES: Partial<Record<MODULES, ReadonlySet<string>>> = {
  [MODULES.AUTH]: new Set<string>([AUTH_FEATURE.AUTHORIZE]),
  [MODULES.ORGANIZATION_CONSTANT]: new Set<string>([
    ORGANIZATION_CONSTANT_FEATURE.GET_FROM_APP,
    ORGANIZATION_CONSTANT_FEATURE.GET_FROM_ENVIRONMENT,
  ]),
  [MODULES.VERSION]: new Set<string>([VERSION_FEATURE.GET_ONE]),
};
```

server/src/modules/personal-access-tokens/interceptors/pat-scope.interceptor.ts
```diff
@@ -29,16 +47,51 @@ export class PatScopeInterceptor implements NestInterceptor {
...
+       this they would land in the app-pinned branch below and be capped mid-flight, breaking every
+       live embed session the moment this deploys.
+       
+       Safe precisely because it is narrow: until this change ships, the ONLY thing that could put an
+       appId on a JWT was the embed flow — a workspace token pinning a session to an app is what this
+       change introduces. So "appId and no patScope" cannot mean anything else.
+       
+       Remove once every pre-deploy session has expired (PAT sessions idle out in 60 minutes; embed
+       sessions follow the token's sessionExpiryMinutes). */
+    if (!user.patScope && user.patAppId) {
```
akshaysasidrn: This fallback branch has no end date.

It is safe today: the `scope` column is NOT NULL (migration `1746727457618`, unchanged by `1787724102950`) and both places that create PAT sessions now write `patScope`, so "appId but no patScope" can only be a session from before deploy. It stays safe only as long as no future code creates a PAT session with an appId and forgets `patScope`. Nothing will remind anyone to delete it.

Should we file a follow-up issue with a date, or make the branch deny once the longest embed `sessionExpiryMinutes` has passed since deploy?

Either way, the ten-line comment reads from the code apart from the removal condition:

```suggestion
    // TRANSITIONAL: pre-patScope embed JWTs; drop once sessions minted before deploy expire
    if (!user.patScope && user.patAppId) {
```

server/test/modules/personal-access-tokens/e2e/pat-session.spec.ts
```diff
@@ -101,6 +127,49 @@ describe('Personal access token session exchange', () => {
...
+      const res = await exchange(token).send({ appId }).expect(201);
+
+      const payload = JSON.parse(Buffer.from(res.body.authToken.split('.')[1], 'base64').toString());
+      expect(payload.isPATLogin).toBe(true);
+      expect(payload.appId).toBe(appId);
+      expect(payload.organizationIds).toEqual([orgId]);
+      /* The token's own kind, and it stays 'workspace'. This is the distinction PatScopeInterceptor
+         branches on: pinning a session to an app must NOT turn a workspace token into an app one,
+         which is what would hand it the embed flow's unrestricted exemption. */
+      expect(payload.patScope).toBe('workspace');
```
akshaysasidrn: `server/docs/testing.md` ("Assertions") asks for one `toMatchObject` instead of a chain of `toBe`.

```suggestion
      expect(payload).toMatchObject({ isPATLogin: true, appId, organizationIds: [orgId], patScope: 'workspace' });
```

**Also on these lines.** `createApp` (line 49) reloads the user and patches `organizationId`; `beforeAll` already gets `user` from `createUser`, so hoisting it to a `let user` and using it as `owner` drops the reload.

server/src/modules/personal-access-tokens/interceptors/pat-scope.interceptor.ts
```diff
@@ -48,4 +101,93 @@ export class PatScopeInterceptor implements NestInterceptor {
...
+    return undefined;
+  }
+}
+
+/* Deliberately a local copy rather than an import of the identical helper in @otel/tracing:
+   importing it pulls telemetry -> utils.helper -> data-sources/types -> the compiled plugin bundle
+   -> thrift -> an ESM-only uuid, which Jest cannot parse, and this file's unit test is otherwise
+   hermetic. A two-line regex is a smaller price than coupling an auth check to the telemetry
+   module. Keep the two in step; the shape is fixed by ToolJet's own routing.
+   App-scoped routes look like /api/apps/<uuid>/... — the uuid match avoids catching /apps/slugs. */
```
akshaysasidrn: Six lines for a two-line regex. The one fact a reader cannot get from the code is the Jest import chain.

```suggestion
// Local copy of @otel/tracing helper: importing it drags ESM-only uuid into Jest
```

server/src/modules/personal-access-tokens/constants/scopes.ts
```diff
@@ -122,6 +126,116 @@ export const PAT_ALLOWED_BUNDLES: PAT_BUNDLE[] = [
...
+  MODULES.APP_ENVIRONMENTS,
+  MODULES.ORGANIZATION_CONSTANT,
+  MODULES.DATA_QUERY,
+  MODULES.GLOBAL_DATA_SOURCE, // the queries cannot run without their datasource
+  MODULES.CUSTOM_STYLES, // changes how the app PAINTS — linting the DOM without it measures a lie
+
+  /* Not observed on the measured boot, and required anyway: useAppData only fetches
+     GET /v2/apps/:id/versions/:versionId when the URL carries ?version=, and the measured boot had
+     none. A versioned boot 403s without this. Narrowed to the single read below; the route is
+     /apps/:id/versions/:versionId, so the app pin covers it and the PUTs are already read-only. */
```
akshaysasidrn: Reads from the code apart from one trap: this module is not hit on every boot, so someone trimming the list by observation would drop it.

```suggestion
  // ?version= boots hit GET versions/:id; not on every boot
```

server/test/modules/personal-access-tokens/e2e/pat-session.spec.ts
```diff
@@ -44,6 +46,30 @@ describe('Personal access token session exchange', () => {
     return { token: res.body.token, id: res.body.id };
   };
 
+  const createApp = async (name: string): Promise<string> => {
+    const owner = await getDefaultDataSource()
+      .getRepository(User)
+      .findOne({ where: { id: userId } });
+    /* createApplication reads user.organizationId, which is NOT a DB column — createUser sets it in
+       memory and a findOne-loaded user therefore has none. Left undefined it violates
+       apps.organization_id NOT NULL, so set it explicitly from the workspace under test. */
```
akshaysasidrn: The trap is real and worth one line; the rest reads from the assignment below.

```suggestion
    // organizationId is in-memory only; findOne user lacks it, apps.organization_id NOT NULL
```

akshaysasidrn (approved):

gsmithun4 (approved):

kavinvenkatachalam (approved):

PR #17947, opened 2026-09-15
Fix: Skip response-schema dereference in OpenAPI spec processing

akshaysasidrn:
## 📝 What this does
Stops the OpenAPI spec worker from expanding every operation's response schemas, which nothing reads and which are about 99.9% of the expanded bytes on Microsoft Graph. Same rows saved, 3x less CPU, and the job no longer needs a 5 minute lock, a job-wide transaction, or a memory guard to get through.

## 🔀 Changes
- Dereferenced only `parameters` and `requestBody` per operation instead of the whole operation (Graph, 300 sampled ops: 159,899 to 21,463 node visits; full spec 16 s to 5 s)
- Removed the write-only `response_schemas`, `request_body_schema_raw`, `response_schemas_raw` columns, their unshipped migration, `markCircularRefs`, `buildRefPathIndex` and `MemoryGuard`
- Removed the job-wide transaction, its `SET LOCAL` timeouts and the 5 minute BullMQ lock override, since every reader is gated on status until the last batch is saved
- Added processor unit specs (3 were red before the fix) and `run()` specs for the `openapiv2` plugin, plus a skipped Microsoft Graph bench to run on demand
- Trimmed comments across the feature to the ones that explain what the code can't

## 🧪 How to test
- [x] `cd server && SKIP_GLOBAL_SETUP=1 NODE_ENV=test npx jest --config jest.config.ts test/modules/openapi-spec/unit/openapi-spec.processor.spec.ts`: 17 pass
- [x] `cd plugins && NODE_ENV=test npx jest packages/openapiv2`: 10 pass
- [x] Graph bench: change `describe.skip` to `describe` in `openapi-spec.processor.bench.spec.ts`, then run it with `NODE_OPTIONS=--max-old-space-size=4096` and the same server command. 17,777 rows, 0 skipped, about 18 s on an idle machine, p95 about 2 ms per operation
- [x] Add a Microsoft Graph OpenAPI datasource in the UI; status flips to Ready without the worker stalling; query editor lists operations and runs one


server/src/modules/openapi-spec/processors/openapi-spec.processor.ts
```diff
@@ -366,19 +270,18 @@ export class OpenApiSpecProcessor extends WorkerHost {
...
-    // parameter overrides one with the same (name, in) - e.g. a shared {id} path parameter
-    // reused across GET/PUT/DELETE. Previously dropped entirely since only operation.parameters
-    // was read. Dereferenced separately since pathItem.parameters isn't part of rawOperation's
-    // own subtree (a $ref'd shared parameter here still needs resolving).
+    // Only parameters/requestBody are dereferenced: nothing reads responses, which dominate the
+    // expanded size of large specs (e.g. Microsoft Graph). Cloning pre-expansion is cheap.
+    const operation = {
+      ...rawOperation,
+      ...this.dereferenceSubtree(
+        structuredClone({ parameters: rawOperation.parameters, requestBody: rawOperation.requestBody }),
```
akshaysasidrn, line 252: **Changed: only `parameters` and `requestBody` are dereferenced.**

Before, the whole operation was cloned and expanded, including `responses`, and then the responses were thrown away.

Example from Microsoft Graph, `GET /identityGovernance`:
- parameters expand to about 0.1 KB
- responses expand to about 928 KB

Nothing reads responses. The list query skips them, and plugin `run()` only uses `host`, `path`, `operation` and `params`.

On the full Graph spec this cut the dereference work from 16 s to 5 s, with the same rows saved.

server/src/modules/openapi-spec/processors/openapi-spec.processor.ts
```diff
@@ -449,64 +339,35 @@ export class OpenApiSpecProcessor extends WorkerHost {
     return bodyParam?.schema || null;
   }
 
-  private extractResponseSchemas(operation: Record<string, any>, version: string): Record<string, any> | null {
```
akshaysasidrn, line 452: **Removed `extractResponseSchemas`.**

It built `responseSchemas`, which this branch was already saving as `null`, and no code reads it.

Example: for `GET /admin` it built an 834 KB object that was then discarded.

server/src/modules/openapi-spec/processors/openapi-spec.processor.ts
```diff
@@ -123,136 +101,74 @@ export class OpenApiSpecProcessor extends WorkerHost {
...
-      // and immediately fanned out to every environment before moving to the next, instead of
-      // dereferencing the whole spec up front and only then persisting - peak memory is bounded
-      // to one batch's worth of operations (DEFAULT_OPENAPI_SPEC_BATCH_SIZE), not the whole
-      // spec's operation count. IMPORTANT: that bound only holds because `batches[i]` is
-      // explicitly cleared once consumed (see below) - `batches` itself lives for the whole job,
-      // so leaving a processed slot populated would keep every operation's dereferenced subtree
-      // (dereferenceInternal mutates pathItem/operation in place, replacing $ref pointers with
-      // the actual, potentially large, expanded schema objects) reachable for the rest of the
-      // job, growing roughly linearly with batches processed instead of staying flat.
-      await dbTransactionWrap(async (manager) => {
```
akshaysasidrn, line 152: **Removed the transaction around the whole job, and its `SET LOCAL` timeouts.**

Nobody can see half-written rows, so it was not protecting anything:
- status stays `processing` until the last batch is saved
- the editor only loads operations when status is `ready`
- the query guard blocks runs while processing

It did hold one database connection open for the entire run while the worker was busy with CPU work. The timeouts only existed so that long transaction could not hang.

Now the delete, each batch save, and the status update each run on their own.

server/src/modules/openapi-spec/processors/openapi-spec.processor.ts
```diff
@@ -64,21 +57,11 @@ interface PathItemEntry {
 const FETCH_TIMEOUT_MS = 30000;
 const HTTP_METHODS = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'];
 
-// Thrown to unwind out of the single per-job transaction when termination is observed mid-loop -
-// distinguishes "cancelled cooperatively" from a genuine processing failure so the catch block
-// can finalize with CANCELLED instead of FAILED.
-class OpenApiSpecCancelledError extends Error {}
```
akshaysasidrn, line 70: **Removed `OpenApiSpecCancelledError`.**

It was thrown only to break out of the job-wide transaction when a cancel came in.

Example, cancel arrives after batch 3:
- before: throw, roll back, catch, mark cancelled
- now: mark cancelled and return

server/src/modules/openapi-spec/utils/memory-guard.util.ts
```diff
@@ -1,57 +0,0 @@
...
-  }
-
-  currentHeapUsedMB(): number {
-    return Math.round((process.memoryUsage().heapUsed / 1024 / 1024) * 100) / 100;
-  }
-
-  // Call this between bounded units of work (batches). Throws a normal Error - caught by the
-  // processor's existing try/catch, which marks the job FAILED with this message - instead of
-  // letting the process silently grind toward an uncontrolled crash.
-  assertWithinCeiling(stageLabel: string): void {
```
akshaysasidrn, line 41: **Removed `MemoryGuard` (whole file).**

1. It checks memory between batches, after the work that could run out of memory. A real out-of-memory crash kills the process before this check runs.
2. `os.totalmem()` reports the host's memory inside a container, not the container limit.

After the fix one batch of 50 operations holds about 0.2 MB, down from 12.7 MB, so there is nothing left for it to guard.

server/src/modules/openapi-spec/utils/circular-ref.util.ts
```diff
@@ -49,83 +27,3 @@ export function pruneCircularRefs<T>(value: T): T {
...
-
-// Maps every named schema/parameter/etc. object to its canonical document pointer (e.g.
-// "#/components/schemas/UserRequest"), built ONCE per job from the pristine, pre-dereference
-// spec. dereferenceInternal mutates objects in place rather than replacing them, so object
-// identity survives dereferencing - a value found here still matches the same object after
-// dereferenceSubtree has run. Needed because by the time markCircularRefs runs, the original
-// $ref STRING that pointed at a now-live-circular object is long gone (dereferenceInternal
-// already replaced every $ref with the resolved value) - this is the only way to recover a
-// meaningful name instead of a locally-invented, context-free path like "#" or "#/properties/x".
-export function buildRefPathIndex(spec: Record<string, any>, version: string): Map<object, string> {
```
akshaysasidrn, line 61: **Removed `buildRefPathIndex` and `markCircularRefs`.**

They only fed the `request_body_schema_raw` and `response_schemas_raw` columns, which are removed too.

`pruneCircularRefs` stays, because request bodies still have real cycles. Example: `driveItem.children` points back to `driveItem`, so the second visit is saved as `{}`.

server/src/entities/openapi_spec_operation.entity.ts
```diff
@@ -63,23 +60,6 @@ export class OpenApiSpecOperation extends BaseEntity {
   @Column({ name: 'request_body_schema', type: 'jsonb', nullable: true })
   requestBodySchema: Record<string, any> | null;
 
-  // Per-status-code response schemas, e.g. { "200": {...}, "404": {...} }. Only ever fetched via
-  // the single-operation lookup (getOpenApiSpecOperation, by row `id`) - listOpenApiSpecOperations
-  // deliberately selects a lightweight column subset that excludes this.
-  @Column({ name: 'response_schemas', type: 'jsonb', nullable: true })
```
akshaysasidrn, line 69: **Removed `response_schemas`, `request_body_schema_raw` and `response_schemas_raw`.**

They were written but never read:
- `listOpenApiSpecOperations` selects only the light columns
- the editor only uses `parameters` and `requestBodySchema` from `getOpenApiSpecOperation`

Example: on Graph, `response_schemas` alone came to about 2.2 GB per environment, before the two raw text copies.

server/migrations/1784527469000-AddResponseSchemasToOpenapiSpecOperations.ts
```diff
@@ -1,34 +0,0 @@
-import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
-
-export class AddResponseSchemasToOpenapiSpecOperations1784527469000 implements MigrationInterface {
-  public async up(queryRunner: QueryRunner): Promise<void> {
-    await queryRunner.addColumns('openapi_spec_operations', [
```
akshaysasidrn, line 5: **Deleted instead of adding a reverse migration.**

It only exists on this feature branch and has not shipped. If you ran it locally, drop the three columns by hand or reset your dev database.

server/src/modules/openapi-spec/constants/queue-config.ts
```diff
@@ -1,29 +1,5 @@
...
- * expired in Redis - after maxStalledCount (default 1) such reclaims, the job is permanently
- * failed with `UnrecoverableError: job stalled more than allowable limit`. This processor's
- * per-batch dereference/prune work on a large spec (e.g. Microsoft Graph, 17k+ operations) is
- * synchronous CPU work heavy enough to occasionally starve the event loop past BullMQ's default
- * 30s lockDuration, which reads as "the worker died" even though it's still actively working -
- * failing a job that would otherwise have completed. Raised to 5 minutes for both settings so a
- * slow batch has real headroom to renew its lock before either it expires or the next stalled
- * sweep runs.
- */
-export const OPENAPI_SPEC_STALLED_CHECK_INTERVAL_MS = parseInt(
```
akshaysasidrn, line 27: **Removed the 5 minute lock and stall check override.**

It was raised so BullMQ would not mistake a slow Graph batch for a dead worker. Batches are now fast, so the BullMQ defaults (30 s lock) are enough.

Example: on the full Graph spec, 95% of operations take about 2 ms each.

server/src/modules/openapi-spec/constants/index.ts
```diff
@@ -1,24 +1,13 @@
...
+// Operations per dereference/insert batch (OPENAPI_SPEC_BATCH_SIZE). Keep rows x columns under
+// Postgres's 65,535 bind-parameter limit.
 export const DEFAULT_OPENAPI_SPEC_BATCH_SIZE = parseInt(process.env.OPENAPI_SPEC_BATCH_SIZE) || 50;
 
-// Ceiling on the single transaction spanning the whole job (clear + dereference + persist every
-// batch for every environment) - checked once per batch, same synchronization point as the
-// termination/memory checks, since there's no way to abort a transaction mid-query without
-// racing the connection itself. Guards against a transaction (and the row locks/replication slot
-// it holds open) running unbounded on a pathologically large spec.
-export const OPENAPI_SPEC_JOB_TRANSACTION_TIMEOUT_MS =
```
akshaysasidrn, line 19: **Removed `OPENAPI_SPEC_JOB_TRANSACTION_TIMEOUT_MS`.**

Its only use was the `SET LOCAL` timeouts on the job-wide transaction, which is gone.

akshaysasidrn (commented):
Notes added as inline comments

server/src/entities/openapi_spec_operation.entity.ts
akshaysasidrn: ### For later: showing the response in the query editor

When a user picks an API in the query editor, we'll want to show what it returns. These columns were probably the start of that. The schema comes from the spec, so this works before the query is ever run.

The catch was **how** it was stored: every response fully expanded, in every row. On Microsoft Graph that is about 2.2 GB per environment, which is what stalled the job.

#### The idea: store the short version, expand only what the user opens

**What we save for `GET /users/{user-id}`**

Before (removed), the whole shape is copied into the row, and repeated in thousands of other rows:
```json
{ "200": { "id": "string", "displayName": "string",
           "manager": { "id": "string", "displayName": "string", "manager": { ... } } } }
```

After, the row only keeps the name of the schema:
```json
{ "200": { "$ref": "#/components/schemas/microsoft.graph.user" } }
```

Each schema (`user`, `directoryObject`, ...) is saved **once** in its own table and shared by every operation that uses it.

**What the user sees**
```
Response  [200 ▾]
user
├ id: string
├ displayName: string
└ manager: directoryObject  ▸     click to load it
```

#### What to build

| Part | Change |
|---|---|
| Processor | Save `responses` as-is, plus one row per schema in a new `openapi_spec_components` table |
| API | Operation fetch already returns the row. Add one endpoint to load a schema by name |
| Query editor | Response panel with a status picker and a tree that loads a schema when it is expanded |

#### Why this is cheap

| | Expanded (removed) | Short version |
|---|---|---|
| Per operation | ~125 KB, up to ~1 MB | ~1 KB |
| Graph total | ~2.2 GB per environment | ~18 MB rows + ~5 MB schemas |
| Loops like `driveItem.children → driveItem` | cut off as `{}` | just another ▸ |

Request bodies can use the same approach later.

PR #18005, opened 2026-09-19
feat: add OpenRouter marketplace plugin

Navaneeth-pk:
## What

A marketplace data source for [OpenRouter](https://openrouter.ai): one credential reaching hundreds of models across providers, where the provider-specific connectors need a plugin and a key each.

The API is OpenAI-compatible, so this uses the same SDK the OpenAI connector does with the **Base URL** left editable. That costs nothing and makes the same connector serve Together, Groq, Fireworks, LiteLLM, vLLM or a self-hosted gateway, including deployments where prompts may not leave the network.

No competitor low-code platform ships an OpenRouter connector today.

## Operations

| Operation | Returns |
| --- | --- |
| Chat | `message` plus `finish_reason`, the `model` and `provider` that actually served the request, `usage`, `id` |
| Generate embedding | `embedding` for one input, `embeddings` for several |
| List models | The live catalogue with ids, context lengths and pricing, for a dropdown |

Chat returns an object rather than the bare string the OpenAI plugin returns, so a component binds `data.message` while the app can still show which model answered.

## The two fields that are the reason to route at all

- **Fallback models**: models to try when the first choice is down, rate limited, or refuses. Sent as `models` alongside `model`; the response reports which one answered.
- **Provider data policy**: set to "Only providers that do not train on prompts" it sends `provider.data_collection: deny`, refusing any provider that trains on what you send. This is the setting that makes the connector usable for customer data.

The error handler keeps `error.metadata.provider_name` and `.raw`, which is the part that says *which* provider refused and why; without it an upstream failure arrives as an unactionable "400 Bad Request".

## Verification

Tested end to end against the live API with a real key on a local Enterprise stack:

- Connection test, chat, embeddings and list models all run from the query editor.
- Fallback routing: confirmed live that OpenRouter accepts `model` and `models` together, and that an unknown id in the list 400s the whole request rather than being skipped. That finding is recorded as a comment in `buildChatPayload` because an earlier reading of it was wrong.
- Data policy `deny` verified against a prompt that a training provider would otherwise have served.
- Built an app through the AI builder against the installed data source and clicked through it: the model bound `data.message` correctly.
- CI steps run locally in a clean checkout: `npm ci --prefix marketplace`, eslint on the new files, and `npm run build` for the workspace.

## Not included

- The plugin icon and bundle still need publishing to the marketplace S3 bucket (`marketplace/scripts/upload-to-s3.sh`) before release; until then the Marketplace tab shows a broken icon while the Installed tab, which reads the DB copy, renders correctly.
- `openrouter` is not added to `AI_QUERY_SUPPORTED_KINDS`; AI-written queries for this kind are untested.
- `marketplace/package-lock.json` also gains quickbooks' `form-data` dependency, which was missing from the committed lock.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Navaneeth-pk:
Paired with ToolJet/tooljet-mcp#65, which adds this data source to the MCP datasource catalogue so the AI builder can write queries against it and bind `data.message`. That one should merge after this.

Navaneeth-pk:
**On the red "Check New Package Licenses"**: it is a defect in the check, not a licence problem, and every new marketplace plugin trips it.

The workflow's `extractAddedPackages` matches any added line of the form `"key": "<digit|~|*>..."`, so it reads the manifest's own `"version": "1.0.0"` as a package named `version` and looks it up on npm. It also looks up `@tooljet-marketplace/common`, which is an internal workspace package and is not published, so both come back `UNKNOWN` and fail.

The three real dependencies pass: `openai` Apache-2.0, `@vercel/ncc` MIT, `typescript` Apache-2.0.

The workflow (March) predates the ServiceNow plugin (July), so that PR would have failed the same way. Worth fixing separately by skipping `version` and any `@tooljet-marketplace/*` scope.

ganesh8056 (approved):

PR #18044, opened 2026-09-22
Fix: Text input inside a list view shows the previous query run's data

devanshu052000:
Closes - https://github.com/ToolJet/tj-ee/issues/5487

This pull request strengthens and clarifies how `listItem` references are detected in ListView row bindings, ensuring that rows properly track and update with ListView data changes. The main improvements are the addition of comprehensive integration tests covering all relevant reference shapes, and a refined regular expression for detecting standalone `listItem` references.

**Testing improvements:**

* Added a thorough integration test suite in `listViewListItemDependency.spec.js` that:
  - Verifies that ListView row bindings re-resolve when ListView data changes, for various reference shapes (property access, bare identifier, whitespace-padded, etc.).
  - Ensures that bindings referencing both a query and `listItem` update correctly and do not lag behind data changes.
  - Pins exactly which reference shapes are accepted or rejected as `listItem` dependencies, guarding against accidental changes to dependency tracking.

**Dependency detection logic:**

* Refined the regular expression in `checkSubstringRegex` (in `utils.js`) to more precisely match `listItem` as a standalone reference, including cases with property access, index access, optional chaining, whitespace, or end of string, while excluding false positives (e.g., inside string literals, as part of other identifiers).

These changes ensure that ListView row bindings remain in sync with their data and that future changes to dependency detection are deliberate and well-tested.


kavinvenkatachalam (dismissed):

kavinvenkatachalam (approved):

noordeen123:
Out of scope as discussed
- {{queries.runjs1.isLoading ? "" : listItem}}: one run behind
- {{ listItem}}, {{String(listItem)}}, {{JSON.stringify(listItem)}}, Row: {{listItem}}: never populated

